PR #119, opened 2016-03-03
PerryOddsAndEnds

perrydv:


perrydv:
This merge contains several changes:
declare and setSize now both handle either each dimension (size) as an argument or c(size1, size2, etc) notation for the sizes.
Now model$X(...) is supported, where X is calculate, simulate, getLogProb, calculateDiff, or getParam
There is better error trapping and messages from some major compilation steps.
There is a test-getParam and a test-declare (which also includes setSize)

PR #154, opened 2016-05-13
numeric, integer, matrix, array

danielturek:
Added numeric(), integer(), matrix() and array() to the DSL.  Has passed testing, and includes Rd documentation for each.  Ready for review and to merge.

UserManual already contains documentation for these.


perrydv:
I suggest adding to test-declare some tests that use numeric, matrix and/or array as part of larger expressions, to ensure they are handled as intermediates correctly.

danielturek:
Tests have already been added in test-declare.R.

perrydv:
I see you added a test for use of numeric in an addition.  Thanks.

PR #155, opened 2016-05-16
Merge filtering branch with devel

NLMichaud:


danielturek:
@NLMichaud one oddity I noticed here, that I'm not sure is right.  Your PR includes changes to MCMC_spec.R, which put the controlDefaultsList back into MCMC_spec.R.  Is that correct?

That had since been removed from there, and these default values have been put into .options.R  (where you also made some changes / additions).

NLMichaud:
@danielturek Thank you for catching that! I've commented out the controlDefaultsList from MCMC_spec.R

PR #173, opened 2016-06-07
Made FLIBS optional in configure.ac

perrydv:
I’ve modified configure.ac to add an additional —enable-flibs argument so that we can leave $(FLIBS) out of the Makevars when not building for CRAN.  My autoconf seems to be 0.01 version ahead from what was used previously, giving a surprising amount of changes in configure.  Duncan let me know any issues there.


duncantl:
Hi Perry

Looks good to me. 
Was nimbleGraph.cpp supposed to be in this pull. It seems quite different from the FLIBS issue.

perrydv:
No, nimbleGraph.cpp was a recent addition, and github desktop thought it was a new file when it shouldn’t have been, but I went with that and included it in this merge.  When I first tried to create it (last week), I had confusing behavior from github and github desktop.  It wouldn’t recognize a new file, just didn’t see it.  So I went to the repo online and created the file, and then github dekstop recognized my changes to the file locally.  And then for some reason during this merge for the FLIBS issue, github desktop indicated the nimbleGraph.cpp was new, even though it wasn’t.  I don’t understand what happened in the git workflow, but anyway nimbleGraph.cpp should be on devel so I went ahead and included it in this merge.

PR #174, opened 2016-06-08
added timing as a run-time option to buildMCMC

perrydv:
I have already merged run.time({arbitrary expression}) functionality.  This pull request is just for the use of timing functionality in buildMCMC, so we can review that together.

Example:
library(nimble)

mc <- nimbleCode({
    a ~ dnorm(0,1)
    b ~ dnorm(a, 1)
})

m <- nimbleModel(mc)
mcmc <- buildMCMC(m)

compiled <- compileNimble(m, mcmc, dirName  = '.')
compiled$mcmc$run(10000, time = TRUE)
compiled$mcmc$getTimes()
compiled$mcmc$run(10000, time = FALSE)
compiled$mcmc$getTimes() ## zeros indicate timing was not run


danielturek:
Thanks @perrydv for submitting via PR.

Looks good.  I couldn't think of any better naming than those you already chose:
time (= FALSE),
samplerTimes,
and getTimes()

Merging in.

Issue #175, opened 2016-06-13
enhance setData for simulation studies


paciorek:
We can't currently do this:
m$resetData()
m$simulate('y')
m$setData('y')

although I think that would be the natural workflow for a simulation study in which we resimulate data nodes.

The alternative is clunky
m$simulate('y')
m$setData(y = list(m$y))

danielturek:
@paciorek just to clarify, I think they currently supported syntax is actually:

m$simulate('y')
m$setData(list(y = m$y))

But your comment is understood, none the less.

paciorek:
Yes, I garbled it.

perrydv:
I think 

m$simulate(‘y’, includeData = TRUE) 

is a good option for this.

I think the core of the suggestion is that m$setData should allow simple naming of variables or nodes, in which case their values should not be modified but their data flags should be set.  I like that idea (and have also run into this).  We could also consider using … to take arguments to setData so that the user need not user list().

danielturek:
Worth talking about this option:

m$simulate(‘y’, includeData = TRUE)

Maybe I'm missing your implication here.  If y is already tagged as being data, I agree, this already works fine, and simulates new values into y.  But if y is *not* tagged as data, are you suggesting this should also have the affect of tagging y as data?  That I don't agree with.

I think a call to setData() should be necessary to do the "tagging" as data.  Which gets right back to @paciorek 's original suggestions.

danielturek:
Nevermind, I see what you meant now.  Replacing the workflow of resetData(), simulate(), setData(), instead with simulate(..., includeData=TRUE).

danielturek:
Perry, did you see the github PR regarding this?  Looking for something
more sophisticated?  If so, that's ok, just let me know.

Daniel

paciorek:
Perry addressed this in full generality.

PR #176, opened 2016-06-14
new character vector argument functionality for setData method

danielturek:
Added functionality to `model$setData(data)` method.  `data` argument can now be a character vector of model variable names.  In this case, the values of these model variables are not changed, and only the designation of these variables as being 'data' occurs.

This addition was done in a very simplistic way, where the values of the model variables are accessed, to create a named list `data` argument, which then goes through all of the original processing (including re-setting the model variable values).  It wasn't done this way to be a quick fix, but instead to leave the original core of `setData()` unchanged, and have this addition to be a simple and well-defined advance step.

That being said, it's not the most efficient implementation possible, so certainly open to revision (but bear in mind exactly when and how often this inefficiency will be imposed).

roxygen documentation for `setData()` is also updated.

### DIFF
--- a/packages/nimble/R/BUGS_model.R
+++ b/packages/nimble/R/BUGS_model.R
@@ -284,26 +284,51 @@ Resets the \'data\' property of ALL model nodes to FALSE.  Subsequent to this ca
                                       return(invisible(NULL))
                                   },
                                   
-                                  setData = function(data, warnAboutMissingNames = TRUE) {
+                                  setData = function(..., warnAboutMissingNames = TRUE) {
 '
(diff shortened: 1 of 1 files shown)


danielturek:
Also note this is addressing #175.

perrydv:
@danielturek I made setData take arguments even more flexibly using ...  Now the following work:
```
model$setData('a','b')
model$setData(c('a','b'))
model$setData(a = 1:5, 'b')
model$setData(list(a = 1:5, 'b'))
```

I also added test-setData.R

Are you ok with this added flexibility?  If so please also look at my roxygen edits and revise as you see fit.

I think this PR has behaved as I hoped: I pushed a change on this branch and I think that means the PR is now for the updated branch.

If this seems good to you I'll put it back in your court to merge this PR.

perrydv:
Per discussion yesterday I am merging this PR.

danielturek:
Looks good.  Thanks for sending the PR my way in the first place.  I like the new functionality, quite flexible.

PR #180, opened 2016-06-27
Multinomial

DRJP:
RW_multinomial for sampling multinomial distributions.
A test script is included.


danielturek:
Great.  I will look at it in detail.  One request, can you please move the test script from:
`packages/nimble/R/test_RW_multinomial.R`
to:
`packages/test_RW_multinomial.R`

DRJP:
OK, file moved.

danielturek:
@DRJP I'm following along with the changes you've been making.  The improved adaptation and setup function quantities look good, and I especially really appreciate your attention to stylistic detail (similar to mine, I have to say).

My main question -- do you believe the implementation is somewhat stable, or do you envision more changes in the near future?

Minor comments:

- I would remove all the "checks" except for the final one (regarding the dmulti distribution), since they're all self-enforced in the setup function, and will never fail.
- In your test model, I would make pVecX, pVecY, and N be "inits" rather than "data" for building the model.  It all comes out the same in the wash, we catch these sorts of thigns, but technically "data" is values of observed stochastic nodes which contribute to the likelihood calculation (such as "Z"), and the others (pVecX, pVecY, N) are what we call right-hand-side only (RHSonly) nodes, which don't appear on the left hand side of any stochastic or deterministic declaration, so you only want to provide values for them (can still modify these values later), rather than attempting to set the "data" flag for these, which cannot be done, since they're not stochastic.  Make sense?  The distinctions between constants, data, and inits can certainly be confusing.

DRJP:
Thanks Daniel for pointing out the inits ~ data difference, that's really helpful!  This change and the redundant checks are removed.

The main new change is to have a better scheme for recycling the random number u. My previous attemp at this recycling would degenerate in less than 40 iterations due to finite precision in digital numbers. Here I avoid the problem by using an irrational number (pi) when recycling.  This permits us to call u ~ runif() just once (in setup) which provides a nice speed boost. 

I've tested fairly extensively with the test_RW_multinomial.R. For that problem I think the code is quite optimised.  The only limitation of the code I can think of right now would be in a complex model with very strong auto-correlation - in this situation the user would probably need to increase adaptInterval.  

Right now I don't envisage any further modifications. An alternative strategy to the one used in RW_multinomial would be to replace the nSwap <- rbinom() proposal with slice sampling, but that would merit an entirely new funciton.

danielturek:
@DRJP one additional request.  Could you add an appropriate description & documentation of the RW_multinomial sampler to the bottom of `MCMC_samplers.R` ?  You can easily follow the examples (syntax and content) of the sampler descriptions already there.

DRJP:
OK, I'll take a look.

I have been running some tests on models with both continuous and multinomial nodes. In these examples the multinomial node prevents autoBlock from running.  A call to autoBlock returns

Error in samplerFunction(model = model, mvSaved = mvSaved, target = target,  :
     cannot use RW sampler on discrete-valued target; try slice sampler

It would be a nice feature if the user could specify which subset of nodes autoBlock should be considering. That way autoBlock could still be used on continuous nodes in this scenario.

danielturek:
That would be a nice feature, to specify a subset of nodes for autoBlock.  Duely noted.

In the meantime, it'll require an update to the autoBlock code, to handle multinomial nodes differently.  I'll do that when I incorporate the RW_multinomial sampler into the MCMC engine.

When you're using NIMBLE, what version are you using?  Or are you building off one of our development branches on GitHub?  If so, which branch do you use?

DRJP:
OK

Right now I am using version 0.6 from the devel branch.  I install with  "R CMD INSTALL --build nimble"  from the "packages" directory.

danielturek:
Great, when I work this in, I'll do so on devel.

Documentation for the sampler looks good, thank you.

danielturek:
@DRJP after a fair bit of reworking, I ended up merging this in via another branch, rather than directly merging the pull request from your branch.  These changes are now on devel, and the RW_multinomial sampler is incorporated into the MCMC, and the autoBlock algorithm (although not including the feature you suggested, of specifying a subset of nodes for autoBlock to operate on).  At your leisure, please pull and build the package from devel branch, and try it out.

Thank you for the contribution!  I hope you stay actively involved in using, testing, and contributing to the NIMBLE package.

PR #181, opened 2016-06-29
fix to dependent path separation in new conjugacy checking

danielturek:
Fix to new conjugacy "dependent paths" processing, that was added by @perrydv .

Removed dependentNodePaths that have length = 0.

@perrydv please take a look at this one-line change, and merge into newNodeFxns if you approve.  It seems to work just fine, but I suggest you take a look at the email I sent with subject "Error in conjugacy checking", and the attached model and data file (which triggered the previous error), since I'm not sure why this error never showed up before.

@perrydv I leave it to you to merge this change in.

### DIFF
--- a/packages/nimble/R/MCMC_conjugacy.R
+++ b/packages/nimble/R/MCMC_conjugacy.R
@@ -127,7 +127,7 @@ conjugacyRelationshipsClass <- setRefClass(
                 if(is.null(conjugacyObj)) next
                 
                 depPathsByNode <- lapply(nodeIDsFromOneDecl, getDependencyPaths, maps = maps)  ## make list (by nodeID) of lists of paths through graph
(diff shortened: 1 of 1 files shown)


perrydv:
@danielturek nice catch.  I think I follow and it makes sense.  FWIW I think the error may be triggered by a somewhat more rare circumstance that it may at first appear.  I think a node that has no dependencies would return NULL from getDependencyPaths, and that case was stripped out in checkConjugacy2.  I think the case that makes getDependencyPaths return list() instead of NULL (cause of the bug) is if there are deterministic dependencies that do not lead to a stochastic dependency.  That would mean the stochastic node being checked is posterior predictive, with some deterministic calculation being made from it as well.  

Looking at this example also made me want to check with you if mean() is handled correctly in the linearity checking of dependencies.  But that's a different topic.  I'll merge this pull request.

PR #183, opened 2016-07-08
Windows uninitialized warning

perrydv:
This puts compiler directives in NimArrBase.h to avoid -Wmaybe-uninitialized from a case that occur in generated getParam code.


perrydv:
I am accepting this but did it via pull request to leave a clearer record.

PR #184, opened 2016-07-10
Added nimCat to DSL and progressBar to MCMC

danielturek:
@perrydv @paciorek Two changes in this PR:

1. Addition of nimCat to the DSL. Identical to nimPrint, except doesn't add a "newline" at the end.
2. Addition of progressBar option to MCMC, as a runtime argument (default TRUE).

To see how the progress bar looks, build off this branch and run the following examples:

library(nimble)
Rmodel <- readBUGSmodel('birats2.bug', dir = getBUGSexampleDir('birats'), data = 'birats-data.R', inits = 'birats-inits.R')
Rmcmc <- buildMCMC(Rmodel)
Cmcmc <- compileNimble(Rmcmc, project = Rmodel)

Cmcmc$run(20000)
Cmcmc$run(30007, reset = FALSE)         ## continue previous run
Cmcmc$run(10000, progressBar = FALSE)   ## turn off progress bar
Cmcmc$run(100003, reset = FALSE)        ## sort of slow run
Cmcmc$run(5000)     ## faster
Cmcmc$run(2000)     ## faster still
Cmcmc$run(1000)     ## ...
Cmcmc$run(500)
Cmcmc$run(100)
Cmcmc$run(40)  ## no progress bar when too few iterations


paciorek:
I'm slightly wary of defining nimCat simply as nimPrint without newlines
given how cat operates in R is a bit different than print. But I guess it's
an ongoing question of how much the DSL should mimic R.

What do you think about adding an arg to nimPrint, with default, as
"newline = TRUE" instead?

danielturek:
I'm certainly open to that way of going.  Perry, thoughts?

perrydv:
I think adding nimCat (as the replacement for the DSL synonym “cat”, I presume) makes sense.  The reason nimPrint (or print) is inconsistent with R behavior is that we haven’t had any string processing in compiled-generated C++, so we don’t have paste() in the DSL.  As a result it made sense to simply allow comma-separated arbitrary arguments to print() that are automatically pasted together.  Unfortunately that is not consistent with R’s print() but it is consistent with R’s cat().  So really the NIMBLE DSL print should arguably have been named cat in the first place.  When we add meaningful string processing, including paste, it would make sense to change print() to behave like R’s print().  To that end it would make sense to name what we have more consistently with R, which would be cat.

danielturek:
So I've missed the consensus here.... For background to Perry, I created a
pull request, which adds a new DSL keyword: cat() / nimCat(), which
functions identical to our current print() / nimPrint() except without
outputing a newline "\n" at the end.

So which way are we leaning here?

- Change our current print() to be named cat() ?
- Change our current print() to be named cat(), and add a newline=FALSE
argument ?  (breaks old code).
- Add a newline=TRUE argument to our current print() ?  (old code works,
and we have new functionality).
- Have both print() and cat()  (as my pull request does).

Thanks.  I just wasn't sure the conclusion of Perry's comments.

Daniel

paciorek:
I think consensus is to go with Daniel's PR for now and change nimPrint in
the future to be more like R.

danielturek:
If that's the case, then I'll leave it to @perrydv to look over the changes I made in the DSL / compiler, and merge this pull request.  Until I hear otherwise, I'm assuming this is out of my court.  Hopeful for this to go in the 0.6 release, though.

If this PR is merged, I still need to make updates to the User Manual, FYI.

I'll also encourage both of you to try out the example code I sent (first message in this conversation) to see if you like the "look & feel" of the progress bar.  Many different ways to go about it.

perrydv:
I looked at this and ran the example on OS X and Windows.  I like it and think we should merge it.  On my machine the cases with niter<=2000 were essentially instantaneous (the entire bar fills in at once), so perhaps the threshold for not showing a bar could be raised, but it may not hurt to have it lower rather than higher.   I sent an email note about the roxygen entry.  Since GitHub says this branch has conflicts to resolve, I am going to throw it back to @danielturek to merge.

perrydv:
@danielturek can I ask a favor while you are merging this PR?  The first line in each of cppOutputCout and cppOutputCoutNoNewline (which has "std::cout" in it) can be removed.  These have no impact whatsoever since the next line provides the return value of each function.  We made the switch a while ago to using a std::ostringstream object for output from C++, called _nimble_global_output.  Apparently the paste() creating the old return value from cppOutputCout was left there accidentally, with no effect.  Thanks.

danielturek:
@perrydv The threshold for not showing the progress bar is when the number of MCMC iterations is less than the number of '-' characters in the progress bar, in which case the logic for determining when to display the next '-' fails.  I'm leaving this as-is, since for some MCMCs (large model, lots of samplers) even a small number of MCMC iterations (e.g., 500) can take appreciable time, hence a progress bar is useful.

I touched up the roxygen documentation, per your other email.  Note, though, you might want to revise the documentation for print() / nimPrint(), since I had basically copied it from there.

I fixed the merge conflict.

I removed the old paste() lines (no longer used) from genCppCout and genCppCoutNoNewline, as you described.  I was a little perplexed by those, but what you said makes sense.

Merging this in.

PR #193, opened 2016-07-27
Effective times adapted

DRJP:
A variant on RW_block that maintains greater flexibility in adaptation when the current set of parameters is far from the target distribution.


danielturek:
@DRJP Thank you very much for the contribution.  I'll look over this in careful detail, but one question first.  Is the modified adaption routine in the block_ETA sampler based upon published theory?  Or is it more of a home-grown approach?

DRJP:
It's home-grown, so it needs thorough checking, testing & a paper ;0)

DRJP:
Daniel, I can see issues with the sampler as it is. Specifically, ETA does not grow as fast as timesAdapted once a point close to the global maximum has been sampled - this needs correcting. Also the tracking of the MaxLogProb would be better somewhere else. Do you prefer we develop here or delete the pull request and develop on my fork?

DRJP:
@danielturek I have performed fairly major surgery on the adaption part - it's quite simplified which helps clarify how adaptation will eventually go to zero. But I have not tested because right now I have a compile error which I don't yet understand when compiling the MCMC with the new sampler.

packages/nimble/R/MCMC_samplers.R
```diff
@@ -284,44 +286,52 @@ sampler_RW_block_ETA <- nimbleFunction(
             returnType(double(1))
             return(propValueVector)
         },
-        adaptiveProcedure = function(jump = logical()) {
-            timesRan <<- timesRan + 1
+        adaptiveProcedure = function(jump = logical(), alpha = double()) {
+            LogProb         <- model$getLogProb()
+            timesRan       <<- timesRan + 1
+            EWMALogProb    <<- alpha * LogProb + (1-alpha) * EWMALogProb
+            maxEWMALogProb <<- max(maxEWMALogProb, EWMALogProb)
```
DRJP: It seems this is the line generating the compile error. It is the only line where I use "max()". Perhaps then the problem is with alpha (parameter for exponential weighted moving average), which I had tried to write as a user definable run-time parameter with a default.

danielturek:
@DRJP Is this implementation reasonably stable now?  Or still in flux?

DRJP:
@danielturek Yes, I'm pretty happy with the new version, especially now that the original RW_sampler is recovered if "readaptability" is set to zero.  It works well on the toy example with readaptability=0.01. I put a default value of readaptability=0.1 so an increase in maxEWMA_LogProb of 1 results in a 10% reduction in effTA. Time will tell if this value is a good choice in a variety of situations. Perhaps you can think of a better (i.e. more user friendly) name for this sampler.

danielturek:
@DRJP Very sorry for the long delay on this, it had fallen off our radar.  In the recent release of version 0.6-4, we've introduced a new joint sampler, the adaptive factor slice sampler (AFSS) which generally outperforms our exisiting RW_block sampler.  We're shifting attention towards this and other variations of joint samplers, and welcome your thoughts or test cases along those lines.

DRJP:
OK - sounds great. Look forward to trying it out.

PR #210, opened 2016-08-02
Better error trapping for parseEvalNumericMany, getParam, and unknown names in size processing

perrydv:
Better error trapping for:
- parseEvalNumericMany, which is the place of looking up graphIDs used eventually by many model query functions including getDependencies.
- getParam for multiple or invalid variables or params
- unknown names during nimbleFunction compilation during size processing.


perrydv:
error trapping for getParam was fixed. Merging this.

PR #215, opened 2016-08-11
enforcing no "ragged" 2D conjugacy (Wishart)

perrydv:
This fixes incompatible matrix dimensions in 2D conjugate sampler by simply requiring that all dependencies use the full prior, i.e. no “ragged” uses.


danielturek:
Made a few changes to minimize code repetition, and remove all double-quotes.  Thanks for taking the initiative @perrydv and submitting via PR.

PR #219, opened 2016-08-22
Extern issue

perrydv:
- Cleans up extern template warnings

- Adds suppressCppCompileOutput to nimbleOptions, with default TRUE, which does not roll C++ compiler text into R output.

- removes blank “;” lines in newNodeFxn C++ class definitions that could also generate -Wpedantic warnings.


perrydv:
BTW if you are testing this it is up to you to put in compiler flags like -Wpedantic for your tests, since we each seem to be doing that in different spots and we have different issues we are able to flag on different systems.

PR #234, opened 2016-10-04
fixing conjugacy bug

danielturek:
Fixing uniqueness bug in conjugacy system.

@perrydv I have to point out this bug was introduced with the checkConjugacy2() system, where you added tracing dependency paths through graph.  It arose from overlooking that different paths through the graph could terminate at the same dependent node, with both paths satisfying conjugacy requirements, and duplicating those terminal nodes in the final list of dependent nodes.  This didn't occur in the original checkConjugacy() system.

Anyway, fixed with a simple addition of unique().

@paciorek would be great if this can make the submission & release, but perhaps it's too late?


paciorek:
I haven't submitted yet as Perry was working through your and my issues. So
we're good.

PR #259, opened 2017-01-31
Reduce path checking

paciorek:
doing this merge as a pull request with the thought that it might be good to have Perry or Daniel look at the changes to checkConjugacy before we merge to devel.


danielturek:
Changes reviewed, appear correct with basic testing.  Merging in.

Thanks for the improvement, @paciorek

PR #264, opened 2017-02-27
Values indexing

paciorek:
would be good for Perry to take a quick look at the diffs. 

One question: I set accessArgList$sortUnique <- FALSE in values_keywordInfo.  Don't see any reason we should be sorting/uniqueing nodes coming into values() but didn't look to really understand what sortUnique is all about.

### DIFF
--- a/packages/nimble/NEWS
+++ b/packages/nimble/NEWS
@@ -4,6 +4,8 @@ USER LEVEL CHANGES
 
 -- Updates to MCMC configuration API.  Member methods conf$addMonitors() and conf$addMonitors2() now accept argument ..., allowing multiple monitors to be added using the syntax conf$addMonitors('x', 'y', 'z').  Introduced a new member method, conf$printMonitors(), which nicely prints the current MCMC monitors.  This takes the place of the former method conf$getMonitors().  Now, conf$getMonitors() returns a character vector of the MCMC monitors, and the new method conf$getMonitors2() returns a character vector of the MCMC monitors2 field.
 
+-- Indexing of values() now allowed in the DSL, so syntax such as `x <- model$values(nodes[i])` or `values(model, nodes[i]) <- x` is supported.
+
 -- New sampling algorithm (RW_dirichlet) added to MCMC engine, for sampling non-conjugate Dirichlet distributions.
 
 -- Fixed bug preventing use of round() in DSL and BUGS code.
--- a/packages/nimble/R/RCfunction_compile.R
+++ b/packages/nimble/R/RCfunction_compile.R
@@ -132,8 +132,6 @@ RCfunProcessing <- setRefClass('RCfunProcessing',
                                    ),
                                methods = list(
                                    process = function(debug = FALSE, debugCpp = FALSE, debugCppLabel = character(), doKeywords = TRUE) {
-                                       print('here4')
-                                       if(exists('STOP')) browser()
                                        if(!is.null(nimbleOptions()$debugRCfunProcessing)) {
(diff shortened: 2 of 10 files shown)


perrydv:
Most of this makes sense to me.  A few questions:

What's up with the change to sizeWhich?  It seems unrelated to values and doesn't look right, but at the moment I'm rushing and haven't checked.  But I thought which() was working.

Did you want if(exists('STOP')) to stay in the code or was that a temporary debugging line?

Does model$values(nodes) work in R execution?  I think I ran into a problem but will have to check later.

Building nimble gives me this compiler warning:

accessorClasses.cpp:322:7: warning: unused variable 'k' [-Wunused-variable]
  int k = SMVA_Vec.size();

Let's discuss if we want the kind of indexing flexibility I just created for model$calculate etc.  That could be done for model$values as well.  Your idea of model$values(nodes)[i] <- value would be harder but doable.  

I still want to consider the vector <- scalar topic, but I see that as more general.

paciorek:
Replies inline:

Thanks for the check. Glad you took a look since there was obviously some
sloppiness I hadn't realized.  I'll make the relevant changes detailed
below and merge it into devel.


On Wed, Mar 8, 2017 at 11:51 AM, perrydv <notifications@github.com> wrote:

> Most of this makes sense to me. A few questions:
>
> What's up with the change to sizeWhich? It seems unrelated to values and
> doesn't look right, but at the moment I'm rushing and haven't checked. But
> I thought which() was working.
>
Hmm, weird. Not sure how that happened. I don't think I was monkeying with
that though at one point I was playing with caller$callerArgID on account
of working on catching when when there is LHS indexing of a function call.


> Did you want if(exists('STOP')) to stay in the code or was that a
> temporary debugging line?
>
> Whoops. Not supposed to be there.


> Does model$values(nodes) work in R execution? I think I ran into a problem
> but will have to check later.
>

Not sure. I don't think I did anything on the front of working with $.

> Building nimble gives me this compiler warning:
>
> accessorClasses.cpp:322:7: warning: unused variable 'k' [-Wunused-variable]
> int k = SMVA_Vec.size();
>

I thought I removed that from the branch but it must have stayed in
somehow.

> Let's discuss if we want the kind of indexing flexibility I just created
> for model$calculate etc. That could be done for model$values as well. Your
> idea of model$values(nodes)[i] <- value would be harder but doable.
>

Yeah, more indexing flexibility could be nice but not essential.
 values(model,nodes)[i] or model$values(nodes)[i]  could be nice because it
allows easier assignment from scalars, in particular rDIST on RHS.

perrydv:
To be clear, what I meant was values(model, nodes[ expression ] ), similar to model$calculate(nodes[ expression ] ) that I just implemented.  values(model, nodes)[expression] should work on RHS.  To make it work on LHS would take additional work.

I thought at some point we talked about model$values(nodes) on RHS.  Where did that idea go?

paciorek:
oh, right. Yes, values(model, nodes[expression]) could be useful and good
to have if we allow for calc/sim/etc. for consistency.

Higher priority for me (and on my todo list) would be model[[nodes[i]]].

If we can't do model$values(nodes) on LHS, then I think it may be confusing
to allow/publicize on RHS.

perrydv:
Oh right, that must be why we put aside the model$values(nodes) idea.

So do you want to clean up those two items, then we'll merge the pull request, then we'll get values(model, nodes[expression]) working, then we can look at the other topics.

paciorek:
Yes, will clean up and merge today or early tomorrow.

PR #275, opened 2017-03-19
removed copy from use of getMapAccessVector in values and modelValues copying

paciorek:
Perry, I followed your template code both for values and modelValues where getMapAccessVector is used. Please take a look and see if anything seems awry.


perrydv:
looks good

perrydv:
oops, closed without merging.  now it's merged

PR #276, opened 2017-03-22
changed conjugate wishart case to only handle link=identity

danielturek:
Conjugate Wishart case was not handling multiplicative link correctly.

One point to note, @paciorek, the multiplicative case isn't restricted to scalar multiples, (c*W), but is generally built for a matrix multiplication (A*W).  So, the incorrectness was probably more extreme than we realized.  I'm curious about this, though, since I thought we had worked through these cases previously.


paciorek:
I think it was handling AW correctly, but it wouldn't handle AWA^\top,
which I think would be the more likely scenario. But neither of  them are
all that likely and the implication of handling cW or AW is that we have a
bunch of calculations that are not needed if we just handle W alone.

PR #289, opened 2017-04-13
Consistent behavior and control over numeric object initialization and setSize copying etc.

perrydv:


perrydv:
I fixed an earlier problem in test-declare (not actually related to this PR, but I fixed it here anyway).

Now there is a problem I'm not understanding.  test-declare fails because it says nimC can't be found.  But nimC is in the NAMESPACE exports, and test-declare works on my machine.  Chris can you try it?

paciorek:
is all of this in devel? I just did git pull and didn't pull anything new
in.

perrydv:
No, it's in PR #289, which is where I made the comment from (but I guess that's not clear on receiving end).    What I'd like to do is merge this and the other pull request now for simplicity and then we can figure out the test-declare issue on devel.  OK?

paciorek:
yes, at this stage, having things in other branches is obviously confusing
me!

perrydv:
Also merging this.

PR #290, opened 2017-04-13
fixes to numerical imprecision in determining sequence length and possibly missing nrow or ncol for nimMatrix

perrydv:


perrydv:
Merging this.  We will pursue the test-declare issue on devel.

PR #292, opened 2017-04-17
nimbleList fixes

NLMichaud:
- This PR has fixes for CRAN check notes and warnings involving the use of `<<-` and `.self` in `nimbleList()`.  

- also removes the call to:  
`assign("makeNewNimListSEXPRESSIONFromC", makeNewNimListFromC, envir = globalenv()) 
` in cppDefs_RCfunction.R, instead assigning the ` makeNewNimListSEXPRESSIONFromC()` function to the nimbleInternalFunctions environment, where it can be called from C++.  

- `@export` tags were added to  `is.nl`, `eigenNimbleList`, and `svdNimbleList`

- slight change to roxygen help for `nimEigen` and `nimSvd`


paciorek:
Hmm, there are failures in testing. The first one I see is pretty basic, in test-declare (see below). Not sure the best way to proceed as the errors I'm first noticing don't seem like they would related to nimbleLists.

Here's the error in the 'numeric 2' test of test-declare:

> expected <- numeric(10)
> Rfun <- nimbleFunction(run = function() {
+     ans <- numeric(10)
+     returnType(double(1))
+     return(ans)
+ })
> Cfun <- compileNimble(Rfun)
compiling... this may take a minute. Use 'showCompilerOutput = TRUE' to see C++ compiler details.
compilation finished.
> test_that('numeric 1', expect_equal(Rfun(), expected))
> test_that('numeric 2', expect_equal(Cfun(), expected))
Error: Test failed: 'numeric 2'
* Cfun() not equal to `expected`.
Types not compatible: character vs double
> Cfun
function () 
{
    if (is.null(CnativeSymbolInfo_)) {
        warning("Trying to call compiled nimbleFunction that does not exist (may have been cleared).")
        return(NULL)
    }
    ans <- .Call("CALL_rcFun_6")
    ans <- CnativeSymbolInfo_[[1]]
    ans
}
<environment: 0x39435b8>
> debug(Cfun)
> Cfun()
debugging in: Cfun()
debug: {
    if (is.null(CnativeSymbolInfo_)) {
        warning("Trying to call compiled nimbleFunction that does not exist (may have been cleared).")
        return(NULL)
    }
    ans <- .Call("CALL_rcFun_6")
    ans <- CnativeSymbolInfo_[[1]]
    ans
}
Browse[2]> 
debug: if (is.null(CnativeSymbolInfo_)) {
    warning("Trying to call compiled nimbleFunction that does not exist (may have been cleared).")
    return(NULL)
}
Browse[2]> 
debug: ans <- .Call("CALL_rcFun_6")
Browse[2]> 
debug: ans <- CnativeSymbolInfo_[[1]]
Browse[2]> ans
[[1]]
 [1] 0 0 0 0 0 0 0 0 0 0

Browse[2]> 
debug: ans
Browse[2]> ans
[1] "CALL_rcFun_6"
Browse[2]> CnativeSymbolInfo_
$name
[1] "CALL_rcFun_6"

$address
<pointer: 0x7f4b6d108760>
attr(,"class")
[1] "NativeSymbol"

$package
NULL

attr(,"class")
[1] "NativeSymbolInfo"
Browse[2]>

NLMichaud:
The commit I just pushed should hopefully fix the above errors.  They arose because previously there was a line of code in the RwrapperFunCode that created the function that would create new nimbleLists if necessary.  This line of code is no longer in the RwrapperFunCode , but I had forgotten to update the indexing of the RwrapperFunCode after its removal , so a separate line of code was being overwritten.

PR #293, opened 2017-04-17
added isMultivariate method to model objects

danielturek:
I added a method `model$isMultivariate(nodes)` for model objects, for future use.

This followed the framework of recent updates to `model$isDiscrete()`, `model$getDistribution()`, etc, but I'm requesting that @paciorek review this addition, and make comments, changes, or merge this in, if it looks alright.

### DIFF
--- a/packages/nimble/R/BUGS_model.R
+++ b/packages/nimble/R/BUGS_model.R
@@ -161,11 +161,12 @@ nodes: A character vector specifying one or more node or variable names.
 Details: The return value is a logical vector with an element for each node indicated in the input. Note that variable names are expanded to their constituent node names, so the length of the output may be longer than that of the input. 
 '
                                       nodeNames <- expandNodeNames(nodes, unique = FALSE)
-                                      multi <- rep(FALSE, length(nodeNames))
-                                      for(i in seq_along(nodeNames)) {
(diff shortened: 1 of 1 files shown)


packages/nimble/R/BUGS_model.R
```diff
@@ -150,6 +150,25 @@ Details: The return value is a character vector with an element for each node in
...
+                                      multi <- rep(FALSE, length(nodeNames))
+                                      for(i in seq_along(nodeNames)) {
+                                          nodeExpanded <- expandNodeNames(nodeNames[i], returnScalarComponents = TRUE)
+                                          if(length(nodeExpanded) > 1) multi[i] <- TRUE
+                                      }
+                                      return(multi)
+                                  },
+
                                   isDiscrete = function(nodes) {
                                                                           '
```
paciorek: We already have the info in the distributions object, so how about grabbing it as follows and checking if >= 1?
```
> getDistributionInfo('dmnorm')$types$value$nDim
[1] 1
> getDistributionInfo('dnorm')$types$value$nDim
[1] 0
```

danielturek:
@paciorek , I modified the `isMultivariate` method to use `getDistributionInfo()`.

If you think this implementation is better, then please merge, but I suppose that would be after the upcoming release.

Issue #296, opened 2017-04-20
Running a compiled function crashes RStudio


fritzo:
To reproduce: http://rpubs.com/fritzo/compiler_demo

danielturek:
I think this is an easy one.  Specifying arg1 and arg2 as double(2) in the
run function definition means *2-dimensional* arguments, which is what the
compiled version is expecting. (i.e. arrays or matrices).

So either change the argument specifications to "arg1 = double(1)", to a
vector such as c(0.0, 0.5), or change the values of arg1, arg2, and
expectedResult to be R matrices (or dimension=2 arrays).  (for example,
array(0, c(2,2)).

perrydv:
Right, this error *was* caught, but in the manner of our current run-time error-trapping, we emit an informational message and continue execution.  A topic that came up in our discussion earlier this week is that our run-time error trapping (from both permanent and generated C++) has been stable for a while, so it could make sense to make them bail out of execution and return safely to R.

I am going to close this issue since it is not a bug but leads to a more general feature request.  Let's make a wiki page or project to keep track of ideas for that.

fritzo:
The current warn-and-continue behavior of assertions leads to a bad user experience, in my case crashing RStudio and causing me to lose about 30 minutes of unsaved work. I'll write up a little NIM_ASSERT(-,-) macro in C++ to avoid this pain in the future.

perrydv:
That would be great.  See [section 6.2 of Writing R Extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Error-handling) for R's API to stop() and warning().  I'd guess error() is a good way to bail out back to R.  You could start by using it at the warning you triggered, which is in inst/include/nimble/RcppNimbleUtils.h for testing purposes.  Then we can build it into generated run-time checks ("assertions" in the exprClass objects).  I note that in the particular case you triggered, we could probably provide correct treatment for wrong number number of dimensions or wrong scalar type, but at the time we preferred not to.

fritzo:
Fixed by #297

PR #297, opened 2017-04-20
Add NIM_ASSERT macro and use it in RcppNimbleUtils.h

fritzo:
This introduces a C++ macro `NIM_ASSERT(-,-)` that can replace existing `if(-) PRINTF(-)` with actual calls to `error()`, fixing #296 .

## What's changed
The old `if(-) PRINTF(-)` behavior simply warned the user that an error may be encountered, and then optimistically continued execution. This is problematic, in that the executed code might crash R and destroy user work, as in #296 . The new behavior immediately breaks execution, thereby avoiding crashing R.

## What remains the same
This PR intends to exactly preserve the circumstances under which NIMBLE errors. That is, the behavior-on-error does change, but there is no change to when an error happens. In particular, there is some inconsistent behavior across `bool`, `int`, `double` versions as visible in https://rpubs.com/fritzo/nimble_error_messages, but this PR does not attempt to change that behavior.

## What remains to be done
This PR currently uses `NIM_ASSERT(-,-)` only in `RcppNimbleUtils.h`, and subsequent PRs will replace other old-style `if(-) PRINTF` with the new stricter macro.

### DIFF
--- a/packages/nimble/inst/CppCode/eigenUsingClasses.cpp
+++ b/packages/nimble/inst/CppCode/eigenUsingClasses.cpp
@@ -7,20 +7,21 @@
 
 template<>
 void SEXP_2_NimArr<1>(SEXP Sn, NimArr<1, double> &ans) {
-  if(!(isNumeric(Sn) || isLogical(Sn))) PRINTF("Error: SEXP_2_NimArr<1> called for SEXP that is not a numeric or logical!\n");
+  NIM_ASSERT(isNumeric(Sn) || isLogical(Sn),
+    "SEXP_2_NimArr<1, double> called for SEXP that is not a numeric or logical: actual type %s\n",
+    type2str(TYPEOF(Sn)));
   int nn = LENGTH(Sn);
-  if(ans.size() != 0) PRINTF("Error: trying to reset a NimArr that was already sized\n");
+  NIM_ASSERT(ans.size() == 0, "trying to reset a NimArr that was already sized\n");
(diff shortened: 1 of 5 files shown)


fritzo:
OK, I've fixed this up and manually tested in
  https://rpubs.com/fritzo/nimble_error_messages
I'd still like to add a couple unit tests before merging.

packages/nimble/inst/include/nimble/Utils.h
```diff
@@ -25,9 +25,18 @@ class nimbleTimerClass_ {
   double endNimbleTimer() { return(((double)(clock() - t_start))/CLOCKS_PER_SEC); }
 };
```
fritzo: It's pretty common for `ASSERT` macros to use `unlikely(-)` as a hint to the compiler that there will probably not be an error.

packages/nimble/inst/CppCode/eigenUsingClasses.cpp
```diff
@@ -7,44 +7,46 @@
```
fritzo: WDYT about moving these template specializations to their neighbors in `RcppNimbleUtils.h`, in a follow-up PR?

fritzo:
OK, I've added some unit tests, so I'm ready for anyone to hit Squash-and-Merge.

fritzo:
@perrydv May I Squash-and-Merge now?

perrydv:
Looks good to me.

PR #305, opened 2017-04-26
trap model[[nodeVector]]. trap invalid returnSizeArgID when assignment by first arg.  avoid eigen-casting of scalars amid non-scalar expressions.

perrydv:
This PR addresses issues 301, 303, 304.


perrydv:
@paciorek this passed TravisCI testing.  I suggest you take a look at run the reproducible examples from the 3 cited Issues.

paciorek:
LGTM

PR #310, opened 2017-04-28
addition of "sampler assignment rules" system into MCMC

danielturek:
Introducing a new system for handling the assignment of MCMC sampling algorithms, called samplerAssignmentRules.  This provides flexible and modifiable control of how `configureMCMC()` chooses what samplers to assign to what nodes, which is controlled by ordered set of assignment rules.  The default set of rules is contained in system option `MCMCdefaultSamplerAssignmentRules`, which can be modified or restored.  The default behaviour is identical to previous.

Documentation on this system is available in the roxygen documentation in `MCMC_configuration.R`, or in the User Manual (already rebuilt).


danielturek:
Merging #310

PR #312, opened 2017-04-28
Add assertthat to Imports: section of DESCRIPTION

fritzo:
This adds the [assertthat](https://cran.r-project.org/web/packages/assertthat/index.html) package in preparation for adding many more assertions to our codebase.

## Why?
The `asserthat` package makes it easier to add preconditions and postconditions in code. In particular, it encourages custom assertions rather than verbose `if(-)stop(-)` syntax, which improves readability.

Roughly half of NIMBLE's issues appear to be of the form "poor error reporting in circumstance ___". I believe NIMBLE could benefit from having many more assertions of pre- and post-conditions to detect errors closer to where they occur. I've personally wasted half of today diagnosing an error that was detected only distantly from where it occurred.

## Caveats
This package is still beta (version 0.2.0).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ install:
 ##  - Rscript -e "install.packages('devtools', repos = 'http://cran.us.r-project.org')"
   - Rscript -e "install.packages('igraph', repos = 'http://cran.us.r-project.org')"
   - Rscript -e "install.packages('coda', repos = 'http://cran.us.r-project.org')"
+  - Rscript -e "install.packages('assertthat', repos = 'http://cran.us.r-project.org')"
   - Rscript -e "install.packages('testthat', repos = 'http://cran.us.r-project.org')"
   - Rscript -e "install.packages('mvtnorm', repos = 'http://cran.us.r-project.org')"  ## needed for test-distributions.R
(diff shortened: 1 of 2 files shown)


paciorek:
Yes, I've experienced this a lot when I write nimbleFunctions and models. I
think we may want to have a focused meeting where we try to think about a
general strategy for better detection and reporting of errors.

I made some efforts in this direction when I added a bunch of checking of
dimensionality for models. Would be nice to have more checking / reporting
on nimbleFunctions too.

fritzo:
@paciorek Are you ok with adding this dependency?

I'd also be happy to give a 10-minute lecture on [Hoare logic](https://en.wikipedia.org/wiki/Hoare_logic) and assertions during some team meeting. It really improves health if everyone makes a habit of write assertions, so that ~15% of all code is assertions.

paciorek:
Yes, makes sense to add the dependency, if you're convinced that particular
package is the best way we see right now to add this functionality (I'm
just not sure offhand if there are competitor approaches...)

danielturek:
Interested as well.  Hugely in favor of defensive programming.

fritzo:
FYI I'm holding off on merging this until we can verify that `assert_that()` does not add too much computational overhead, as suggested by @perrydv .

perrydv:
And I think some other error-trapping strategy aspects are worth discussion as well.

PR #316, opened 2017-05-01
flat priors functionality (dflat, dhalfflat)

paciorek:


paciorek:
For dinvgamma, I was just mimicing what dgamma in R does. Hadn't really
thought through why it's set up that way. Note that dinvgamma is not part
of this PR; it's already in 0.6-4.

Good point about the as.numeric; I'm not sure why I put that in there...

perrydv:
LGTM (so go ahead and merge)

PR #318, opened 2017-05-03
Support optim() in nimbleFunctions

fritzo:
**Status:** Ready to merge

This PR removes Cliff's old partial implementation of `optim()` support and adds a simpler, more limited implementation that outputs a `nimbleList`. See the [Design Document](https://github.com/nimble-dev/nimble/wiki/Design-Doc:-support-optim()-and-optimize()#task-summary) for details.


perrydv:
This looks good to me.  Only comment is that some of the removed stale content of optimTools.R may turn out to be useful during next steps, but it seems ok to remove it now and wait to see if or how we adapt any pieces from it in the future.

perrydv:
Ok if I remove the old "Optim-Branch"?

fritzo:
It's fine with me if you remove the old Optim-Branch branch.

packages/nimble/R/genCpp_sizeProcessing.R
```diff
@@ -991,6 +1000,23 @@ sizeNimbleListReturningFunction <- function(code, symTab, typeEnv) {
...
 }
 
+sizeOptim <- function(code, symTab, typeEnv) {
+    # TODO input an optional nimbleList for control
+    asserts <- nimble:::recurseSetSizes(code, symTab, typeEnv)
+    code$type <- 'nimbleList'
+    code$sizeExprs <- symTab$getSymbolObject('OptimResultNimbleList', inherits = TRUE)
+    code$toEigenize <- 'maybe'
+    code$nDim <- 0
+    asserts <- c(asserts, sizeInsertIntermediate(code$caller, code$callerArgID, symTab, typeEnv))
```
perrydv: This line lifts the expression out of whatever was calling it.  Typically we do this only `if(!(code$caller$name %in% assignmentOperators))`.

packages/nimble/inst/CppCode/nimOptim.cpp
```diff
@@ -0,0 +1,51 @@
...
+    return ((NimObjectiveFn*)fn)(nim_par);
+}
+
+// This attempts to match the behavior of optim() defined in the documentation
+//   https://stat.ethz.ch/R-manual/R-devel/library/stats/html/optim.html
+// and in the reference implementation
+//   https://svn.r-project.org/R/trunk/src/library/stats/R/optim.R
+//   https://svn.r-project.org/R/trunk/src/library/stats/src/optim.c
+//   https://svn.r-project.org/R/trunk/src/include/R_ext/Applic.h
+nimSmartPtr<OptimResultNimbleList> nimOptim(NimArr<1, double>& par,
```
perrydv: We could probably be a bit more general if we take the parameter vector as a NimArrBase<double>, meaning we'll access high-dimensional objects in column-major order as vectors.  One could argue it is preferable to always require it to be 1-dimensional for clarity.  Not very important - just mentioning it.

packages/nimble/inst/CppCode/nimOptim.cpp
```diff
@@ -0,0 +1,51 @@
...
+                                            NimObjectiveFn fn) {
```
perrydv: will the `NimObjectiveFun` type need to be different when fn includes additional parameters?

packages/nimble/R/genCpp_sizeProcessing.R
fritzo: Oh, what is the correct behavior when `code$caller$name %in% assignmentOperators`?

BTW I just copied this from `sizeNimbleListReturningFunction` above, which implements `nimSvd` and `nimEigen`. I guess we should fix that code too.

packages/nimble/inst/CppCode/nimOptim.cpp
```diff
@@ -0,0 +1,51 @@
...
+    const int n = par.dimSize(0);
+
+    nimSmartPtr<OptimResultNimbleList> result = new OptimResultNimbleList;
+    result->par = par;
+    result->counts.setSize(2);
+    result->hessian.setSize(n, n);
+
+    // Use Nelder-Mead by default.
+    double* Bvec = par.getPtr();
```
perrydv: This is not completely safe.  If `parMatrix` is n-x-n and the user wants to provide the 3rd row as the initial parameters to optim, they can give `par = parMatrix[3,]`.  NIMBLE would turn this into a NimArr that has map=TRUE.  This treats its contents as a vector of length n but accesses the contents using a stride of n.  When we need to ensure a NimArr argument coming into a C++ function uses contiguous memory (e.g. to pass it along to nmmin), we have used a `nimArrCopyIfNeeded` step.  Example here:
https://github.com/nimble-dev/nimble/blob/devel/packages/nimble/inst/CppCode/nimDists.cpp#L42

packages/nimble/inst/CppCode/nimOptim.cpp
fritzo: Yes, I plan to expand the `NimObjectiveFunction` type in subsequent PRs. I believe we'll only need one additional parameter (that packs all other arguments), but in any case this general shape of plumbing should work.

packages/nimble/inst/CppCode/nimOptim.cpp
fritzo: Let's discuss intended usage. The `NimArr` code is quite difficult to read, so I've been guessing at how to use it. Later I'd like to either clean up that code or replace it with pure `Eigen`.

fritzo: Also, I'd like to add an example like yours to the unit test suite. I didn't realiize R's `optim()` could operate on matrices or slices/maps.

packages/nimble/R/genCpp_sizeProcessing.R
perrydv: Usually the situation is if we have a function like `nimOptim` that we can code-generate only for a direct assignment, e.g. `result <- nimOptim(...)`.  In that case `code$caller$name %in% assignmentOperators` and no modification is needed.  But if a user has written `result <- foo(nimOptim(...))`, we need to transform that to two lines: `Interm1 <- nimOptim(...)` and `result <- foo(Interm1)`.  So when size processing `nimOptim`, we can see that its caller is not an assignment operator and perform the `sizeInsertIntermediate` step, which creates the two lines described.  Summary: if the caller is already an assignment operator, no step is needed.

I saw that other case and figured you had just imitated it.  I already ran into that and fixed it on branch nimbleList-RCfun, so you could either fix it on your branch or wait to pick up the change later.

fritzo: Oh perfect! Thanks for pointing me to your branch. I'll grab your changes and incorporate them into this PR.

packages/nimble/inst/CppCode/nimOptim.cpp
perrydv: Let's talk about the `NimArr` system.  I think it is ripe for an overhaul, but we should be be cautious about priorities  and time commitments.   I think if we discuss in person we can decide how to handle the issue.

packages/nimble/inst/CppCode/nimOptim.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <limits>
+#include <nimble/nimOptim.h>
+
+// This wraps a NimObjectiveFn as an R `optimfn` type
+//   typedef double optimfn(int, double *, void *);
+// defined in
+//   https://svn.r-project.org/R/trunk/src/include/R_ext/Applic.h
+static double optimfn_wrapper(int n, double* par, void* fn) {
+    NimArr<1, double> nim_par;
+    nim_par.setSize(n);
```
perrydv: It is slightly more efficient to use `nim_par.setSize(n, false, false)`, which makes no initialization be done.  I would have never thought it would be noticeable, we saw in MCMCs that unnecessary initialization (at the time we used std::vector as the underlying memory allocation, which always initializes) took up a meaningful amount of computation time when repeats zillions of times.

perrydv:
I’m not sure R’s optim will operate on a matrix — that was a thought that should be checked.

For a slice, R would simply copy it using its pass-by-copy semantics and it would not longer be a slice.  The issue inside our C++ system is we do not follow pass-by-copy semantics.

packages/nimble/inst/CppCode/nimOptim.cpp
fritzo: Good point, since this is called the inner loop of optimization! Done.

fritzo:
OK Perry, I'm pretty happy with the state of this PR. `nimOptim` now behaves almost exactly the same in R and C++, and I've used your `assign(..., envir = .GlobalEnv)` trick to fix the scoping issues in the tests.

danielturek:
Out of curiosity, what are the differences in the "almost exact" behaviour between R and C++ executions?  And, when you refer to R behaviour, do you mean R's native optim(), or R execution of optim() in the NIMBLE DSL?

fritzo:
@danielturek The only difference between `nimOptim()` in R and `nimOptim()` in C++ is an NA in one position of an R vector, whereas the C++ version returns a garbage integer (do we reserve any integer for NA value?). There are more differences between `optim()` in R and `nimOptim()` in R, due to the limitations of NA and NULL handling in `nimbleList`s:

- `optim()` returns a `list`, whereas `nimOptim()` returns an `optimResultNimbleList`
- `optim()` sets some of its list values to `NULL` whereas `nimOptim()` sets those to default values (an empty string, a 0x0 matrix).

fritzo:
@perrydv Could you PTAL at the [latest changes](https://github.com/nimble-dev/nimble/pull/318/files/8253d8466b20287d044c08c25e5b2fcef686c556..038b4083e1c011a0768ad4a0a2d8e9e435c4b4d6) where I've converted `nimFun$run` to `NimBind(&NimFun::run, nimFun)` in `genCpp_generateCpp.R`? This seemed like an appropriately late state in compilation to add this processing, since `nimFun$run` seems like a natural object (a bound method) and we just need to perform a little C++ gymnastics to handle it.

perrydv:
@fritzo I think the places you've done things look good.  We  haven't used `processSpecificCalls` purely for error-trapping before, but it makes sense to do so.

fritzo:
@NLMichaud I just added [a script](https://github.com/nimble-dev/nimble/pull/318/files#diff-09de097a71ef6a555460cc6abc5ef924) to generate our static C++ code for `OptimResultNimbleList`. Could you please take a look and review for sanity?

I've automated this so that it's easier to generate additional static C++ code, e.g. an `OptimControlNimbleList` that will add soon.

fritzo:
As we discussed at the team meeting, I'll merge this as soon as CI tests pass.

PR #322, opened 2017-05-03
Rename neededRCfuns field to NeededTypes

fritzo:
This change was recommended by @perrydv .  Perry is this what you intended?


perrydv:
Yes, that was the idea, so it can be used more generically like the neededTypes in the nfProcessing class.  But let's discuss to get our workflow clear.

fritzo:
Closing since there is already a field named `neededTypes`, and I don't understand what this PR intends to do.

perrydv:
Makes sense to close this.  But the neededTypes exists in nfProcessing and neededRCfuns in RCfunProcessing, so they're in different places.  nfProcessing has several different types it might need, hence the generic label "neededTypes".  In the past, RCfunProcessing would only need other RCfuns, hence the narrower name neededRCfuns.  now (soon) RCfuns can also need nimbleList types, hence the suggested name change.  However it is a very minor point and most naturally dealt with in the nimbleList modifications going on right now.

PR #323, opened 2017-05-03
Log actual error and traceback when size processing fails

fritzo:
Our current `try()...stop()` construction allows us to show the nimbleCode context of where an error occurred, but it also obfuscates the lower level details of the error. This PR adds some of that lower-level info back in by replacing `try(...)` with `tryCatch(withCallingHandlers(...), ...)`.

As we gain users, we'll want to include as much context information as possible, so that when they paste errors to nimble-users we have a better chance at diagnosing the error.

**Before this PR:**
```
Error: There is some problem in the the setSizes processing step for this code:
 {
    return(nimFakeOptim(par = par, fn = example$fn))
}
```

**After this PR:**
```
Error: There was some problem in the the setSizes processing step for this code:
{
    return(nimFakeOptim(par = par, fn = example$fn))
}
Internal Error:
Error in if (!(isSymFunc || isSymList)) stop(exprClassProcessingErrorMsg(code, : missing value where TRUE/FALSE needed

Traceback:
29: compileNimble(nimFun, showCompilerOutput = TRUE, dirName = "~/tmp")
28: project$compileNimbleFunctionMulti(units[whichUnits], control = control, reset = reset, showCompilerOutput = showCompilerOutput)
27: compileNimbleFunction(funList[thisBool], isNode = isNode, filename = filename, initialTypeInferenceOnly = initialTypeInferenceOnly, control = control, reset = reset, returnCppClass = returnCppClass, where = where, fromModel = fromModel, generatorName = uGN, alreadyAdded = alreadyAdded, showCompilerOutput = showCompilerOutput)
26: buildNimbleFunctionCompilationInfo(funList, isNode = isNode, initialTypeInferenceOnly = initialTypeInferenceOnly, control = control, where = where, fromModel = fromModel)
25
```

### DIFF
--- a/packages/nimble/R/RCfunction_compile.R
+++ b/packages/nimble/R/RCfunction_compile.R
@@ -208,11 +208,20 @@ RCfunProcessing <- setRefClass('RCfunProcessing',
                                        compileInfo$typeEnv[['.ensureNimbleBlocks']] <<- FALSE ## will be TRUE for LHS recursion after RHS sees rmnorm and other vector dist "r" calls.
                                        passedArgNames <- as.list(compileInfo$origLocalSymTab$getSymbolNames()) 
                                        names(passedArgNames) <- compileInfo$origLocalSymTab$getSymbolNames() 
-                                       compileInfo$typeEnv[['passedArgumentNames']] <<- passedArgNames ## only the names are used.  
-                                       tryResult <- try(exprClasses_setSizes(compileInfo$nimExpr, compileInfo$newLocalSymTab, compileInfo$typeEnv))
-                                       if(inherits(tryResult, 'try-error')) {
-                                           stop(paste('There is some problem at the setSizes processing step for this code:\n', paste(deparse(compileInfo$origRcode), collapse = '\n'), collapse = '\n'), call. = FALSE)
-                                       }
+                                       compileInfo$typeEnv[['passedArgumentNames']] <<- passedArgNames ## only the names are used.
+                                       tryCatch(withCallingHandlers(exprClasses_setSizes(compileInfo$nimExpr, compileInfo$newLocalSymTab, compileInfo$typeEnv),
+                                                                    error = function(e) {
+                                                                        stack <- sapply(sys.calls(), deparse)
+                                                                        .GlobalEnv$.nimble.traceback <- capture.output(traceback(stack))
+                                                                    }),
+                                                error = function(e) {
+                                                    stop(paste('There was some problem in the the setSizes processing step for this code:',
+                                                               paste(deparse(compileInfo$origRcode), collapse = '\n'),
+                                                               'Internal Error:', e,
+                                                               'Traceback:', paste0(.GlobalEnv$.nimble.traceback, collapse = '\n'),
+                                                               sep = '\n'),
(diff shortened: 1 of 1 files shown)


perrydv:
I like this.  In the future it can be combined with the idea of tracking where in the AST we are so the problem expression can be pinpointed more accurately, but I think that will be a reasonably independent extension of this.  

Is it redundant to wrap withCallingHandlers inside tryCatch?

fritzo:
Good question! Is it not redundant to wrap `withCallingHandlers()` inside `tryCatch()`, since the former executes the error callback deep in the stack (where we want to call `sys.calls()`), and the latter executes the calling context at the shallower stack location (where we want to examine the `nimbleCode`). I'm adopting this pattern from a similar [use case on stack overflow](http://stackoverflow.com/questions/15282471/get-stack-trace-on-trycatched-error-in-r).
-

fritzo:
@perrydv May I merge this?

perrydv:
Yes.  LGTM.

danielturek:
👍

perrydv:
I have a follow-on question about the PR (showing `traceback()`s when trapping errors during `compileNimble`).  I like the point that a user can send a more meaningful Issue or nimble-users message if they provide the richer error output.  But what about when we do nicely catch a programming error that they should go fix in their code.  Do we then want to avoid showing all the traceback() output?

fritzo:
That sounds reasonable. Although as an internal developer, I'd prefer to usually set something like `nimbleOptions(alwaysShowTraceback = TRUE)`.

Actually, does anyone else find it cumbersome to work around our error catching logic? I often wish I could just use `options(error = recover)` instead of manually finding and stepping through `debug(-)`. Maybe we could add a setting `nimbleOptions(catchErrors = FALSE)` where we would use `tryCatch(-, finally = -)` construct to avoid eating the errors?  I think the following could both report on the code (as you originally intended) and allow use of `options(error = recover)`:
```
succeeded <- FALSE
tryCatch({ functionThatMayErr() ; succeeded <- TRUE}, finally = function() {
    if(!succeeded) {
        ...do whatever error reporting we currently do now...
    }
})
```

perrydv:
To clarify: Is the issue that in size processing the errors we trap and call `stop` on end up landing us back in the `try()` region of `RCfunProcessing` even if `options(error = recover)` was set?

fritzo:
Correct.

fritzo:
wait, sorry, I must have been doing something wrong. `options(error = recover)` seems to be working now :blush:

perrydv:
I'd have to play with your suggestion to see how it would help.  We already wrap a `try()` around the main entry point `exprClasses_setSizes` and then catch individual errors via `stop`.  Oh, just saw your newest message come through.  Great.   I was going to suggest we could have a lightweight wrapper to `stop` or (`stopifnot`) that would do this:

```{r}
nimbleDevelStop <- function(msg, browse) {if(browse) evalq(browser(), envir = parent.frame())}
```
where `browse` could instead be obtained from a nimbleOption.  This would let us set a flag that always throws us into a browser at the point of failure.  I like that, but maybe it's what you didn't want.

PR #330, opened 2017-05-09
Attempt to fix broken calls to internal methods of roxygen2

fritzo:
- Renames `prep_pkg` to `prep_pkg.R` (so I can debug it line-by-line in RSudio)
- Updates our use of the internal `roxygenize2:::roc_output` to `roxygenize:::roclet_output`
- Replaces `roxygenise()` with `devtools::document()` to work around https://github.com/klutometis/roxygen/issues/595

This fixes a few bugs, but I still haven't been able to generate documentation :-(

### DIFF
--- a/packages/prep_pkg.R
+++ b/packages/prep_pkg.R
@@ -69,7 +69,7 @@ for(file in files) {
   print(file)
   test <- try(a <- roxygen2:::roc_proc_text(roclet, paste(readLines(file.path(Rpath, file)), sep="\n")))
   if(!is(test, 'try-error'))
-    roxygen2:::roc_output(roclet, a, "nimble", options = list(wrap=FALSE, roclets = "rd"), check = FALSE)
+    roxygen2:::roclet_output(roclet, a, "nimble", options = list(wrap=FALSE, roclets = "rd"), check = FALSE)
 }
 
 
@@ -138,7 +138,7 @@ cat(paste(text, collapse = "\n"), file = file.path("nimble", "man", "nimble-math
 # now create real NAMESPACE file
 ## NAMESPACE
 system(paste("mv", file.path('nimble','NAMESPACE'), file.path('/tmp', 'NAMESPACE')))
-roxygenise('nimble','namespace')
+devtools::document('nimble', roclets = c('namespace'))
 
 namespace <- readLines(file.path('nimble','NAMESPACE'))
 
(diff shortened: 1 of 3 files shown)


paciorek:
Glad you're looking into this. There's a lot of stuff just cobbled together
in prep_pkg to just make stuff "work" on my Linux machine. Wouldn't mind
revisiting it, but maybe the time to do that is when we are putting
together the 0.6-5 final packaging in a week or so. But if you'd like to be
able to generate documentation yourself sooner than that, I'm happy to sit
down with you and see what is going wrong on your machine.

I should note that Makefile.nimble is very out-of-date and not expected to
work...

fritzo:
Thanks, I eventually got documentation kind of working in RStudio, so at least I can proofread my updates :-) I gave trying to use prep_pkg, but hopefully these fixes will help whomever uses it next.

fritzo:
Should I also remove Makefile.nimble in this PR. It would be great to avoid further maintenance cost by removing any unused utilities like that.

paciorek:
how about you and I sit down and talk through changes to the testing
(minimal vs. full testing) in the context of the larger workflow of
building the full package (with roxygen re-creation). This is a good time
to do it since I am about to build for 0.6-5. So maybe we can sit down
later this week?

fritzo:
Sounds good to me. My schedule is open. How about Wednesday?

packages/prep_pkg.R
```diff
@@ -138,7 +138,7 @@ cat(paste(text, collapse = "\n"), file = file.path("nimble", "man", "nimble-math
 # now create real NAMESPACE file
 ## NAMESPACE
 system(paste("mv", file.path('nimble','NAMESPACE'), file.path('/tmp', 'NAMESPACE')))
-roxygenise('nimble','namespace')
+devtools::document('nimble', roclets = c('namespace'))
```
fritzo: @paciorek suggests avoiding a dependency on `devtools`

paciorek: I'm merging this but am then going to do some revamping of prep_pkg.R to more cleanly use roxygen2. 

Note that we'll need to use roxygen2 v5.0.1 because of issues with v6.0.1 noted earlier in this thread (I also had problems with devtools::document, though we could revisit that and see if that would work instead of falling back to older roxygen2...)

Issue #327, opened 2017-05-09
Using nodeNames[i] when nodeNames has length 0 ?


danielturek:
Is there any way we could make things like:
`model$getParam(nodeNames[i], "mean")`

get through the compiler, when `nodeNames` has length 0?  This would be helpful / necessary for true model-generic programming.

This may also extend for other functions using `nodeNames[i]`, e.g.:

`model$calculate(nodeNames[i])`

although I haven't tried.  Fully reproducible example:

```
library(nimble)

code <- nimbleCode({  a ~ dnorm(0, 1)  })
Rmodel <- nimbleModel(code)

nfDef <- nimbleFunction(
    setup = function(model, nodeNames) { },
    run = function() {
        a <- 0
        for(i in seq_along(nodeNames))
            a <- a + model$getParam(nodeNames[i], "mean")
    }
)

Rnf <- nfDef(Rmodel, character(0))

Cmodel <- compileNimble(Rmodel)
Cmcmc <- compileNimble(Rnf, project = Rmodel)
compiling... this may take a minute. Use 'showCompilerOutput = TRUE' to see C++ compiler details.
Error in makeParamInfo(model, nodeNames, "mean") : 
  cannot have an indexed vector of nodes used in getParam if they have different types or dimensions for the same parameter.
```

perrydv:
This looks feasible to implement.  `model$getParam` is fundamentally different than other model member functions during compilation because we must determine the type of the parameter and require that all nodes in `nodeNames` have the same type for "mean", e.g.  

In the case that `length(nodeNames)==0`, we cannot determine the type of "mean".  However, if this occurs in one of multiple instances, we can determine the type of "mean" from other instances, and the assumption/requirement is that it will always be consistent.

A corner case would be if all instances considered (like if there is just 1) have `length(nodeNames)==0`.  In that case we could either throw an error or default to type `double()`

fritzo:
Hi Daniel, do I understand correctly that this is kind of a mismatch between R's optimisic `seq_along(-)` function and `-[-]` indexing (which know how to flatten out any array to a linear array) and the C++'s pedantic indexing? If so then maybe we could do this cleanly with a `nimFlatten` operation that provided an `nDim=1` view of an array with arbitrary `nDim`:
```{r}
    ...
    for(i in seq_along(nodeNames))
            a <- a + model$getParam(nimFlatten(nodeNames)[i], "mean")
    ...
```
or maybe we could do this internally.

[...]

PR #331, opened 2017-05-09
support for model$getParam(nodes[i], 'mean') if nodes is empty

perrydv:
@danielturek This fixes Issue #327.  If there are no instances with a non-empty `nodes`, it defaults to assigning a return type of double().  If subsequent compilations of the same nf are not compatible with that type, there will be an error.  But at least it works safely as you’ve requested in the Issue.


danielturek:
@perrydv Fantastic, thanks.  LGTM.

PR #336, opened 2017-05-12
Expunge obsolete optim code from codebase

fritzo:
This makes way for a clean reimplementation of `optim` in #318


perrydv:
I'm not seeing where this broke in testing.  the travis-CI report ends early but where is the failure?

fritzo:
I think I accidentally cancelled the build. It looks like github doesn't display the difference between CANCEL and FAIL.

perrydv:
Oh that makes sense.

perrydv:
Given that these are simple changes (removed old code) and testing runs longs, LGTM and we could go ahead with merge if we want to move quickly today.

fritzo:
OK @perrydv , we should now be ready to merge `devel` into our branches.

perrydv:
will do

PR #342, opened 2017-05-15
system option to toggle use of MCMC sampler assignment rules (default FALSE)

danielturek:
Added a system option to toggle use of the sampler assignment rules system.

Default is FALSE, that is, to use the older, slightly faster, static system.

Do not merge.  Waiting to make sure tests pass.


paciorek:
looks good as far as I can tell; I haven't looked into the rule evaluation code from the previous commit so don't fully have my head wrapped around this.

danielturek:
The best, albeit very brief, introduction, is the example provided in the roxygen documentation for `samplerAssignmentRules`, which appears in MCMC_configuration.R

perrydv:
LGTM.  Only small comment is the option name `useSamplerAssignmentRules` might confuse a new code reader.  Either pathway uses a set of rules.  The difference is in how they are set up.  Maybe `useDynamicSamplerAssignmentRules` ?  Ok either way.

perrydv:
@danielturek there is still message if one says `RWonly = TRUE` warning that it will be deprecated.  Is this part of the new system, and if so should we hold off on such warnings?

danielturek:
I was going to leave the message in there, but if you want to entirely remove the option (and the resulting warning message) that's ok with me.

PR #343, opened 2017-05-15
inverse wishart

paciorek:
newDists has inverse wishart, which I'd like in 0.6-5. Could Daniel or Perry take a look and let me know if it looks good?


packages/nimble/R/genCpp_operatorLists.R
```diff
@@ -103,7 +103,7 @@ brackOperators <- list('[' = c('[',']'), '[[' = c('[[',']]'))
 
 ## see distributions_processInputList for some relevant lists of distributions functions 
 
-callToSkipInEigenization <- c('copy','setValues', 'setValuesIndexRange', 'getValues', 'getValuesIndexRange', 'setSize', 'resize', 'getsize', 'size', 'resizeNoPtr','assert', 'return', 'blank', 'rankSample', 'nimArr_dmnorm_chol', 'nimArr_dmvt_chol', 'nimArr_dwish_chol', 'nimArr_dmulti', 'nimArr_dcat', 'nimArr_dinterval', 'nimArr_ddirch', 'nimArr_rmnorm_chol', 'nimArr_rmvt_chol', 'nimArr_rwish_chol', 'nimArr_rmulti', 'nimArr_rcat', 'nimArr_rinterval', 'nimArr_rdirch', 'calculate', 'calculateDiff', 'simulate', 'getLogProb', 'nimEquals', 'startNimbleTimer', 'endNimbleTimer')
```
fritzo: nit: This line would be easier to review if it were split into multiple lines, e.g.
```
callToSkipInEigenization <- c(
    'copy',
    'setValues',
    'setValuesIndexRange',
    'getValues',
    'getValuesIndexRange',
   ...)
```

packages/nimble/R/registration.R
```diff
@@ -17,6 +17,8 @@ if(!exists('.NimbleUseRegistration') || !.NimbleUseRegistration) {
   ## R_setSinglePtrFromSinglePtr = 'setSinglePtrFromSinglePtr'
    C_dwish_chol = 'C_dwish_chol'
    C_rwish_chol = 'C_rwish_chol'
+   C_dinvwish_chol = 'C_dinvwish_chol'
```
fritzo: (Unrelated to this review:) Is this registration needed only so that `dinvwish_chol` can be used in uncompiled nimbleFunctions? I ask because I have not added an entry for `nimOptim` since we use R's builtin `optim`. I just want to make sure we don't need an additional `C_nimOptim`.

paciorek: C_dinvwish_chol is needed as the entry point for R functions to make use of dinvwish_chol. In particular, the dinvwish_chol function in R calls C_dinvwish_chol, which then dispatches to the core C++ function for dinvwish_chol. 

So I don't think you need a C_nimOptim, since R's optim will itself call the needed C++ code.

packages/nimble/inst/CppCode/dists.cpp
```diff
@@ -271,6 +279,241 @@ SEXP C_rwish_chol(SEXP chol, SEXP df, SEXP scale_param)
...
+  if(!overwrite_inputs)
+    delete [] cholCopy;
+}
+
+SEXP C_rinvwish_chol(SEXP chol, SEXP df, SEXP scale_param) 
+// generates single Inverse Wishart draw given Cholesky of scale or rate matrix
+// Cholesky matrix should be given as a numeric vector in column-major order
+//   including all n x n elements; lower-triangular elements are ignored
+{
+  if(!isReal(chol) || !isReal(df) || !isReal(scale_param))
```
fritzo: nit: what do you think of using `NIM_ASSERT` here:
```
NIM_ASSERT(isReal(chol), "Error (C_rinvwish_chol): wrong type for argument chol");
NIM_ASSERT(isReal(df), "Error (C_rinvwish_chol): wrong type for argument df");
NIM_ASSERT(isReal(scale_param), "Error (C_rinvwish_chol): wrong type for argument scale_param");
```

packages/nimble/R/registration.R
fritzo: Thanks! deleting comment.

packages/nimble/inst/CppCode/dists.cpp
paciorek: I haven't followed the whole discussion in terms of revision of our error trapping. If NIM_ASSERT is what we've agreed on, then this is fine by me.

I would like to have a bit of discussion amongst all of us in-person as to the plan for error trapping going forward so we all know and are consistent (and ideally go back and revise current error trapping).

Fritz, perhaps you can remind us to discuss this next Monday?

fritzo: Sure I'll bring up error trapping for discussion next Monday.

perrydv:
LGTM. I haven't gone through the math details but I'm sure you've got some tests.

PR #344, opened 2017-05-15
Move test invocations from .travis.yml to test-devel.sh

fritzo:
This moves the list of test files from .travis.yml to a new script test-devel.sh. This has two advantages:
- In the future we can split test-devel.sh into an expensive test-master.sh and a cheaper test-devel.sh
- When we get [appveyor branch](https://github.com/nimble-dev/nimble/tree/appveyor) working, we can share test-devel.sh between both .travis.yml and appveyor.yml (so that new tests need only be added in onc place.


fritzo:
@paciorek Do you mind if I merge this in preparation for our discussion later this week?

danielturek:
LGMT

fritzo:
Thanks, Daniel!

PR #345, opened 2017-05-16
Add option to quiet down errors

fritzo:
This quiets down the verbose error messaging introduced in #323 and gates it behind a global option.


perrydv:
LGTM

Issue #341, opened 2017-05-15
Error in getDependencies(..., self=FALSE) for components of multvariate nodes


danielturek:
`model$getDependencies(...)` doesn't respect the `self=FALSE` option, when getting dependencies of a multivariate node, or those of a scalar component of a multivariate node.

This cannot be remedied using `returnScalarComponents = TRUE`, as I thought used to work.

See examples below, which also shows "what I would expect" to see.

```
library(nimble)
code <- nimbleCode({
    x[1:4] ~ dmnorm(mu[1:4], C[1:4,1:4])
    y ~ dexp(x[1])
})
constants <- list(mu = rep(0,4), C = diag(4))
Rmodel <- nimbleModel(code, constants)

> Rmodel$getDependencies('x[1:4]')
##[1] "x[1:4]" "y"     

> Rmodel$getDependencies('x[1:4]', self=FALSE)
##[1] "x[1:4]" "y"      ------ I would expect: "y"

> Rmodel$getDependencies('x[1:4]', returnScalarComponents = TRUE)
##[1] "x[1]" "x[2]" "x[3]" "x[4]" "y"   

> Rmodel$getDependencies('x[1:4]', self=FALSE, returnScalarComponents = TRUE)
## [1] "x[1]" "x[2]" "x[3]" "x[4]" "y"   ------ I would expect: "y"

> Rmodel$getDependencies('x[1]')
## [1] "x[1:4]" "y"     

> Rmodel$getDependencies('x[1]', self=FALSE)
## [1] "x[1:4]" "y"     ------ I would expect: "y"

> Rmodel$getDependencies('x[1]', returnScalarComponents = TRUE)
## [1] "x[1]" "x[2]" "x[3]" "x[4]" "y"   

> Rmodel$getDependencies('x[1]', self=FALSE, returnScalarComponents = TRUE)
## [1] "x[1]" "x[2]" "x[3]" "x[4]" "y"    ------ I would expect: "y"
```

perrydv:
I'm guessing the `omit` argument also doesn't work (and is never used internally but should be supported).

[...]

PR #346, opened 2017-05-16
fix getDependencies with self=TRUE and with omit that includes input nodes

perrydv:
@danielturek This PR fixes Issue #341 .  See also new test-graphStructure.  This test works by comparing to a saved output file confirmed by inspection to give desired results.  Please check and add more cases as you see fit.


packages/nimble/inst/tests/graphStructureTestResults_Correct.Rout
```diff
@@ -0,0 +1,31 @@
+graph structure tests case 1
```
fritzo: Nice tests!

danielturek:
@perrydv I'm quite happy with the test cases you've included there.  Looks good to merge.

Issue #340, opened 2017-05-15
Error in model building: dependencies from multivariate nodes


danielturek:
The following code produces an error, and I don't believe it should.

This error does **not** seem to be new, since merging refactor-modelDef.  I went back before that merge, and the error occurs just the same.

Note this error does not occur if the second declaration depends on x[1]:  `y[1] ~ dnorm(x[1], 1)`

```
library(nimble)
code <- nimbleCode({
    x[1:4] ~ dmnorm(mu[1:4], C[1:4,1:4])
    y[1] ~ dnorm(x[2], 1)
})
constants <- list(mu = rep(0,4), C = diag(4))
Rmodel <- nimbleModel(code, constants)

## defining model...
## Error in genExpandedNodeAndParentNames3(debug = debug) : 
##   Error building model, there are multiple definitions for nodes:x[1:4]
```

perrydv:
This is a bug in the system that splits nodes into multiple vertices based on RHS expressions and determines the resulting vertex names.  Specifically `x[1:4]` should be split into `x[2]` and `x[1-s-4]` but determination of when the split is contiguous is incorrect and results in a vertex `x[1:4]` that duplicates the name of the original node `x[1:4]`.  That happens [here](https://github.com/nimble-dev/nimble/blob/e8cfb1240000fbfa1f4c8d70bdbf2c4c05f62ad1/packages/nimble/R/BUGS_modelDef.R#L1211)

More generally we would benefit from detailed unit tests of BUGS processing.  That has been a gap.

danielturek:
Thanks for the quick explanation.  This is something that can be fixed, then?

perrydv:
I think so, but it is a bit more involved than the other issue you just filed.  It would make sense to put in place detailed unit tests as part of fixing this bug.

perrydv:
Fixed by PR #347

PR #347, opened 2017-05-16
Fix graph splitting in middle of non-scalar nodes

perrydv:
@danielturek This fixes Issue #340.

* In makeVertexNamesFromIndexArray2, fixed handling of non-contiguous results of a split.
* added cases to test-graphStructure
* moved test utilities for saving a correct results text file and comparing to it line by line to test_utils.R
* removed vestigial makeVertexNamesFromIndexArray


perrydv:
Not for merging.  Will fail tests.  Is there a way to stop travis-CI when I know it will fail but I want to merge for sync'ing branch across machines?

danielturek:
If you mean travis running tests on the PR branch automatically, with
results shown with the PR, then no, I'm not aware of any way to disable
that.

fritzo:
@perrydv The best solution I know of is to manually cancel a travis test that has just started. To to that, navigate to the travis test progress page for a test, and click `X Cancel` in the upper right.

perrydv:
Thanks @fritzo .  For some reason I don't see the Cancel button, but no matter.  

If these tests pass, I am ready for code review on this.

Issue #337, opened 2017-05-12
multiple lifted nodes from identical code fails


paciorek:
I haven't investigated at all as I'm bogged down in some PalEON coding that revealed this:

```
code = nimbleCode({
for(j in 1:3) {  
      omega[j] ~ dexp(lambda[j]*lambda[j]/2)
      lambda[j] ~ dgamma(1,1)
  }
  for(j in (order+1):J) {  
      omega[j] ~ dexp(lambda[j]*lambda[j]/2)
      lambda[j] ~ dgamma(1, rate = sigma) 
  }
})

model = nimbleModel(code, constants = list(J=100, order =3))
model$getDependencies('lambda[4]')
[1] "lambda[4]"
```
Note that the dependencies of lambda[4] are incorrect. 

Things are fine if I use a distinct index:
```
code = nimbleCode({
for(j in 1:3) {  
      omega[j] ~ dexp(lambda[j]*lambda[j]/2)
      lambda[j] ~ dgamma(1,1)
  }
  for(k in (order+1):J) {  
      omega[k] ~ dexp(lambda[k]*lambda[k]/2)
      lambda[k] ~ dgamma(1, rate = sigma) 
  }
})

model = nimbleModel(code, constants = list(J=100, order =3))
model$getDependencies('lambda[4]')
[1] "lambda[4]"                                         
[2] "lifted_lambda_oBk_cB_times_lambda_oBk_cB_over_2[4]"
[3] "omega[4]"                                          
```

perrydv:
This occurs because the two `omega[j]` declarations create a lifted declaration with *identical* code.  But lifted declarations are meant to be unique by name (a mangling of the expression), and a declaration can have only one `context` (for-loop info).  This works well if say `rate = sigma` appears in multiple different for-loops, because only a *single* `d1_over_sigma <- 1 / sigma` declaration is lifted.  But in the example here, the second declaration that would lift `lambda[j] * lambda[j] / 2` does not do so because it appears identical to the first one, so it is not lifted again. The first one appears in the `j in 1:3` context, so the next context is effectively ignored for the lifted declaration 

[...]

PR #350, opened 2017-05-16
force lifted nodes that use for-loop indices to be uniquely named

perrydv:
This PR fixes Issue #337.  @paciorek see what you think (also note discussion in that issue thread)


fritzo:
Could you please merge the latest devel and rerun tests, to avoid false negatives due to #354 ?

perrydv:
@paciorek I'm pinging you again to see if you are ok with this solution (see Issue #337 for more)

Issue #351, opened 2017-05-17
rnorm(-) fails to compile inside return()


fritzo:
The following code fails:
```{r}
nimFun1 <- nimbleFunction(
    run = function() {
        return(rnorm(4))
        returnType(double(1))
    }
)
compiledFun <- compileNimble(nimFun1, showCompilerOutput = TRUE)  # This fails
```
with an error
```
P_45_rcFun_123.cpp:14:8: error: no matching function for call to 'Rf_rnorm'
return(Rf_rnorm(4, 0, 1));
       ^~~~~~~~
/usr/share/R/include/Rmath.h:394:8: note: candidate function not viable: requires 2 arguments, but 3 were provided
double Rf_rnorm(double, double);
       ^
```
A workaround is to manually create a temporary variable
```{r}
nimFun2 <- nimbleFunction(
    run = function() {
        x <- rnorm(4)
        return(x)
        returnType(double(1))
    }
)
compiledFun <- compileNimble(nimFun2, showCompilerOutput = TRUE)  # This works
```

perrydv:
Fixed by PR #355

PR #355, opened 2017-05-17
force lifting of any return argument with toEigenize=='yes'. add tests of return expressions in test-coreR.

perrydv:
This fixes Issue #351.  There was commented-out old logic forcing lifting of any return argument with toEigenize==‘yes’.  It had been replaced with different logic.  I reinstated the toEigenize check and combined with the other logic.

### DIFF
--- a/packages/nimble/R/genCpp_sizeProcessing.R
+++ b/packages/nimble/R/genCpp_sizeProcessing.R
@@ -368,7 +368,7 @@ sizeRecyclingRuleRfunction <- function(code, symTab, typeEnv) {
     code$sizeExprs <- newSizeExprs
     code$type <- 'double' ## will need to look up from a list
     code$nDim <- 1
-    code$toEigenize <- TRUE
+    code$toEigenize <- 'yes'
     return(asserts)
 }
 
@@ -2267,8 +2267,8 @@ isIntegerEquivalent <- function(code) {
 }
 
 sizeSeq <- function(code, symTab, typeEnv, recurse = TRUE) {
-    message('still need to handle -1L sequences')
-    message('check that arguments are scalars')
+    ##message('still need to handle -1L sequences')
+    ##message('check that arguments are scalars')
     asserts <- if(recurse) recurseSetSizes(code, symTab, typeEnv) else list()
     byProvided <- code$name == 'nimSeqBy' | code$name == 'nimSeqByLen'
     lengthProvided <- code$name == 'nimSeqLen' | code$name == 'nimSeqByLen'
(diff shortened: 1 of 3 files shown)


packages/nimble/inst/tests/test-coreR.R
```diff
@@ -716,6 +716,45 @@ logicalTests <- list(
          outputType = quote(double(2)))
 )
 
+returnTests <- list(
+    list(name = "return(rnorm scalar)",
```
fritzo: Thanks for fixing and adding a regression test!

packages/nimble/R/genCpp_sizeProcessing.R
```diff
@@ -368,7 +368,7 @@ sizeRecyclingRuleRfunction <- function(code, symTab, typeEnv) {
     code$sizeExprs <- newSizeExprs
     code$type <- 'double' ## will need to look up from a list
     code$nDim <- 1
-    code$toEigenize <- TRUE
```
fritzo: Gosh, was this a bug due to a type error? If we do separate out lifting-for-eigenization (as we discussed yesterday), it's probably worth checking the type in the one place where the field will be read.

perrydv: No, that was actually an incidental catch.  It wasn't the source of the bug in question.  BTW I think the global view of the toEigenize logic is a little more complicated that we discussed but still separable.

packages/nimble/R/genCpp_sizeProcessing.R
```diff
@@ -2560,11 +2560,13 @@ sizeReturn <- function(code, symTab, typeEnv) {
     asserts <- recurseSetSizes(code, symTab, typeEnv)
     if(inherits(code$args[[1]], 'exprClass')) {
         if(!code$args[[1]]$isName) {
-##            if(code$args[[1]]$toEigenize == 'yes') {
-            if(anyNonScalar(code$args[[1]])) {
+            liftArg <- FALSE
+            if(code$args[[1]]$toEigenize == 'yes')
```
fritzo: Do I understand correctly that if `code$args[[1]]$toEigenize == 'maybe'`, then the code will not be eigenized? Specifically, if a child node of the code has `toEigenize == 'yes'`, then we still handle this correctly (either by propagating `'yes'` up before we get here, or by deciding not to eigenize)?

fritzo:
LGTM after the `'maybe'` comment is addressed.

packages/nimble/R/genCpp_sizeProcessing.R
perrydv: There are two aspects to your question.  Yes, if `code$args[[1]]$toEigenize == 'maybe'`, the code will not be eigenized.  The canonical case would be `A + B` where both are scalars.  This is marked `maybe` because if it is part of `C + A + B` where `C` is a vector, then it is fine to include the scalar arithmetic along with the vector arithmetic.  But otherwise the `maybe` would not be eigenized because it can be handled without eigen.  `maybe` can be interpreted as "either".  

Regarding propagation of `toEigenize`: currently each size handler needs to determine if there are any conflicts and create an intermediate variable if necessary.  Example: if `foo` is an RCfunction and is used as `foo(A %*% b)`, the size handler for `foo` (which will be `generalFunSizeHandler`) is responsible for creating an intermediate for `A %*% b`.  

This is an example of what we've chatted about: some steps of eigenization are enmeshed in the type (size) processing, while some are cleanly separated.  Determination of which operations can be eigenized is enmeshed, while transformation of the syntax tree for eigen syntax is separate.  We could benefit from separating some of the enmeshed steps.

I'd like to merge this PR because I plan to modify `sizeReturn` by adding error-trapping on incorrect types.

PR #356, opened 2017-05-17
fix use of nimC, nimRep, nimSeq and ':' in BUGS code

perrydv:
Uses of “core R” functions were not working in BUGS code.  This PR fixes that.  Regression tests are at bottom of test-coreR


packages/nimble/inst/tests/test-coreR.R
```diff
@@ -726,3 +726,81 @@ seqTestsResults <- lapply(seqTests, test_coreRfeature)
...
+    a ~ dnorm(0,1)
+    b[1:2] <- c(a, 1.1)
+})
+
+m <- nimbleModel(mc, inits = list(a = 2))
+cm <- try(compileNimble(m))
+test_that('c(a, 1.1) in BUGS compiles',
+          expect_false(inherits(cm, 'try-error')))
+test_that('c(a, 1.1) in BUGS',
+          expect_identical(as.numeric(m$b), c(2, 1.1)))
```
fritzo: Is it worth strengthening your first `test_that(...)` to also check compiled behavior?
```
test_that('c(a, 1.1) in compiled BUGS', {
    cm <- compileNimble(m)
    expect_identical(as.numeric(cm$b), c(2, 1.1))
})

File: packages/nimble/inst/tests/test-coreR.R [hunk 2]
```diff
@@ -726,3 +726,81 @@ seqTestsResults <- lapply(seqTests, test_coreRfeature)
...
+
+##
+mc <- nimbleCode({
+    b[1:2] <- c(1.2, 1.1)
+})
+
+m <- nimbleModel(mc)
+cm <- try(compileNimble(m))
```
fritzo: I believe you can simplify these three lines to simply
```
test_that('c(1.2, 1.1) in BUGS compiles', compileNimble(m))
```
because `test_that` wraps its arguments in a try-catch. BTW this is why I have been using the style
```
test_that('...', {
    # Do as much as possible here
    # so that if if fails, the tests below will still run.
})
```

PR #357, opened 2017-05-17
MCEM Covariance Estimation

NLMichaud:
This commit adds an `estimateCov()` method to the methods returned by a call to `buildMCEM()`.  A test of `estimateCov()` was added to test-mcem.R. 

Note that this commit will break any previous uses of `buildMCEM()` - previously, a call to `buildMCEM()` would return a function that could be called to obtain the MLEs.  Now, a call to `buildMCEM()` returns a list with two elements: 

1.   `run()`, a function that can be called to obtain the MLEs, and
2.   `estimateCov()`, a function that can be called to estimate the asymptotic covariance

Thus any historical uses of the function returned by a call to `buildMCEM()` will have to add a `$run` to their call.


perrydv:
@NLMichaud, I'm wondering if we can arrange the pieces so that someone could do the estimateCov step even if they got MLE from some other method?

NLMichaud:
Sorry if I did a poor job of making this clear from the documentation - that should be possible for someone to do by providing a named vector of MLEs as the `MLEs` argument to `estimateCov`

perrydv:
But they would still need to use `buildMCEM`, even if they have no interest in the MCEM part?  If so, that seems awkward.

NLMichaud:
I agree that it is a bit awkward.  We could define the `estimateCov` function separately from `buildMCEM`, but then a user would not be able to use the MLEs and latent state samples that were produced within `buildMCEM()$run` in the `estimateCov` function without manually entering them as arguments.   However, it would be a somewhat cleaner user experience - I'd be happy to separate these two functions if you'd like.

perrydv:
Based on our chat, LGTM.  In future we will refactor this for some components to be available as separate nimbleFunctions.

PR #359, opened 2017-05-18
Refactor and add comments to sizeNewNimbleList

fritzo:
Earlier today I was struggling to read this function, so to "actively read" it I have:
- Refactored to eliminate duplicated code.
- Refactored by returning early. This practice reduces nesting depth and increases readability.
- Removed all code that is commented out. (That's what version control is for.)
- Made comments more grammatical. Sentences start with capital letters and end with a period.
- Added a TODO asking somebody to explain a mysterious comment. (This is the second time I've asked about this comment. Please somebody help me understand.)

### DIFF
--- a/packages/nimble/R/genCpp_sizeProcessing.R
+++ b/packages/nimble/R/genCpp_sizeProcessing.R
@@ -577,17 +577,13 @@ sizeRep <- function(code, symTab, typeEnv) {
 }
 
 sizeNewNimbleList <- function(code, symTab, typeEnv){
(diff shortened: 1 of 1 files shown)


fritzo:
Thanks for the explanation, @NLMichaud ! I've updated the comment and added a TODO to remove the extra test for "" as you suggest (I'm trying to avoid any behavior changes in this PR).

perrydv:
LGTM.  Thanks @fritzo and these are good practices to emulate more actively.

PR #360, opened 2017-05-18
dcar_normal distribution

danielturek:
This PR contains NIMBLE support for the improper CAR distribution `dcar_normal`, the name chosen to be consistent with the geoBUGS distribution `car.normal` (which is also provided as an alias).

Includes R and cpp support for density evaluation of `dcar_normal` nodes.  This uses a correction discussed to make it a proper density function, but careful review is welcomed from @paciorek.  The calculation used is described in a comment preceding the cpp implementation in `CppCode/dists.cpp`, which is used for both compiled and uncompiled execution.  It's been thoroughly checked that the cpp code correctly evaluates the expression given in the comment.

Also provides MCMC support for sampling from the `dcar_normal` distribution.  `dcar_normal` nodes are assigned a `CAR_normal` sampler, which manages and assigns one of three univariate samplers to each dimension of the target node:

- `CAR_scalar_postPred` for dimensions with no stochastic dependencies
- `CAR_scalar_conjugate` for normal-normal conjugate dimensions
- `CAR_scalar_RW` for all other dimensions.

These component samplers make correct use of the new `getDependencies` functionality for getting dependencies of single dimensions of multivariate nodes.

Further, the `CAR_scalar_conjugate` includes logic to check whether the "two-point method" calculations are necessary or not (on a case-by-case basis).  The "two-point method" calculations of linear coefficients are skipped when the model structure allows.

At present, there are a few diagnostic messages built into the sampler construction process, telling what sampler is assigned to each dimension.  These messages can (and should) be removed before merging, but for now, are helpful.

This PR also provides two utility functions, for transforming CAR structural parameters into the `adj`, `weights`, `num` parameters required for the  `dcar_noral` distribution.  `CAR_convertNeighborWeightLists()` converts lists of neighbor indices and weights, and `CAR_convertWeightMatrix()` converts a weight matrix.


paciorek:
Some (mostly minor) thoughts (and recognizing that some of these are nit-picks on cleanup things you just haven't done yet):

* perhaps we want the functions that convert to/from lists/matrices (`CAR_convertWeightMatrix`, etc.) to have names like "matrix2list" or "matrix_to_list" (hopefully we already have a quasi-standard way of naming these kinds of conversion functions).

* I don't think we need `nimble:::` when calling `isMixedSizes`.  See also use of `nimble:::` in new code in MCMC_samplers.R.

* We'll want a full roxygen for dcar_normal eventually. Let's put a note in there soon that we either need to handle the case or warn the users when the graph separates into multiple cliques (islands) or the graph is set up as a higher-order model. In either case  `c` would not be 1. 

* See use of `storage.mode` in dinvwishart as a way to only cast if needed while preserving any dimensionality of non-scalars. I'd apply that approach to x, adj, weights, and num.

* In C++ for dcar_normal where you have log(tau/2/pi), please see use of R's C constants such as M_LN_SQRT_PI and M_LN_SQRT_2PI (let me know if you need more of a pointer). Also shouldn't it be tau/(2*pi)?

* I'd use `count++` rather than `count = count + 1 ` in the C++ for dcar_normal.

* Side note: we hard code things like 0.44 as the `optimalAR`. We should probably pull such constants out as part of some sort of configuration object. 

* 200 is hard-coded in as the adaptInterval in CAR_scalar_RW.

danielturek:
@paciorek Thanks for the helpful review!

Some comments back for you:

- I changed your previous comment (above) into checkboxes, as I address them.
- `a/b/c` is the same as `a/(b*c)`.
- I'm quite open to changes on the naming of `CAR_convertWeightMatrix`, but I do think it's helpful to keep the `CAR_` or `car_` prefix for everything related to CAR models.  That said, I'm at a bit of a loss for what these should be named (I went around in circles here, a bit).  `car_list2XXX` and `car_matrix2XXX` still don't seem to make intuitive sense to me, regardless of your choice of `XXX`.  I'm going to hold off making changes to these names, until if/when we have a concrete suggestion that we (or just you) like better -- feel free to make such a suggestion if you have an idea, and I'm happy to go with it.

danielturek:
Most uses of `nimble:::` were removed, but that proceeding `nimble:::isMixedSizes` seems to be necessary. (See many other similar uses in BUGS_model.R).

I think the optimal acceptance rate of 0.44 is pretty fixed for MH-RW (single dimension) sampling, so that's unchanged.  But I added flexibility for adaptiveness, adaptation interval, and initial scale of the CAR component RW sampler.

I added preliminary documentation for the CAR_normal sampler (in MCMC_samplers.R).  I encourage @paciorek to review this, and elaborate as you desire.

I also added documentation for the `dcar_normal` distribution in distributions_implementations.R.  It's reasonable, but I left the `@details` section blank, and would encourage @paciorek to add something here.  Is that something you could do?

I'll hold off on modifying the names of `CAR_convertNeighborWeightLists` and `CAR_convertWeightMatrix` until we have concrete better suggestions.

paciorek:
Ok, all sounds good.

perrydv:
Generally looks good to me.  I'm not going into all details to check correctness.  I assume the new samplers successfully avoid recomputing the entire `dcar_normal` node, and that's why they need their own samplers.  Thanks @danielturek!

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -805,7 +808,11 @@ modelDefClass$methods(liftExpressionArgs = function() {
                 requireNewAndUniqueDecl <- any(contexts[[BUGSdecl$contextID]]$indexVarNames %in% all.vars(paramExpr))
                 uniquePiece <- if(requireNewAndUniqueDecl) paste0("_L", BUGSdecl$sourceLineNumber) else ""
                 newNodeNameExpr <- as.name(paste0('lifted_', Rname2CppName(paramExpr, colonsOK = TRUE), uniquePiece))   ## create the name of the new node ##nameMashup
-                newNodeNameExprIndexed <- addNecessaryIndexingToNewNode(newNodeNameExpr, paramExpr, contexts[[BUGSdecl$contextID]]$indexVarExprs)  ## add indexing if necessary
+                if(deparse(paramExpr[[1]]) %in% liftedCallsDoNotAddIndexing) {   ## skip adding indexing to mixed-size calls
```
perrydv: `deparse` is expensive.  I see that inside `isExprLiftable` there are multiple implied deparses (or `as.character`s, but in general `deparse` is safer).  Can we either assume `as.character` is valid (since it works in `isExprLiftable`) or *once* do `deparse(paramExpr[[1]]` and use the result in these various places?

perrydv:
Regarding the parts you wanted me to look at, I made one minor comment and otherwise LGTM.

packages/nimble/R/BUGS_modelDef.R
danielturek: Yes, can do.

PR #366, opened 2017-05-20
new testing tools: withTempProject and expect_compiles, along with force01 option

perrydv:
EXPERIMENTAL: This contains new testing tools demonstrated in test-testTools.R (intended as transient file).  

* withTempProject takes charge of the nimbleProject and forces clearCompiled(), i.e. unloads the DLL.  

* expect_compiles stops the compilation after generating -o files but before linking.  It *also* replaces any -O# flags with -O1.  That behavior can be controlled with nimbleOptions()$forceO1


perrydv:
@fritzo here are experimental versions of the tools we talked about.

packages/nimble/R/options.R
```diff
@@ -7,6 +7,8 @@ nimbleUserNamespace <- as.environment(list(sessionSpecificDll = NULL))
 # These options are for development use at this point.
 .nimbleOptions <- as.environment(
     list(
+        nimbleProject = NULL,  ## only used by withTempProject and compileNimble in testing code.
```
fritzo: Consider renaming this to `nimbleProjectForTesting` to avoid confusing users.

packages/nimble/inst/tests/test_utils.R
```diff
@@ -26,6 +26,49 @@ temporarilyAssignInGlobalEnv <- function(value) {
...
+    code <- substitute(code)
+    project <- nimble:::nimbleProjectClass()
+    nimbleOptions(nimbleProject = project)
+    on.exit({
+        ## messages are suppressed by test_that, so "assign('.check', 1, globalenv())" can be used as a way to verify this code was called 
+        .project <- nimbleOptions()$nimbleProject
+        nimbleOptions(nimbleProject = NULL) ## clear this before clearCompiled step, in case clearCompiled() itself fails
+        .project$clearCompiled()
+    })
+    try(eval(code, envir = parent.frame()))
```
fritzo: I'm pretty sure you don't want a try here, right? Otherwise tests cannot fail.

packages/nimble/inst/tests/test_utils.R
```diff
@@ -26,6 +26,49 @@ temporarilyAssignInGlobalEnv <- function(value) {
...
+}
+
+expect_compiles <- function(...) {
+    nimbleOptions(stopCompilationBeforeLinking = TRUE)
```
fritzo: nit: Consider preserving previous values, as in
```
oldSCBL <- nimbleOptions$stopCompilationBeforeLinking
nimbleOptions(stopCompilationBeforeLinking = TRUE)
...
on.exit({
  nimbleOptions(stopCompilationBeforeLinking = oldSCBL)
  ...
})
```

packages/nimble/inst/tests/test_utils.R
```diff
@@ -26,6 +26,49 @@ temporarilyAssignInGlobalEnv <- function(value) {
...
+    nimbleOptions(forceO1 = TRUE)
+    on.exit({
+        assign('.check', 1, globalenv())
```
fritzo: What does this do?

packages/nimble/inst/tests/test_utils.R
```diff
@@ -26,6 +26,49 @@ temporarilyAssignInGlobalEnv <- function(value) {
...
+        nimbleOptions(stopCompilationBeforeLinking = FALSE)
+        nimbleOptions(forceO1 = FALSE)
+    })
+    ans <- try(compileNimble(...)) ## expecting a thrown error
```
fritzo: I guess I don't understand the R exception mechanism, but couldn't you simply
```
    })
    compileNimble(...)
}
```
so that if `compileNimble` fails (via stop()), that will propagate up?

fritzo: Oh, now I get it! This is the exception relay mechanism we talked about.

packages/nimble/inst/tests/test_utils.R
fritzo: I think this would simpler using a `tryCatch(-, finally)` construct:
```
withTempProject <- function(code) {
    code <- substitute(code)
    project <- nimble:::nimbleProjectClass()
    nimbleOptions(nimbleProject = project)
    tryCatch(eval(code, envir = parent.frame()), finally = {
        ## messages are suppressed by test_that, so "assign('.check', 1, globalenv())" can be used as a way to verify this code was called 
        .project <- nimbleOptions()$nimbleProject
        nimbleOptions(nimbleProject = NULL) ## clear this before clearCompiled step, in case clearCompiled() itself fails
        .project$clearCompiled()
    })
}
```
This has different behavior than your first implementation, since it runs cleanup code without catching the exception. To observe the difference, note that in composite tests, the return value will be hidden:
```
test_that('function compiles and DSL version works', {
    nimFun <- nimbleFunction(run = function() { return(0); returnType(double(0)) })
    expect_compiles(nimFun)  # This can never fail in your version 1, but can fail in the tryCatch version.
    expect_equal(nimFun(), 0)
})
```
Thanks for taking the time to work this out! I'm definitely learning more about R's exception mechanism while trying to get this right.

packages/nimble/inst/tests/test_utils.R
```diff
@@ -26,6 +26,49 @@ temporarilyAssignInGlobalEnv <- function(value) {
...
+    if(identical(ans, "safely stopping before linking")) 
+        NULL
+    else
+        ans
```
fritzo: I think you probably want to `stop(ans)` rather than `return(ans)` here, so that tests can still fail. Actually, the testthat package has a bunch of nice tests using `expect_failure()`, and maybe we add similar tests for our test helpers, e.g.
```
test_that('expect_compiles can fail', {
    expect_failure(expect_compiles('a string cannot be compiled'))
})
```

packages/nimble/R/options.R
perrydv: 👍

packages/nimble/inst/tests/test_utils.R
perrydv: Using `tryCatch` did not seem meaningfully different to me.  However I note `?on.exit` indicates that only one `on.exit` instruction will be stored for a function.  We may be at risk of `temporarilyAssignInGlobalNamespace` and `withTempProject` stepping on each other's use of `on.exit.

packages/nimble/inst/tests/test_utils.R
perrydv: 👍

packages/nimble/inst/tests/test_utils.R
perrydv: Yes, but in the update I'm about to sync for testing, I now have an explicit `expect_identical`.  It seems safest to return from our own `expect_....` with a call to a native `expect_...` function from `testthat`.

packages/nimble/inst/tests/test_utils.R
perrydv: On the second point: I now have verification of correct test failing in `test-testTools.R`

fritzo: I think we're ok if we take care to `on.exit(..., add = TRUE)`.

packages/nimble/inst/tests/test_utils.R
perrydv: Replaced the whole clause with an `expect_identical`

perrydv:
Thanks for the code review @fritzo !  Very helpful.  This got old so I've updated from devel, revised, and will sync soon to let tests run on the PR.  This stuff can get tricky quickly.

packages/nimble/inst/tests/test_utils.R
perrydv: Oh great!

PR #367, opened 2017-05-22
Bugfixes to optim support

fritzo:
This PR fixes lots of bugs in the initial optim implementation and adds support for many missing parameters.
- Support for gradients
- Support for `lower` and `upper` bounds
- Support for `fnscale` parameter


fritzo:
@perrydv Could you please do a quick review of this code?
@NLMichaud This PR should cover all uses of `optim()` needed to compile your MCEM code.
@paciorek Please let me know when 0.6-5 is released and devel is unfrozen.

packages/nimble/inst/CppCode/predefinedNimbleLists.cpp
```diff
@@ -246,12 +247,14 @@ void OptimControlNimbleList::copyFromSEXP(SEXP S_nimList_) {
     SET_STRING_ELT(S_pxData, 0, mkChar(".xData"));
     PROTECT(S_trace = findVarInFrame(GET_SLOT(S_nimList_, S_pxData),
```
perrydv: Per last round of rchk, I think we when there are multiple arguments to an R API function we need to ensure everything is `PROTECT`ed, with the exception of `install` calls.  So `PROTECT(GET_SLOT(S_nimList_, S_pxData))`.  This applies to all similar lines, and the corresponding `UNPROTECT` argument needs to match the number of `PROTECT`s.

perrydv:
I made one minor comment about adding some `PROTECT`s.  Otherwise :thumbsup: .  Thanks @fritzo!

packages/nimble/inst/CppCode/predefinedNimbleLists.cpp
fritzo: This is entirely generated code, so the actual error is in our compiler. Could you point me towards where in our compiler these lines are being generated?

fritzo: Ah I see in `cppDefs_nimbleList.R`. This issue is independent of this PR, so I'll fix that in a separate PR.

perrydv: Thanks.  I forgot I manually fixed them for 0.6-5 but they are generated.

fritzo: Lots of questions:
- Did you manually fix the file marked 'DO NOT EDIT BY HAND' or another file?
- Could you point me to the commit where you did this modification, so I can merge upstream into the compiler?
- Do you know how to run `rchk`? It seems very difficult.

fritzo: Ah I'll try to follow the pattern in bb93076b

fritzo: I've merged the optim work and am addressing the independent `PROTECT` issue in #406

PR #368, opened 2017-05-23
ReturnErrorTrapping

perrydv:
This traps errors where the type or dimensionality of the argument to return() does not match the declared returnType.

It also fixes a bug where the messages from carefully trapped errors in size processing were not being displayed to the user.  That explains some of the frustrations we’ve had with error trapping recently.


paciorek:
Merging per Perry's Slack request.

PR #369, opened 2017-05-23
WAIC calculation

NLMichaud:
This commit adds two versions of a `calculateWAIC()` function, which calculate a model's WAIC given using posterior samples from an MCMC algorithm.  

- The first version is a new method of a NIMBLE MCMC object, named `calculateWAIC`. If someone calls `mcmcObject$calculateWAIC()` after having called `mcmcObject$run(nIter)`, the `calculateWAIC` method will use the posterior samples saved in `mvSamples` to calculate the WAIC. 

- The second version is a standalone `nimbleFunction` named `calculateWAIC()`, that takes as arguments a `nimbleModel` object, and a `modelValues` object that contains  posterior samples (e.g. an `mvSamples` object obtained after running an MCMC algorithm).  

 Roxygen documentation and testing is included.   Reference for WAIC calculation is [here](http://www.stat.columbia.edu/~gelman/research/published/waic_understand3.pdf), see Equations 5, 12, and 13.


packages/nimble/inst/tests/test-waic.R
```diff
@@ -0,0 +1,38 @@
...
+schoolSATmcmcConf <- configureMCMC(schoolSATmodel, monitors = c('schoolmean'))
+schoolSATmcmc <- buildMCMC(schoolSATmcmcConf)
+CschoolSATmcmc <- compileNimble(schoolSATmcmc, project = schoolSATmodel)
+CschoolSATmcmc$run(50000)
+schoolWAICfunc <- calcWAIC(schoolSATmodel, schoolSATmcmc)
+CschoolWAICfunc <- compileNimble(schoolWAICfunc, project = schoolSATmodel)
+schoolWAIC <- CschoolWAICfunc$run(20000)
+
+##below WAIC value from Gelman '13 "Understanding predictive information criteria for Bayesian models"
+try(test_that("Test that WAIC is accurate: ",
```
fritzo: What's the `try()` for? Is this test a known failure?

NLMichaud:
Once tests are passed, this should be ready to merge.  I've included two tests in the `test-waic.R` file, each of which calculate the WAIC and compare it to known values for a different model.  

@paciorek @danielturek I've put information about `calculateWAIC()` into the manual in chapter 7 (the MCMC chapter).  Most of the information I put into a new section directly after section 7.4 (Extracting MCMC samples).  Let me know if either of you, or anyone else, have a preference for it to be in some other part of the manual and I'd be happy to move it or reorganize!

danielturek:
@NLMichaud Just from an organizational standpoint, that seems reasonable to me.

PR #370, opened 2017-05-23
add .onLoad()

perrydv:
Propose adding .onLoad() message inviting users to visit R-nimble.org


danielturek:
I think it's a great idea.

paciorek:
Agreed. Do we want to print out the version number too?

perrydv:
@paciorek  yes please add code to the version number, while keeping the message concise.

Issue #379, opened 2017-05-24
Compiled code doesn't like sum of 0-length vectors


danielturek:
Example of trying to sum a 0-length vector.  Gives a frightening warning message.

In R, `sum(numeric())` returns 0, which makes sense to me.

Reproducible example:

```
library(nimble)

code <- nimbleCode({
    a ~ dnorm(0, 1)
})
constants <- list()
data <- list()
inits <- list(a = 0)
Rmodel <- nimbleModel(code, constants, data, inits)

nfDef <- nimbleFunction(
    setup = function(model, nodes) {},
    run = function() {
        a <- values(model, nodes)
        b <- sum(a)
    }
)

Rnf <- nfDef(Rmodel, character(0))

Cmodel <- compileNimble(Rmodel)
Cnf <- compileNimble(Rnf, project = Rmodel)

Rnf$run()
Cnf$run()
## Warning: after completing ManyModelAccess_2_nimArr, nimCurrent != nimEnd.
## Perhaps the NimArr was longer than the accessor?
```

perrydv:
See PR #380

perrydv:
This was fixed.

PR #380, opened 2017-05-24
add default constructors in some accessorClasses.cpp classes for empty cases.

perrydv:
Fixes Issue #379

### DIFF
--- a/packages/nimble/inst/include/nimble/accessorClasses.h
+++ b/packages/nimble/inst/include/nimble/accessorClasses.h
@@ -161,7 +161,8 @@ class SingleVariableMapAccessBase {
  public:
   int offset, length;
   bool singleton;
-  vector<int> sizes, strides; 
+  vector<int> sizes, strides;
+ SingleVariableMapAccessBase() : offset(0), singleton(false) {length = 0;}; /* length(0) triggers the Rf_length macro mixup */
   virtual ~SingleVariableMapAccessBase();
   virtual NimArrType *getNimArrPtr()=0;
   void calculateLength() {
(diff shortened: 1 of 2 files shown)


packages/nimble/inst/include/nimble/accessorClasses.h
```diff
@@ -161,7 +161,8 @@ class SingleVariableMapAccessBase {
  public:
   int offset, length;
   bool singleton;
-  vector<int> sizes, strides; 
+  vector<int> sizes, strides;
+ SingleVariableMapAccessBase() : offset(0), singleton(false) {length = 0;}; /* length(0) triggers the Rf_length macro mixup */
```
fritzo: Remove trailing semicolon here and everywhere else.

Also consider simplifying to `... : offset(0), length(0), singleton(false) {}`

packages/nimble/inst/include/nimble/accessorClasses.h
```diff
@@ -285,6 +287,7 @@ class SingleVariableAccessBase {
  public:
   int flatIndexStart, flatIndexEnd; // For 1-3: In R these should be 1,3; In C++: 0, 3
   int length; // copying function can check for singletons if it wants to
+ SingleVariableAccessBase() : flatIndexStart(0), flatIndexEnd(0) { length = 0; };
```
fritzo: simplify to `... : flatIndexStart(0), flatIndexEnd(0), length(0) {}`

danielturek:
If it fixes the problem, then looks good to me.  Thanks @perrydv

packages/nimble/inst/include/nimble/accessorClasses.h
perrydv: length(0) triggers error from R's macro `length` which replaces `length(x)` with `Rf_length(x)`.  You're getting us fixed on that but meanwhile `length(0)` breaks.

perrydv:
I cleaned up the trailing semi-colons.  That is such a trivial change I am now going to squash and merge. @danielturek this should let you move ahead with Issue #379 fixed

danielturek:
Thanks!

Issue #386, opened 2017-05-28
Issue with binary sampler


vandenman:
Hi, I encountered an issue with the binary sampler. That is, it gives the wrong result. The slice sampler does give the correct result. Thoughts?

```
library(nimble)

# the model: is the mean b[1] or b[2]?
modelBern <- nimbleCode(code = {
	
	indicator ~ dbern(0.5)
	mu <- indicator * b[1] + (1 - indicator) * b[2]
	for (i in 1:n) {
		
		x[i] ~ dnorm(mu, 1)
		
	}
	
})

# simulate data
set.seed(31415)
b = c(1e3, 5)
n = 1e3
x = rnorm(n, b[2], 1) # mean is b[2], indicator should be 0.
mean(x) # 5.01399

rModel <- nimbleModel(code = modelBern,
				  data = list(x = x),
				  constants = list(b = b, n = n),
				  dimensions = list(b = 2),
				  calculate = TRUE)
MCMCconf <- configureMCMC(model = rModel)
MCMCconf$printSamplers()

bMCMC <- buildMCMC(conf = MCMCconf)
cModel <- compileNimble(rModel)
mcmcObjBern <- compileNimble(bMCMC, project = cModel, resetFunctions = TRUE)

samplesBern <- runMCMC(mcmc = mcmcObjBern, niter = 1e3, nchains = 2, nburnin = 1e2, returnCodaMCMC = TRUE)
plot(samplesBern) # should be 0, but it is 1
colMeans(as.matrix(samplesBern)) # should be 0, but it is 1

# now with slice sampler

MCMCconf <- configureMCMC(model = rModel)

MCMCconf$removeSamplers("indicator")
MCMCconf$addSampler(target = 'indicator', type = 'slice')
MCMCconf$printSamplers()

bMCMC <- buildMCMC(conf = MCMCconf)
mcmcObjBern <- compileNimble(bMCMC, project = cModel, resetFunctions = TRUE)

samplesBern <- runMCMC(mcmc = mcmcObjBern, niter = 1e3, nchains = 2, nburnin = 1e2, returnCodaMCMC = TRUE)
plot(samplesBern) # should be 0 and is 0
colMeans(as.matrix(samplesBern)) # should be 0 and is 0
```

danielturek:
@vandenman thanks for providing a reproducible example.  We'll look into this soon, and get back to you.

perrydv:
[...]

PR #387, opened 2017-05-29
fix underflow in sampler_binary

perrydv:
This pull request proposes a fix for Issue #386

### DIFF
--- a/packages/nimble/R/MCMC_samplers.R
+++ b/packages/nimble/R/MCMC_samplers.R
@@ -56,10 +56,11 @@ sampler_binary <- nimbleFunction(
         if(!model$isBinary(target))     stop('can only use binary sampler on discrete 0/1 (binary) nodes')
     },
     run = function() {
-        currentProb <- exp(getLogProb(model, calcNodes))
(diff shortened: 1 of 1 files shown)


danielturek:
LGTM @perrydv. Thanks for jumping on this.  I appreciate the variable name changes, too.

I'd say good to merge, but might be too late for this mini release?

paciorek:
there's still time. i'll merge in.

perrydv:
Yesterday while quickly jumping into it I didn't go over whether the behavior of that `is.nan` is what we want.  I think we want to make sure cases of `acceptanceProb` 0 and Inf work correctly.

perrydv:
I just looked at acceptanceProb values of 0 and 1 and they seem to work correctly.  But should the `is.nan` check throw an error instead of default rejection of the proposal?

danielturek:
I think the `is.nan` is correct and what we want.  I wrestled with this extensively when writing the slice sampler.  Please let me know however if I'm wrong and you find a good failure case.

paciorek:
I'm happy to include in release if you both are good with that.

danielturek:
Ok with me.

PR #389, opened 2017-05-30
Re-enable `sort = TRUE` functionality for `model$expandNodeNames()`.

NLMichaud:
When the `expandNodeNames()` function was revised in #347, the `sort` argument lost its functionality.  This PR simply passes the `sort` argument from the `expandNodeNames()` function into the `expandNodeNamesFromGraphIDs()` function, which should allow nodes to be topologically sorted again.

### DIFF
--- a/packages/nimble/R/BUGS_model.R
+++ b/packages/nimble/R/BUGS_model.R
@@ -453,7 +453,7 @@ unique: should names be the unique names or should original ordering of nodes (a
 
                                       if(length(nodes) == 0) return(if(returnType=='names') character() else numeric())
                                       graphID <- modelDef$nodeName2GraphIDs(nodes, !returnScalarComponents, unique = unique)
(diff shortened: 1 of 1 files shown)


perrydv:
Thanks @NLMichaud .  This looks like the right fix.  What happened is in a recent change I split `expandNodeNames` into two methods, `expandNodeNames` and `expandNodeNamesFromGraphIDs`.  The idea was that the former would call the latter to achieve identical behavior as in the past, but it would also be useful from other methods to call the latter directly, skipping the sometimes-redundant steps at the beginning of `expandNodeNames`.  I see that I forgot to propagate the `sort` argument from the first to the second function.  OTOH it appears the `unique` argument is never needed in the second function (and CRAN-checking often flags such problems), so I am going to remove that and push into the branch of this PR.

PR #396, opened 2017-05-31
fix use of predefined nimbleLists

perrydv:
This PR:

1. Adds predefined `nimbleList` `SEXP` functions to dynamicRegistrations.h, so they will be compiled into the `sessionSpecificDll`

2. Updates `nimbleList` interface system to handle the case that the dll where relevant functions are (in a predefined list) may not be the same as the one other parts of a compiled project are in.

Managing the registrations and dlls correctly will be important since CRAN is moving towards enforcement of registration, and we are generally in good shape as we do use registration extensively.

### DIFF
--- a/packages/nimble/R/cppInterfaces_nimbleFunctions.R
+++ b/packages/nimble/R/cppInterfaces_nimbleFunctions.R
@@ -324,14 +324,32 @@ makeNimbleListBindingFields <- function(symTab, cppNames, castFunName) {
     return(fieldList)
 }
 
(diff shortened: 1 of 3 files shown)


paciorek:
So this PR should fix the nimbleList test failures we're seeing on OS X?

paciorek:
whoops, just saw Nick's slack message. I can test on OS X if no one else
has.

On Wed, May 31, 2017 at 12:12 PM, Chris Paciorek <paciorek@stat.berkeley.edu

PR #399, opened 2017-05-31
Modernize test coreRfeature

perrydv:
This modernizes `test_coreRfeature` as follows:

* Everything is wrapped in `test_that`
* All `try(test_that(…))` idioms removed.
* Remaining `try` for `compileNimble` uses `silent=TRUE` and does lead to a managed use of `expect_equal` to check if compiler failure or success was as expected.
* Uses in `test-coreRfeature`, `test-allocation` and test-numericTypes` work on my machine.

### DIFF
--- a/packages/nimble/inst/tests/test_utils.R
+++ b/packages/nimble/inst/tests/test_utils.R
@@ -51,10 +51,18 @@ indexNames <- function(x) {
     lapply(x, function(z) {z$name <- paste(i, z$name); i <<- i + 1; z})
 }
 
-test_coreRfeature <- function(input, verbose = TRUE, dirName = NULL) { ## a lot like test_math but a bit more flexible
+
+test_coreRfeature <- function(input, verbose = TRUE, dirName = NULL) {
+    test_that(input$name, {
+        test_coreRfeature_internal(input, verbose, dirName)
+    })
+}
+
+test_coreRfeature_internal <- function(input, verbose = TRUE, dirName = NULL) { ## a lot like test_math but a bit more flexible
   if(verbose) cat("### Testing", input$name, "###\n")
   runFun <- gen_runFunCore(input)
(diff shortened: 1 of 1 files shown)


packages/nimble/inst/tests/test_utils.R
```diff
@@ -51,22 +51,28 @@ indexNames <- function(x) {
...
+    })
+}
+
+test_coreRfeature_internal <- function(input, verbose = TRUE, dirName = NULL) { ## a lot like test_math but a bit more flexible
   if(verbose) cat("### Testing", input$name, "###\n")
   runFun <- gen_runFunCore(input)
   nfR <- nimbleFunction(run = runFun)
-  nfC <- try(compileNimble(nfR, dirName = dirName))
+  ## This try is safe because failure is caught by expect_equal below
+  nfC <- try(compileNimble(nfR, dirName = dirName), silent = TRUE)
```
fritzo: This looks reasonable. I'm not sure it's appropriate here, but I've also found `expec_error` helpful in `test_size()`:
```r
if(failureExpected) {
  expect_error(compileNimble(...))
} else {
  nf <- compileNimble(...)
}
```

fritzo:
LGTM

BTW Have you checked with both `test_package`/`test_file` and `Rscript`?

perrydv:
That `expect_error` example would be cleaner but equivalent in behavior.  I'll try it.  Thanks for the reminder to try Rscript mode as well.   I ran via `test_that` locally and I guess they are running via Rscript on travis.

fritzo:
@perrydv It looks like this merge conflicted with my travis-osx merge, possibly due to different behavior of `expect_identical` on linux and OS X. The current `devel` is now failing OS X but passing Linux. As you suggested, I think this should be an easy fix.

Issue #395, opened 2017-05-31
Test failures are silenced when Rscript and try(test_that()) are combined


fritzo:
## Overview

Many of our tests on travis-ci are (and always have been) silenced. To fix this, we can set `runViaTestthat` in [`run_tests.R`](https://github.com/nimble-dev/nimble/blob/devel/run_tests.R). However because some of our tests are currently failing, this will require either fixing those tests or `skip()`ping those tests as `KNOWN_FAILURES`.

## Detailed diagnosis

When `test_that()` is wrapped in a `try()`, then test failures will be reported by `test_package('nimble', 'myFile')` but are silenced by `Rscript inst/tests/test-myFile.R`. Specifically, `Rscript ...` will print the errors but then continue to subsequent tests and exit with successful return code 0. This is problematic because our current travis tests are (and always have been) run via `Rscript` rather than `testthat::test_package()`.

It is probable that developers adopted the `try(test_that(-))` idiom to avoid stopping at the first failure when running under `Rscript` (or when sourcing an entire file in an IDE). While this idiom is helpful for manual inspection, it has led to silenced errors when tests are run automatically.

As of 25f24b1, this repo contains 177 occurrences of the `try(test_that())` idiom with discarded return value, spread over 11 files.
```sh
$ find . -name '*.R' \
  | xargs grep -P '^\s*try\([\s\n]*test_that' -hc \
  | awk '{s+=$1} END {print s}'
177

$ find . -name '*.R' \
  | xargs grep -P '^\s*try\([\s\n]*test_that' -c \
  | grep -v ':0$' \
  | sed 's/\(.*\):\(.*\)/\2\t\1/g' \
  | sort -nr
41	./inst/tests/test-distributions.R
38	./inst/tests/test-trunc.R
36	./inst/tests/test-getDistributionInfo.R
16	./inst/tests/test-dsl_dists.R
13	./inst/tests/test_utils.R
11	./inst/tests/test-user.R
8	./inst/tests/test-models.R
6	./inst/tests/test-nimbleFunctionInterfaces.R
5	./inst/tests/test-mcem.R
2	./inst/tests/test-mcmc.R
1	./inst/tests/test-modelValuesInterface.R
```

## Possible fixes

1. Run all travis tests through `testthat::test_package`. This can be done simply by setting setting `runViaTestthat = TRUE` in [`run_tests.R`](https://github.com/nimble-dev/nimble/blob/devel/run_tests.R).

[...]

Issue #363, opened 2017-05-19
test-size.R cannot run under test_package('nimble', 'size')


fritzo:
To reproduce
```{r}
library(methods)
library(testthat)
library(nimble)
test_file('inst/tests/test-size.R')  # PASSES
test_package('nimble', 'size')  # FAILS
```

paciorek:
It runs for me on Ubuntu 16.04. Can you provide more details of the failure?

fritzo:
Here is a [failing test on travis](https://travis-ci.org/nimble-dev/nimble/builds/234183789?utm_source=github_status&utm_medium=notification), and below is a failing test on my Ubuntu 16.04 box on devel branch
```
library(nimble)
library(testthat)
test_package('nimble', 'size')
..### Testing scalar stochastic, parameter expression  with RHS variable ###
defining model...
building model...
setting data and initial values...
running calculate on model (any error reports that follow may simply reflect missing values in model variables) ... 
checking model sizes and dimensions... note that missing values (NAs) or non-finite values were found in model variables: y. This is not an error, but some or all variables may need to be initialized for certain algorithms to operate properly.
model building finished.
..### Testing scalar stochastic, parameter expression with RHS constant ###
defining model...
building model...
running calculate on model (any error reports that follow may simply reflect missing values in model variables) ... 
checking model sizes and dimensions... note that missing values (NAs) or non-finite values were found in model variables: y. This is not an error, but some or all variables may need to be initialized for certain algorithms to operate properly.
model building finished.
..### Testing scalar stochastic, parameter expression non-scalar, no indices  with RHS variable ###
defining model...
building model...
setting data and initial values...
running calculate on model (any error reports that follow may simply reflect missing values in model variables) ... 
checking model sizes and dimensions... note that missing values (NAs) or non-finite values were found in model variables: y. This is not an error, but some or all variables may need to be initialized for certain algorithms to operate properly.
model building finished.
12### Testing scalar stochastic, parameter expression non-scalar, no indices with RHS constant ###
defining model...
building model...
running calculate on model (any error reports that follow may simply reflect missing values in model variables) ... 
checking model sizes and dimensions... note that missing values (NAs) or non-finite values were found in model variables: y. This is not an error, but some or all variables may need to be initialized for certain algorithms to operate properly.
model building finished.
..### Testing scalar stochastic, parameter expression non-scalar, RHS index  with RHS variable ###
defining model...
building model...
setting data and initial values...
[...]

PR #400, opened 2017-05-31
Fix silent test failures in test-size.R

fritzo:
This addresses #363 and #395 by avoding the `try(test_that())` idiom in the helper functions `test_size()` and `test_size_specific_error()`. This fix surfaced a few new failures and a few previous failures that have silently been fixed.

@paciorek After this PR it should no longer be necessary to compare the text outputs against a golden file.
 any change will be considered an error (both pass-to-fail and fail-to-pass). All changes will need to be accounted for in the `knownProblem` arguments. But at least we're more automated :smile: .

Note that the automated checking against golden `knownProblem` values required splitting that flag into a finer-grained pair of flags `knownProblem` and `knownProblemWithConst`.

### DIFF
--- a/packages/nimble/inst/tests/test-size.R
+++ b/packages/nimble/inst/tests/test-size.R
@@ -474,11 +474,12 @@ testsLHSRHSmismatch <- list(
 )
 
 # for these we simply want errors caught at model-building not compilation
+if(0){
 sapply(testsScalar, test_size)
 sapply(testsMultivarParam, test_size)
 sapply(testsDeterm, test_size)
 sapply(testsMultivar, test_size)
 sapply(testsTrunc, test_size)
-
+}
 # for these we want errors caught by NIMBLE error-checking not by R errors at model-building
 sapply(testsLHSRHSmismatch, test_size_specific_error)
--- a/packages/nimble/inst/tests/test-size.R
+++ b/packages/nimble/inst/tests/test-size.R
@@ -28,21 +28,23 @@ testsScalar <- list(
     list(name = 'scalar stochastic, parameter expression', expectPass = TRUE,
          expr = quote({y ~ dnorm(mu1 + mu2, sd = sig)}), 
          inits = list(mu1 = 0, mu2 = 0, sig = 1) ),
-    list(name = 'scalar stochastic, parameter expression non-scalar, no indices', expectPass = FALSE, expectPassWithConst = TRUE,
-         knownProblem = TRUE,
+    list(name = 'scalar stochastic, parameter expression non-scalar, no indices',
+         expectPass = FALSE, expectPassWithConst = TRUE,
+         knownProblem = TRUE, knownProblemWithConst = FALSE,
          expr = quote({y ~ dnorm(mu1%*%mu2, sd = sig)}), 
          inits = list(mu1 = vec2, mu2 = vec2, sig = 1) ),
     
     list(name = 'scalar stochastic, parameter expression non-scalar, RHS index',
-         expectPass = FALSE,
-         knownProblem = TRUE,
+         expectPass = FALSE, expectPassWithConst = FALSE,
+         knownProblem = TRUE, knownProblemWithConst = FALSE,
          expr = quote({y ~ dnorm((mu1%*%mu2)[1,1], sd = sig)}), 
          inits = list(mu1 = vec2, mu2 = vec2, sig = 1) ),
     # passes for RHS init, though compileNimble does give helpful error message
     # passes for RHS const but fails in compilation because of lack of indexing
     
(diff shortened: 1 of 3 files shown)


packages/nimble/inst/tests/test-size.R
```diff
@@ -28,21 +28,23 @@ testsScalar <- list(
...
+    list(name = 'scalar stochastic, parameter expression non-scalar, no indices',
+         expectPass = FALSE, expectPassWithConst = TRUE,
+         knownProblem = TRUE, knownProblemWithConst = FALSE,
          expr = quote({y ~ dnorm(mu1%*%mu2, sd = sig)}), 
          inits = list(mu1 = vec2, mu2 = vec2, sig = 1) ),
     
     list(name = 'scalar stochastic, parameter expression non-scalar, RHS index',
-         expectPass = FALSE,
-         knownProblem = TRUE,
+         expectPass = FALSE, expectPassWithConst = FALSE,
```
perrydv: Is `expectPassWithConst` supposed to be `TRUE` here?  How could it be "known" `FALSE` yet `knownProblemWithConst` also `FALSE`?

packages/nimble/inst/tests/test_utils.R
```diff
@@ -788,49 +788,45 @@ weightedMetricFunc <- function(index, samples, weights, metric, samplesToWeights
...
-        result <- try(
-            { calculate(m); out <- calculate(m)} )
-        try(test_that(paste0("Test 2 of size/dimension check: ", input$name),
-                      expect_equal(!is(result, "try-error"), input$expectPass,
-                                  info = errorMsg)))
+        calculate(m)  ## Calculates from scratch.
+        calculate(m)  ## Uses cached value.
+    })
+    if(xor(input$expectPass, input$knownProblem)) {
+        test_that(paste(input$name, 'with RHS variable works as expected'), eval(code))
```
perrydv: Should this say "works or fails as expected"?  Or have the message depend on the value of `input$expectePass`?  Seems like if `expectPass` is `FALSE` and `knownProblem` is `TRUE` then it should say that it failed as expected.

packages/nimble/inst/tests/test_utils.R
```diff
@@ -788,49 +788,45 @@ weightedMetricFunc <- function(index, samples, weights, metric, samplesToWeights
...
-                  expect_equal(!is(result, "try-error"), input$expectPass,
-                              info = errorMsg)))
-    if(!is(result, "try-error")) {
-        result <- try(
-            { calculate(m); out <- calculate(m)} )
-        try(test_that(paste0("Test 2 of size/dimension check: ", input$name),
-                      expect_equal(!is(result, "try-error"), input$expectPass,
-                                  info = errorMsg)))
+        calculate(m)  ## Calculates from scratch.
+        calculate(m)  ## Uses cached value.
```
perrydv: do you mean `getLogProb`?

packages/nimble/inst/tests/test-size.R
fritzo: What this means is that
- for variables: nimble should FAIL but does not, and this is a known problem.
- for constants: nimble should FAIL and does, so there is no problem
Does that make sense?

run_tests.R
```diff
@@ -20,11 +20,11 @@ allTests <- row.names(testTimes)
 
 # Run each test in a separate process to avoid dll garbage overload.
 for (test in allTests) {
-    runViaTestthat <- FALSE  # TODO Fix test-size.R and set this to TRUE.
+    runViaTestthat <- TRUE  # TODO Fix test-size.R and set this to TRUE.
```
perrydv: Did you intend to set `runViaTestthat <- TRUE` in the PR or save flipping that switch for later?

packages/nimble/inst/tests/test_utils.R
fritzo: Good point, I'll make the error messages clearer.

run_tests.R
fritzo: Oh thanks, right. I did mean to leave that out of the review.

packages/nimble/inst/tests/test_utils.R
fritzo: I was blindly preserving the existing code. @paciorek ?

paciorek:
I can't keep up with this right now. Feel free to do as you see fit and
I'll connect back in next week. An in person discussion might be helpful to
make sure we are all on the same page of how we want our testing process to
work.

fritzo:
@perrydv Do you mind if I merge this?

packages/nimble/inst/tests/test_utils.R
fritzo: I think we can address this behavior later; this PR does preserve existing behavior.

PR #401, opened 2017-05-31
put all of `test_math` inside `test_that` and remove `try()`s

perrydv:
This PR updates `test_math`, used by `test-math.R`, so all steps are inside `test_that` and `try(expect_<condition>(…))` idioms are removed.

### DIFF
--- a/packages/nimble/inst/tests/test-math.R
+++ b/packages/nimble/inst/tests/test-math.R
@@ -198,16 +198,16 @@ testsMatrix = list(
 
 
 set.seed(0)
-sapply(testsVaried, test_math)    ## 12
-sapply(testsBasicMath, test_math) ## 70
+ans1 <- sapply(testsVaried, test_math)    ## 12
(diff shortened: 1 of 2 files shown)


packages/nimble/inst/tests/test-math.R
```diff
@@ -198,16 +198,16 @@ testsMatrix = list(
 
 
 set.seed(0)
-sapply(testsVaried, test_math)    ## 12
-sapply(testsBasicMath, test_math) ## 70
+ans1 <- sapply(testsVaried, test_math)    ## 12
```
fritzo: What are the `ans1` etc. variables for?

fritzo:
LGTM (after possibly removing `ans*` variables)

packages/nimble/inst/tests/test-math.R
perrydv: Those ans variables suppress some output if run by directly sourcing the file (or possibly via Rscript; I'd have to check).

PR #406, opened 2017-06-01
PROTECT calls to GET_SLOT in nimbleList generated code

fritzo:
This follows up on `rchk` results from the 0.6-4 CRAN release of NIMBLE.

The only hand-changed code is in `cppDefs_nimbleList.R`. The changes to C++ code were automatically regenerated.


perrydv:
:thumbsup: thanks for jumping on this!

PR #407, opened 2017-06-01
Enable codecov reporting on travis

fritzo:
**Status:** Do not merge.

This enables R code coverage reporting on the travis linux build.


perrydv:
`codecov` will be useful but let's hold off on using its results until we've cleaned up known remnants.

fritzo:
OK, I'm hoping that `codecov` will help me identify which code is dead (since 'known' only works for you and Chris :smile:)

perrydv:
I get it, but soon we'll have a lot of that out of the way.

PR #409, opened 2017-06-02
Modernize test mcmc

perrydv:
This has passed tests and is ready to merge pending code review.

This pull request attempts to improve MCMC testing.  No details of tests have been modified, only how they are used and managed.

* Wrap the guts of `test_mcmc` inside a `test_that`.
* Put setup steps prior to `test_mcmc` calls or individual tests inside `test_that`.
* Establish system of `knownFailures` list to simply propagate expectations for multiple `expect_*` calls.
* More thorough use of `verbose` to allow management of output.
* Establish a “gold file” system to keep large text output out of CI log files and automatically check for differences.

Current issues:

* Some tempdir() names end up in gold file, triggering failure because tempdir() will be different during a test run.  Will clean up before being ready to merge.
* In future the gold file could be compressed in a variety of ways.  Right now it is similar (not identical) to the old log output.

Except for the tempdir() issue, code review is welcome.


packages/nimble/inst/tests/test-mcmc.R
```diff
@@ -2,25 +2,26 @@ source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
...
-  out <- as.mcmc(out1)
-}
-
-
-### Beginning of actual tests
+RwarnLevel <- options('warn')$warn
+options(warn = -1)
+nimbleVerboseSetting <- nimbleOptions('verbose')
+
+## If you do *not* want to write to results files
```
fritzo: 👍  This seems to me to be a cleaner way to verify results over time, as compared to our current practice of manually snapshotting test results files for each Cran release.

fritzo:
LGTM on the general approach, but I defer to @paciorek since he owns this test file and manages MCMC tests.

@perrydv Thanks for putting in the time to fully automate these tests. I think it will pay off as we refactor the compiler towards tensorization.

packages/nimble/inst/tests/test-mcmc.R
```diff
@@ -2,25 +2,26 @@ source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
...
+## To record a new gold file, nimbleOptions('generateGoldFileForMCMCtesting') should contain the path to the directory where you want to put it
+## e.g. nimbleOptions(generateGoldFileForMCMCtesting = getwd())
+## Comparison to the gold file won't work until it is installed with the package.
+nimbleOptions(verbose = TRUE)
+goldFileName <- 'mcmcTestLog_Correct.Rout'
+tempFileName <- 'mcmcTestLog.Rout'
+generatingGoldFile <- !is.null(nimbleOptions('generateGoldFileForMCMCtesting'))
+outputFile <- if(generatingGoldFile) file.path(nimbleOptions('generateGoldFileForMCMCtesting'), goldFileName) else tempFileName
+
+sink(outputFile)
```
fritzo: How did you end up dealing with the travis-fail-on-silence issue? Ideally we could use an R function like the unix `tee` command, so that results were written to both console and a file.

perrydv: The challenge is that I'm using R's `sink()` to redirect all output to our log file for comparison to gold log.  Although documentation implies that `message`s will not be redirected, they do seem to be, or at least to not be output to the Travis log.  So as a workaround I emitted messages to Travis log via `system("echo <message>")`.

PR #413, opened 2017-06-02
Simplify nodeFun.h

fritzo:
**Status:** Ready to merge as soon as tests pass.

This PR removes an old incomplete attempt at model vectorization so we can start from scratch.

This PR also renames towards a nomenclature of "a program of instructions":

- class `NodeInstruction` (previously named `oneNodeUseInfo`)
- `.operand` field of `NodeInstruction` (previously named `.useInfo`)
- `.instructions` field of `NodeVectorClassNew` (previously named `.useInfoVec`)


fritzo:
@perrydv My current thinking after reading through `nodeFun.h` and `accessorClasses.*` is that we should first simplify those files and then implement vectorization from scratch, as informed by our tensorization work. This PR starts that simplification process.

More generally, I think we will do well to keep the codebase small and simple, so that we can refactor more often and be less bogged down by maintenance.

perrydv:
Generally sounds good.  You're a step ahead of me in that I'm pruning out substantial vestigial code in `accessorClasses.*` right now.  But even after I'm done we could simplify and re-write some parts.  Let's discuss.

perrydv:
LGTM on the revised nomenclature and simplification of `calculateBlock` reflecting that it wasn't really in full intended use.  Looks like you traced it back to the passing of information from R to C++. 

Note for future discussion: we need to maintain support for `model$calculate(nodes[i])`.  This may defeat blocking and require a different nodeFunctionVector or other solution.

perrydv:
It appears the failing tests are in test-mcmc, due to running tests through `test_package` prior to having `test-mcmc` updated.  That PR now passes, so we can update this branch to pass.

PR #416, opened 2017-06-03
Compile backdoor

perrydv:
This creates a specialHandling system for RCfunctions to allow easy swapping of hand-written C++ in place of generated C++, for development purposes.

Example is on [wiki page](https://github.com/nimble-dev/nimble/wiki/Developer-backdoor-to-manually-replace-generated-C)

Branch modernize-test-mcmc was merged with this to make testing pass


fritzo:
Oooh interesting, the Linux build passed, but the OS X build failed in `test-coreR.R`. Any idea what's wrong?
```
Error: Test failed: 'index assignment from boolean vector 2 (2D, mixed types)'
* `out` not identical to `out_nfC`.
Objects equal but not identical
Identical test of math (direct R vs. C++ nimbleFunction): index assignment from boolean vector 2 (2D, mixed types)
Execution halted
Error: FAILED test-coreR.R
Execution halted
```
As the message suggests, maybe we can replace an `expect_identical()` with an `expect_equal()`?

packages/nimble/R/nimbleFunction_core.R
```diff
@@ -117,6 +117,30 @@ nimbleFunction <- function(setup         = NULL,
     return(generatorFunction)
 }
```
fritzo: This looks tricky! Could you describe in a comment how it's used? My guess is
```
## To substitute a user-provided .cpp file in place of a nimble generated .cpp file, first define
## a nimble function via nf <- nimbleFunction(...), and then set nf.specialHandling <- 'my_substitute.cpp'.
```
Maybe a simpler approach would be to just set a global `nimbleOptions(fakeCppFile = 'my_substitute.cpp')` and use this as the file in the one place we call `R CMD SHLIB`?

perrydv: Oops that comment may have been residue of initial thoughts.  The Wiki page linked from the PR online has a full demo.  It is [here](https://github.com/nimble-dev/nimble/wiki/Developer-backdoor-to-manually-replace-generated-C).

The test failure seems to be occurring on multiple PRs.  I will look into it later, but I am fairly sure it is some last bump in all the test updating, not related to this particular PR or feature.

perrydv:
@fritzo This has passed tests.  I'll hold off merging it until you look at the Wiki page and comment.

packages/nimble/R/nimbleFunction_core.R
fritzo: Thanks the wiki makes this look easy to use. Could you add a comment on the function that links to the wiki, so as to improve discoverability of documentation? More generally, what do you think of more generally adding documentation to internal functions?

fritzo:
:+1: after adding a link to wiki page from the function (either as a comment or function documentation).

Do I understand correctly that you've automatically disabled uniqe-prefix-generation for all functions for which there is `specialHandling`?

perrydv:
Good idea to link to the wiki.  Should we prefer to have that large chunk of documentation in the code itself, or to link to it at the wiki?

The disabling of unique-prefix-generation occurs when you provide the `projectName` argument to `compileNimble`.  A prefix like `P_<number>` is the generated unique `projectName` provided by the system of no `projectName` argument is provided.  I'd forgotten exactly how that worked when we talked last week.

perrydv:
I don't think our build process regenerates Rd files (unless we've updated it).  So I added a comment above the `specialHandler` code with the link to the wiki page.

paciorek:
Rds are regenerated manually in prep_pkg.R, and Fritz is going to make a
Makefile target for Rds that calls that.

PR #417, opened 2017-06-04
set checkEqual=TRUE for "index assignment from boolean vector 2 (2D, mixed types)"

perrydv:
This single test in test-coreR works with check_identical on my machine but appears to need to be check_equal on Travis’ OS X.  It was newly revealed by adding OS X testing on Travis.


fritzo:
👍  Thanks for fixing this!

PR #420, opened 2017-06-05
Cleanup nodeFunction R

perrydv:
Some time ago we renovated the `nodeFunction` construction system.  This PR removes now-unused code lying around from the old system.


fritzo:
:+1: Thanks for cleaning this up!

PR #422, opened 2017-06-05
use switch() in copyFromRobject

perrydv:
This replaces a long series of `if` … `else if` with a `switch` statement.  There is a modest gain of 1-2% for compilation of 1000 RW samplers.

### DIFF
--- a/packages/nimble/R/cppInterfaces_nimbleFunctions.R
+++ b/packages/nimble/R/cppInterfaces_nimbleFunctions.R
@@ -842,15 +842,15 @@ copyFromRobject <- function(Robj, cppNames, cppCopyTypes, basePtr, symTab, dll) 
             warning("Problem in copyFromRobject.  There is an object to be copied that is NULL.  Going to browser.", call. = FALSE)
             browser()
         }
-        if(cppCopyTypes[[v]] == 'modelVar') {
+        switch(cppCopyTypes[[v]],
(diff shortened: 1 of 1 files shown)


packages/nimble/R/cppInterfaces_nimbleFunctions.R
```diff
@@ -899,88 +892,58 @@ copyFromRobject <- function(Robj, cppNames, cppCopyTypes, basePtr, symTab, dll)
...
+        },
+        'logicalScalar' = {
             getSetLogicalScalar(v, Robj[[v]], basePtr, dll = dll)
-            next
-        }
-        else if(!(cppCopyTypes[[v]] %in% c('copierVector'))) {
-            warning(paste0("Note: cppCopyTypes not recognized. Type = ", cppCopyTypes[[v]], "\n"), call. = FALSE)
+        },
+        { ## default:
+            if(!(cppCopyTypes[[v]] %in% c('copierVector'))) {
```
fritzo: Consider breaking `copierVector` out into another case, and maybe moving this case to the top?
```
},
'copierVector': {},
{  # default
  warning(paste0("Note: cppCopyTypes not recognized. Type = ", cppCopyTypes[[v]], "\n"), call. = FALSE)
}
)
```

fritzo:
:+1:

PR #423, opened 2017-06-05
Cleanup NimArrBase.h and NimArr.h

fritzo:
This removes commented-out code, applies `clang-format`, and does minor style fixes to make `NimArr*.h` more readable. See the detailed commit history for individual changes.

PR #424, opened 2017-06-05
Add AD functionality for `nimbleFunction` methods

NLMichaud:
This PR implements the framework of AD for `nimbleFunction` methods.  Currently, all AD features are marked as EXPERIMENTAL.  In order to use AD as provided by this PR, see the guide at [the NIMBLE AD wiki page](https://github.com/nimble-dev/nimble/wiki/Automatic-differentiation-in-NIMBLE).  Note that the `cppad/` folder must be placed within `packages/nimble/inst/include/`, and that `nimbleOptions(enableDerivs = TRUE)` must be called before using any AD features.

As of this PR, the limitations of AD in NIMBLE are as follows:

1. `derivs(calculate(model, nodes))` does not currently work. 
2.  Users cannot specify which arguments they would like derivatives with respect to (by default, derivatives are taken with respect to all arguments).
2. Currently, the only distribution with AD support is the univariate normal (`dnorm`).
3.  It is possible (likely) that other, non-distributional functions in NIMBLE's dsl are likewise not supported by AD yet.  More testing is needed to determine which functions are not yet supported.


perrydv:
Not sure but you might need to disable the test file or the travis script will run it automatically and it will fail.

packages/nimble/R/BUGS_model.R
```diff
@@ -1127,7 +1135,12 @@ RmodelBaseClass <- setRefClass("RmodelBaseClass",
...
-                                           nfGenerator <- nimble:::nodeFunctionNew(LHS=LHS, RHS=RHS, name = thisNodeGeneratorName, altParams=altParams, bounds=bounds, logProbNodeExpr=logProbNodeExpr, type=type, setupOutputExprs=setupOutputExprs, evaluate=TRUE, where = where)
+
+                                           
+                                           nfGenerator <- nimble:::nodeFunctionNew(LHS=LHS, RHS=RHS, name = thisNodeGeneratorName, altParams=altParams, bounds=bounds, 
+                                                                                   parentsSizeAndDims = parentsSizeAndDims, logProbNodeExpr=logProbNodeExpr, type=type,
+                                                                                   setupOutputExprs=setupOutputExprs, evaluate=TRUE, where = where)
+                                           # nfGenerator <- nimble:::nodeFunctionNew(LHS=LHS, RHS=RHS, name = thisNodeGeneratorName, altParams=altParams, bounds=bounds, logProbNodeExpr=logProbNodeExpr, type=type, setupOutputExprs=setupOutputExprs, evaluate=TRUE, where = where)
```
fritzo: nit: If we avoid adding commented-out code to the codebase, it makes it easier to do systematic searches using tools like `grep` and `ag` during refactoring. If you do include commented-out code in the repo, I would recommend adding an English description of the commented-out code so as to provide context for other readers, e.g.
```
# Note that AD support requires parentSizesAndDims, and without AD this function would could be simplified to:
```

packages/nimble/R/cppDefs_core.R
```diff
@@ -7,13 +7,11 @@
 ## Base class for C++ file and compilation information
 cppDefinition <- setRefClass('cppDefinition', 
                              fields = list(
-                                 filename = 'ANY',	#  character, what filename (to which .h and .cpp will be appended) is this definition in
-                                 CPPusings = 'ANY',	#  character
```
fritzo: Here's a random idea that's outside the scope of this PR: what if we set these types to something like
```
CPPusings = if(checkNimbleTypes) 'character' else 'ANY'
```
and then set a global variable `checkTypes = FALSE` or ` = (length(Sys.getenv('NIMBLE_CHECK_TYPES')) > 0)`, so that we can check types while developing code that might break type checking, but avoid the overhead in general use. @perrydv What do you think?

packages/nimble/R/cppDefs_core.R
```diff
@@ -77,11 +94,14 @@ cppClassDef <- setRefClass('cppClassDef',
                                extPtrTypes = 'ANY',
                                private = 'list',		# 'list'. This field is a placeholder for future functionality.  Currently everything is generated as public
                                useGenerator = 'ANY',		#'logical',
-                               SEXPgeneratorFun = 'ANY',        ## cppFunctionDef object
-                               SEXPfinalizerFun = 'ANY'),       ## cppFunctionDef object
+                               SEXPgeneratorFun = 'ANY', ## These will be cppFunctionDefs.
+                               SEXPfinalizerFun = 'ANY',
+                               globalObjectsDefs = 'ANY'
```
fritzo: @NLMichaud Could you explain the `globalObhectsDefs` to me in person? It might be useful for the tensorflow work Perry and I are doing.

packages/nimble/R/genCpp_generateCpp.R
```diff
@@ -187,7 +202,7 @@ cppOutputFor <- function(code, symTab) {
     begin <- nimGenerateCpp(code$args[[2]]$args[[1]], symTab)
     end <- nimGenerateCpp(code$args[[2]]$args[[2]], symTab)
     iterVar <- nimGenerateCpp(code$args[[1]], symTab)
-    part1 <- paste0('for(', iterVar ,'=', begin,'; ', iterVar, '<= static_cast<int>(', end, '); ++', iterVar,')')
+    part1 <- paste0('for(', iterVar ,'=', begin,'; ', iterVar, '<= static_cast<int>(', end, '); ', iterVar,'++ )')
```
fritzo: In the early 2000's pre-increment was cheaper than post-increment, though I'm not sure that's still the case. Is there a good reason to switch from pre- to post- here?

packages/nimble/R/makevars.R
```diff
@@ -64,7 +64,14 @@ function(target, vars = character(), .useLib = UseLibraryMakevars)
                                    else
                                      "Makevars", package = "nimble")
 
+    cppad_inc <- nimbleOptions("CppAD_directory")
+    # if(is.na(cppad_inc)) cppad_inc <- ""
+    # else cppad_inc <- paste0("-I\"",cppad_inc,"\"")
+    # browser()
+    # cppad_inc <- paste0("-I ", cppad_inc)
+    cppad_inc <- '-I"/Users/iateb/Documents/cppad/cppad-20170413"'
```
fritzo: Clean this up before committing.

packages/nimble/R/nimbleFunction_nodeFunctionNew.R
```diff
@@ -1,25 +1,44 @@
...
+nodeFunctionNew <- function(LHS, RHS, name = NA, altParams, bounds, parentsSizeAndDims, 
+                            logProbNodeExpr, type, setupOutputExprs, evaluate = TRUE, where = globalenv()) {    
     if(!(type %in% c('stoch', 'determ')))       stop(paste0('invalid argument to nodeFunction(): type = ', type))
     setupOutputLabels <- nndf_makeNodeFunctionIndexLabels(setupOutputExprs) ## should perhaps move to the declInfo for preservation
     LHSrep <- nndf_replaceSetupOutputsWithIndexedNodeInfo(LHS, setupOutputLabels)
     RHSrep <- nndf_replaceSetupOutputsWithIndexedNodeInfo(RHS, setupOutputLabels)
     altParamsRep <- lapply(altParams, nndf_replaceSetupOutputsWithIndexedNodeInfo, setupOutputLabels)
     boundsRep <- lapply(bounds, nndf_replaceSetupOutputsWithIndexedNodeInfo, setupOutputLabels)
     logProbNodeExprRep <- nndf_replaceSetupOutputsWithIndexedNodeInfo(logProbNodeExpr, setupOutputLabels)
+    if(nimbleOptions('enableDerivs')){
```
fritzo: If you plan on switching this flag to default to `TRUE` in the future, you might name it `experimentalEnableDerivs` to indicate that it is only being used during feature development.

packages/nimble/autom4te.cache/output.0
```diff
@@ -631,6 +631,10 @@ NIMBLE_DIR
 R_PACKAGE_DIR
 NEED_MAKEVARS_FILE
 EIGEN_INC
+<<<<<<< HEAD
```
fritzo: Clean this merge conflict up before merging.

packages/nimble/autom4te.cache/requests
```diff
@@ -14,6 +14,7 @@
                         'configure.ac'
                       ],
                       {
+<<<<<<< HEAD
```
fritzo: Clean up this merge conflict before merging.

packages/nimble/autom4te.cache/traces.0
```diff
@@ -173,6 +173,7 @@ m4trace:configure.ac:11: -1- m4_pattern_allow([^EXEEXT$])
 m4trace:configure.ac:11: -1- AC_SUBST([OBJEXT], [$ac_cv_objext])
 m4trace:configure.ac:11: -1- AC_SUBST_TRACE([OBJEXT])
 m4trace:configure.ac:11: -1- m4_pattern_allow([^OBJEXT$])
+<<<<<<< HEAD
```
fritzo: Clean up this merge conflict before merging.

packages/nimble/inst/include/nimble/NimArr.h
```diff
@@ -39,7 +39,7 @@ class NimArr<1, T> : public NimArrBase<T> {
   }
 
   template<class Tother>
-  NimArr<1, T> &templateCopyOperator(const NimArr<1, Tother> &other) { // This makes the copied-to object contiguous-memory.  Use mapCopy to copy into an existing map
+  NimArr<1, T> &operator=(const NimArr<1, Tother> &other) { // This makes the copied-to object contiguous-memory.  Use mapCopy to copy into an existing map
```
fritzo: Nice!

fritzo:
:+1: I haven't reviewed the implementation thoroughly, but I like the idea of keeping closer to mainline and marking features EXPERIMENTAL and adding experimental tests to the blacklist.

packages/nimble/inst/include/nimble/NimArr.h
```diff
@@ -86,10 +86,55 @@ class NimArr<1, T> : public NimArrBase<T> {
     return(*this);
   }
 
-  NimArr<1, T> &operator=(const NimArr<1, double> &other) {return(templateCopyOperator<double>(other));}
-  NimArr<1, T> &operator=(const NimArr<1, int> &other) {return(templateCopyOperator<int>(other));}
-  NimArr<1, T> &operator=(const NimArr<1, bool> &other) {return(templateCopyOperator<bool>(other));}  
+  /* NimArr<1, T> &operator=(const NimArr<1, double> &other) {return(templateCopyOperator<double>(other));} */
```
fritzo: Let me know when you merge this PR. It will conflict with my style-cleanup PR #423, so I'll hold off merging until you've merged this (to save you the conflict resolution effort).

packages/nimble/R/BUGS_model.R
NLMichaud: Good catch!  Commented-out code has been removed.

packages/nimble/autom4te.cache/traces.0
NLMichaud: Thanks for noticing these conflicts, they have been cleaned.

packages/nimble/R/cppDefs_core.R
perrydv: That makes a lot of sense!  We determined in the past that type-checking during routine use imposes very noticeable cost, but your suggestion would allow us to avoid a lot of splintered error-trapping during development.  Let's try it.

packages/nimble/R/genCpp_generateCpp.R
perrydv: Yes that change makes sense.  I think I've read some claims that a compiler can make them equivalent if possible, but I could be mis-remembering or it may be incorrect.

packages/nimble/R/makevars.R
perrydv: Since we only expect this to matter if `nimbleOptions('enableDerivs')==TRUE`, we could make this `cppad_inc` <- paste0("'-I\\"", nimbleOptions('CppADdir'), "\\"'")`

perrydv:
Since we have another PR waiting on this one for clean merging, and this one is experimental, I'm giving this :thumbsup: once the comments have been addressed.  Thanks @NLMichaud !

packages/nimble/R/nimbleFunction_nodeFunctionNew.R
NLMichaud: That's a good idea! I've switched the `nimbleOption` from `enableDerivs` to `experimentalEnableDerivs`.

packages/nimble/R/genCpp_generateCpp.R
NLMichaud: Thanks for the feedback, the switch was reverted.

NLMichaud:
Thanks for the feedback @fritzo and @perrydv !  It seems as though testing tried to run `test-ADmodels.R` (and failed), even though `test-ADmodels.R` was added to the blacklist in `run_tests.R` [here](https://github.com/nimble-dev/nimble/blob/fixed_nickAD/run_tests.R).  Is it possible that the command used to run `test-models.R` somehow picked up `test-ADmodels.R` as well?  If so, I can change the name of AD model testing to something like `test-ADmods.R`.

fritzo:
@NLMichaud looks like you've found a bug in `run_tests.R`. Here's a fix #428. I'm going to patch that into your branch just so test get a chance to run tonight.

NLMichaud:
Thanks for the patch @fritzo!

PR #425, opened 2017-06-06
initial setup for refactoring size processing to separate eigen-specific steps.

perrydv:
This PR is ready to merge if it passes tests.

This PR include basic setup for refactoring size processing to disentangle eigen-specific steps from general size and type annotation steps.  Eigen-specific steps include annotation of what can be handled by eigen and generating intermediates where needed for implementation using eigen.  Separating these steps into distinct compiler passes should make the code cleaner and more amenable to extension such as for other back-ends.

Currently the experimental size processing works for everything in `basicMathTests`.

Changes include:

* Two new compiler flags that toggle new behavior (default off): `experimentalNewSizeProcessing` and `experimentalSelfLiftStage`.

* A new compiler stage, `exprClasses_toEigenize`, modeled on `exprClasses_setSize`, with handlers partially set up for simple cases, with placeholders for non-simple cases.

* A new test file.  This includes a function that will compare whether old and new generated C++ files are identical for a given RCfunction.  It currently has one toy test for that function.  It also includes a function that wraps a call to `test_math` and checks whether current and experimental processing of each case in `basicMathTests` is identical.  That passes.


packages/nimble/R/RCfunction_compile.R
```diff
@@ -269,6 +269,11 @@ RCfunProcessing <- setRefClass('RCfunProcessing',
                                        if(inherits(tryResult, 'try-error')) {
                                            stop(paste('There is some problem at the insertAdditions processing step for this code:\n', paste(deparse(compileInfo$origRcode), collapse = '\n'), collapse = '\n'), call. = FALSE)
                                        }
+
+                                       if(nimbleOptions('useRefactoredSizeProcessing')) {
+                                           exprClasses_setToEigenize(compileInfo$nimExpr, compileInfo$newLocalSymTab, compileInfo$typeEnv)
```
fritzo: Is this the one place where we will eventually toggle based on `nimbleOptions('useTensorflow')`?

perrydv: At the moment, yes.  I think it will get slightly more complicated: If lifting temporaries out of expressions happens at later steps, we'll need to manage which pathway of catch-up processing needs to happen on those temporaries.

packages/nimble/R/RCfunction_compile.R
```diff
@@ -265,10 +265,15 @@ RCfunProcessing <- setRefClass('RCfunProcessing',
                                            browser()
                                        }
 
+                                       if(nimbleOptions('useRefactoredSizeProcessing')) {
```
fritzo: It's a good habit to name these short-lived flags `experimentalMyFlagName` or `internalMyFlagName` to make it clear to our users that they shouldn't use these flags.

packages/nimble/R/all_utils.R
```diff
@@ -2,20 +2,35 @@
...
 double <- function(ndim, dims) {}
 
-labelFunctionCreator <- function(lead, start = 1) {
-  nextIndex <- start
-  force(lead)
-  labelGenerator <- function(reset = FALSE, count = 1, envName = "") {
-    if(reset) {
-      nextIndex <<- 1
-      return(invisible(NULL))
+labelFunctionMetaCreator <- function() {
```
fritzo: Add comment describing the purpose of this function.

packages/nimble/R/genCpp_sizeProcessing.R
```diff
@@ -252,6 +252,8 @@ sizeDim <- function(code, symTab, typeEnv) {
 }
 
 sizeDiagonal <- function(code, symTab, typeEnv) {
+    ## refactor: code$name change step stays here
```
fritzo: nit: If you label your comments `useRefactoredSizeProcessing`, you will have an easier time systematically cleaning up those comments when you're done.

packages/nimble/inst/tests/mathTestLists.R
```diff
@@ -0,0 +1,197 @@
+1### INSTRUCTIONS:
```
fritzo: Thankyou for adding instructions! :+1:

packages/nimble/inst/tests/test-refactorCompilationSteps.R
```diff
@@ -0,0 +1,61 @@
+source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
+
+context("Testing of old vs. new generated C++ during refactoring steps")
+
+oldWarnLevel <- options('warn')
+options(warn = -1)
+
+## Known concern: ordering of asRow()/asCol() and intermediates
+
+compareOldAndNewCompilationRC <- function(input) {
```
fritzo: It would be nice to have unit tests, rather than these tests that "this unit behaves the same as it used to".

run_tests.R
```diff
@@ -2,7 +2,7 @@
 # This script runs most tests in nimble/inst/tests/ prioritized by duration.
 
 # Avoid running these blacklisted tests, since they take too long.
-blacklist <- c('test-Math2.R', 'test-Mcmc2.R', 'test-Mcmc3.R', 'test-Filtering2.R', 'test-ADfunctions.R', 'test-ADmodels.R')
+blacklist <- c('test-Math2.R', 'test-Mcmc2.R', 'test-Mcmc3.R', 'test-Filtering2.R', 'test-ADfunctions.R', 'test-ADmodels.R','test-refactorSizeProcessing.R')
```
fritzo: Why is the test blacklisted? Does it currently fail? This knowledge deserves to live as a comment in the code:
```r
blacklist <- c(blacklist, 'test-refactorSizeProcessing.R')  # These tests of an experimental refactoring currently fail.
```

perrydv:
Thanks for the code review @fritzo !  I've updated for all points except the larger step of adding different flavors of tests.  If tests pass I think this is ready to merge.

fritzo:
:+1:

PR #431, opened 2017-06-13
specialHandling compiler "backdoor" works for models

perrydv:
This PR extends the “backdoor” or “specialHandling” system to models.  This system allows experimentation with rewriting C++ code and seeing if it will compile within the nimble workflow.  The compiled objects would be dangerous to use (could break if meaningful changes have been made), but it is useful to see if C++ compilation is successful in order to experiment with redesigning C++ model code.

[Documentation](https://github.com/nimble-dev/nimble/wiki/Developer-backdoor-to-manually-replace-generated-C#backdoor-for-models)

### DIFF
--- a/packages/nimble/R/nimbleProject.R
+++ b/packages/nimble/R/nimbleProject.R
@@ -520,8 +520,10 @@ nimbleProjectClass <- setRefClass('nimbleProjectClass',
                                      modelDef <- model$getModelDef()
                                      modelDefName <- modelDef$name
                                      Cname <- Rname2CppName(modelDefName)
-                                     if(is.null(filename)) {
-                                         filename <- paste0(projectName, '_', Rname2CppName(modelDefName)) 
+                                     if(!disableWrite) {
+                                         if(is.null(filename)) {
+                                             filename <- paste0(projectName, '_', Rname2CppName(modelDefName)) 
+                                         }
                                          nfFileName <- paste0(projectName, '_', Rname2CppName(modelDefName),'_nfCode')
                                      }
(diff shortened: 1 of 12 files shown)


packages/nimble/R/nimbleProject.R
```diff
@@ -508,8 +520,10 @@ nimbleProjectClass <- setRefClass('nimbleProjectClass',
                                      modelDef <- model$getModelDef()
                                      modelDefName <- modelDef$name
                                      Cname <- Rname2CppName(modelDefName)
-                                     if(is.null(filename)) filename <- paste0(projectName, '_', Rname2CppName(modelDefName)) 
-                                     
+                                     if(is.null(filename)) {
+                                         filename <- paste0(projectName, '_', Rname2CppName(modelDefName)) 
+                                         nfFileName <- paste0(projectName, '_', Rname2CppName(modelDefName),'_nfCode')
```
fritzo: Should `nfFileName` happen outside of this `if` statement? (I'm just confirming that you intended the behavior change.)

perrydv: Ah, you're right, it's an unintended change in behavior.

perrydv: Just fixed it.

fritzo:
:+1:

perrydv:
We are now timing out on Travis-CI due to tests running > 3 hours.  Let's discuss.

fritzo:
I think it's safe to merge this if time-limit is the reason for build failure.

Issue #371, opened 2017-05-23
Possible way to reduce size of built package


perrydv:
I think if at the [-include line of the CppCode GNUmakefile](https://github.com/nimble-dev/nimble/blob/3f2950f45a24ad0c8c167dc6306287fc7dfbda1b/packages/nimble/inst/CppCode/GNUmakefile#L11) we strip out all the "-g" flags from the Makeconf file being included, it will reduce the size of libnimble.a substantially.

paciorek:
I'm inclined to leave as is unless CRAN complains about the size. Looks like we'd have to grab Makeconf, process with some tool like sed via some invocation we'd have to figure out (and we'd want to make sure sed was available on users' systems presumably) and then include the revised file.

fritzo:
@perrydv This issue is resolved, right?

perrydv:
yes

paciorek:
Just a note for future ref if we ever come back to the question of package size and stripping symbols:

https://www.r-bloggers.com/21-a-third-and-final-post-on-stripping-r-libraries/

PR #432, opened 2017-06-14
experimental reduction in libnimble.a size

perrydv:
This PR reduces the size of libnimble.a from about 12.4 Mb to about 700 Kb on my system.  It addresses Issue #371.  Comments on that issue raised the question that we are not sure what scripting tool(s) are guaranteed to exist on user or CRAN build machines.  I have relied only on Rscript, accessed by full path from CppCode/GNUmakefile.  All it does is make a local copy of R’s Makeconf and then strip out “-g” flags before `include`ing it.  Is it platform-independent?


packages/nimble/inst/CppCode/GNUmakefile
```diff
@@ -8,7 +8,12 @@ ifndef R_SHARE_DIR
 endif
 
 # Don't have to succeed in including this if we are calling cleanup in the nimble package directory.
--include $(R_HOME)/etc$(R_ARCH)/Makeconf
+# Following three lines result in much smaller libnimble.a or equivalent by removing -g flags
```
fritzo: Following *two* lines?

Consider converting the human-readable directions into a machine-readable `ifndef NIMBLE_DEBUG_SYMBOLS` or similar.

fritzo: Loos reasonable to me. Your build timed-out so I've restarted it.

paciorek:
Minor note -- I don't think you need $(shell ...) in 
```
$(shell $(R_HOME)/bin$(R_ARCH)/Rscript --vanilla ./customizeMakeconf.R)
```
since you're not assigning the output of the Rscript call to a variable.  I suspect just putting the ... meat directly is what we want.

packages/nimble/inst/CppCode/GNUmakefile
perrydv: The human-readable directions were intended more for ad-hoc testing on different operating systems.  But sure, turning debugging flags back on optionally could be useful.  I think we can add an option through the `configure.ac` to take an option to `R CMD install ...`.

Right, *two* lines.

Two additional changes I see: 

1. add `--vanilla` to `Rscript` call.

2. Add failsafe logic that checks if `Makeconf` exists locally and if not (meaning the `Rscript` job failed), use existing behavior of including Makeconf from R installation location.

perrydv:
@paciorek  (not sure why your comment doesn't have a reply window...): I think I tried what you suggest and it didn't work, but I may have been fumbling.  I agree it looks like it should be unnecessary, so please try to fix.

perrydv:
@paciorek I will to fix this PR so disregard my request that you look at it for now.

perrydv:
@paciorek @fritzo This PR now passes and I think could be ready to merge, so please review.  It looks like `make` has limited features outside of a rule, so I include the raw `Makeconf` followed by the script to copy and modify it locally and include that one.  If the second parts fail, the settings from the raw `Makeconf` will end up being used.  The `Rscript` call does not work for me without the `$(shell ... )`.

packages/nimble/inst/CppCode/customizeMakeconf.R
```diff
@@ -0,0 +1,9 @@
+try({
+    packageMakeconfFile <- file.path(R.home('etc'), 'Makeconf')
+    if(file.exists(packageMakeconfFile)) {
+        file.copy(packageMakeconfFile, 'Makeconf')
+        lines <- readLines('Makeconf')
+        lines <- gsub('-g', '', lines)
+        writeLines(lines, con = 'Makeconf')
+    }
+    })
```
fritzo: It will help users file bug reports if we don't silence failures. Consider logging
```r
tryCatch({
   # do stuff
}, error = function(e) {
  warning(paste('Failed to remove debug symbols from nimble:', e), sep = '\n')
})
```

fritzo:
I see `-g` in both linux and os x builds on travis, but I'm not sure I'm looking in the right place. Cold you paste before-and-after snippets of log files from travis in the description of this PR?

Linux [build job](https://travis-ci.org/nimble-dev/nimble/jobs/243739305)
```
linking libnimble.so
Linux   
g++  -shared -o libnimble.so  RcppUtils.o EigenTypedefs.o eigenUsingClasses.o RcppNimbleUtils.o Utils.o NamedObjects.o ModelClassUtils.o accessorClasses.o dists.o nimDists.o dllFinalizer.o nimbleGraph.o smartPtrs.o predefinedNimbleLists.o nimOptim.o 
rm *.o
** libs
g++  -I/home/travis/R-bin/lib/R/include -DNDEBUG -I../inst/include  -I  -I/home/travis/R-bin/include   -fpic  -g -O2  -c nimble.cpp -o nimble.o
g++ -shared -L/home/travis/R-bin/lib -o nimble.so nimble.o -L../inst/CppCode -lnimble -Wl,-rpath=/home/travis/R/Library/nimble/CppCode -L/home/travis/R-bin/lib/R/lib -lRlapack -L/home/travis/R-bin/lib/R/lib -lRblas
installing to /home/travis/R/Library/nimble/libs
```

OS X [build job](https://travis-ci.org/nimble-dev/nimble/jobs/243739306)
```
Making nimble libnimble.a
echo "RcppUtils.o EigenTypedefs.o eigenUsingClasses.o RcppNimbleUtils.o Utils.o NamedObjects.o ModelClassUtils.o accessorClasses.o dists.o nimDists.o dllFinalizer.o nimbleGraph.o smartPtrs.o predefinedNimbleLists.o nimOptim.o"
RcppUtils.o EigenTypedefs.o eigenUsingClasses.o RcppNimbleUtils.o Utils.o NamedObjects.o ModelClassUtils.o accessorClasses.o dists.o nimDists.o dllFinalizer.o nimbleGraph.o smartPtrs.o predefinedNimbleLists.o nimOptim.o
ar rcs libnimble.a RcppUtils.o EigenTypedefs.o eigenUsingClasses.o RcppNimbleUtils.o Utils.o NamedObjects.o ModelClassUtils.o accessorClasses.o dists.o nimDists.o dllFinalizer.o nimbleGraph.o smartPtrs.o predefinedNimbleLists.o nimOptim.o
rm RcppUtils.o EigenTypedefs.o eigenUsingClasses.o RcppNimbleUtils.o Utils.o NamedObjects.o ModelClassUtils.o accessorClasses.o dists.o nimDists.o dllFinalizer.o nimbleGraph.o smartPtrs.o predefinedNimbleLists.o nimOptim.o
** libs
clang++  -I/Library/Frameworks/R.framework/Resources/include -DNDEBUG -I../inst/include  -I  -I/usr/local/include   -fPIC  -Wall -g -O2  -c RcppUtils.cpp -o RcppUtils.o
clang++  -I/Library/Frameworks/R.framework/Resources/include -DNDEBUG -I../inst/include  -I  -I/usr/local/include   -fPIC  -Wall -g -O2  -c dists.cpp -o dists.o
clang++  -I/Library/Frameworks/R.framework/Resources/include -DNDEBUG -I../inst/include  -I  -I/usr/local/include   -fPIC  -Wall -g -O2  -c eigenUsingClasses.cpp -o eigenUsingClasses.o
clang++  -I/Library/Frameworks/R.framework/Resources/include -DNDEBUG -I../inst/include  -I  -I/usr/local/include   -fPIC  -Wall -g -O2  -c nimble.cpp -o nimble.o
clang++  -I/Library/Frameworks/R.framework/Resources/include -DNDEBUG -I../inst/include  -I  -I/usr/local/include   -fPIC  -Wall -g -O2  -c nimbleGraph.cpp -o nimbleGraph.o
clang++ -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/Library/Frameworks/R.framework/Resources/lib -L/usr/local/lib -o nimble.so RcppUtils.o dists.o eigenUsingClasses.o nimble.o nimbleGraph.o -L/Library/Frameworks/R.framework/Resources/lib -lRlapack -L/Library/Frameworks/R.framework/Resources/lib -lRblas -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
installing to /Users/travis/R/Library/nimble/libs
```

paciorek:
I'm getting these warnings (and more that look like them) when installing on Linux:
```
Building libnimble
Makeconf:159: warning: overriding recipe for target '.c.o'
/usr/lib/R/etc/Makeconf:159: warning: ignoring old recipe for target '.c.o'
Makeconf:161: warning: overriding recipe for target '.c.d'
/usr/lib/R/etc/Makeconf:161: warning: ignoring old recipe for target '.c.d'
Makeconf:164: warning: overriding recipe for target '.m.d'
/usr/lib/R/etc/Makeconf:164: warning: ignoring old recipe for target '.m.d'
```

That said, this looks to work well on my Linux box -- libnimble.so goes from 11 MB to 600KB and libnimble.a goes from 20 MB  to 1.1 MB, so it helps with libnimble.so as well, which will be nice in terms of avoiding R CMD check NOTEs.

paciorek:
Per comment in email from @perrydv, removing 
```
-include $(R_HOME)/etc$(R_ARCH)/Makeconf
```
from inst/Cppcode/GNUMakefile causes the Makeconf warnings noted above to go away.

perrydv:
@NLMichaud Can you try this branch on Windows? @paciorek do you want to try CRAN check on this branch?

NLMichaud:
@perrydv This branch was installed, and mcmc and filtering tests were run without error on Windows.

paciorek:
Just ran R CMD check and this LGTM.

paciorek:
Turns out that as implemented this causes major installation problems on Windows, CentOS, Solaris, possibly others. Comments from @duncantl for when we revisit this:

It seems like there is a template Makeconf file in inst/CppCode.
The attempt to copy the local $R_HOME/etc$R_ARCH/Makeconf
won't work with file.copy() unless overwrite = TRUE is added to the call.
As a result, we are dealing with a Makeconf file that was created
on an entirely different machine with very different configure options.
So this will cause major problems.
So adding
   file.copy(packageMakeconfFile, 'Makeconf', overwrite = TRUE)
will fix this for most cases.

Sorry - I am talking about the code in inst/CppCode/customizeMakeconf.R  which is called from the GNUmakefile.

And the call to try()  in that script will mean that if it fails for any reason, we will be working with a Makeconf that is probably not for that machine

The call to R CMD build calls the cleanup script. That in turn calls make in inst/CppCode and the GNUmakefile automatically creates the
Makeconf.
I believe if inst/CppCode/Makeconf is actually wanted, it should be created at configure time.

fritzo:
When we revist this, Dirk has a couple solutions including:
- http://dirk.eddelbuettel.com/blog/2017/08/14/#009_compact_shared_libraries
- http://dirk.eddelbuettel.com/blog/2017/08/20/#010_stripping_shared_libraries

paciorek:
@perrydv I think you were looping back to this. I also independently noticed Dirk's posts that Fritz refers to above, so thought I'd point you to those also.

Issue #378, opened 2017-05-24
Make options(error = recover) work when compiler fails


fritzo:
Nimble's `try()` logic makes it difficult for developers to diagnose errors. Consider avoiding error trapping altogether, or allowing a developer flag to avoid error trapping.

Issue #374, opened 2017-05-24
clean up some files from repo above package


perrydv:
Can we get rid of the `MCMC_comparisons`... and `MCMC_suite`... files?  They are old, once seemed sensible to include, but now seem like junk.

fritzo:
:+1: to removing vestigial files and thereby reduce maintenance burden. @paciorek suggested that `Makefile.nimble` may also be removed.

paciorek:
They're possibly still useful to me to remind me of what we need to comment
out for CRAN  in terms of stan, but I made a copy of them elsewhere so feel
free to remove if you like.

Fritz, as far as Makefile.nimble, I suppose we need to figure out what to
name the updated Makefile we discussed last week. I think GNUmakefile may
take precedence over "Makefile" if just invoking make. I guess we could
include the configure target as well as all the targets we've discussed in
GNUmakefile, but we don't want to have the configure target called unless a
dev specifically requests it.

Issue #365, opened 2017-05-20
Support keyword next


fritzo:
This just translates to `continue` in C++

Issue #314, opened 2017-04-30
check on translation of NaN from R to C++


paciorek:
If we have a NaN on the R side, this can cause issues in compilation.

E.g., giving a dinvgamma(-1/2,0) results in getParam('mean') or getParam('sd') being NaN because it involves 0/0 or 0*Inf.  The NaN gets propagated into C++ where it cannot be interpreted.

Interestingly, apart from this, I think use of an improper invgamma will actually work with our conjugacy system. improper gamma actually already works. 

This is a bit of a corner case, but OTOH having an NaN not cause compilation failure is probably worth cleaning up at some point.

```
code = nimbleCode({
  vv ~ dinvgamma(-1/2, 0)
  for(i in 1:100)
    y[i] ~ dnorm(0, var = vv)
})

m = nimbleModel(code, data = list(y=rnorm(100)), inits = list(vv = 1))
conf = configureMCMC(m)

mcmc = buildMCMC(conf)
cc = compileNimble(mcmc, m, showCompilerOutput = TRUE)
P_9_code_nfCode.cpp:54:16: error: use of undeclared identifier 'NaN'
PARAMANSWER_ = NaN;
               ^
```

PR #433, opened 2017-06-14
Omnibus

perrydv:
This PR fixes six old and/or minor issues:

1. Issue #378: I made options(error = recover) work for the main case of interest: size processing called from `RCfunProcessing::process`.  Proposed solution is to defer to R’s native handling if options(‘error’) is non-null, and otherwise to use our existing  `tryCatch` clause.

2. Issue #374: I removed the `MCMC_*` files from the repo above the R package level.  According to the issue, @paciorek has these files saved separately for when they are needed but they do not need to live in the repo.

3. Issue #365: support for keyword `next` was added. A test was added to `test-misc.R`.

4. Issue #314: `NaN` is now successfully written to C++ from R via a `nimble_NaN()` function that attempts to use `std::numeric_limits::quiet_NaN()` when available and otherwise computes a NaN as 0./0.  A test was added to `test-misc.R`.

5. Issue #302: dots in R variable names in the DSL (e.g. `a.b`) are now handled by processing through our heavily used utility `Rname2CppName`. This occurs at the final step of writing C++ code.  The mangled result is `a_dot_b`.

6. Issue #92: This old issue was to support `pi` in the DSL.  Brief discussion led to a sticking point and it was left to go stale.  In R, `pi` exists as a constant that can be overwritten by the user.  Proposed solution to yield similar compiled behavior is that if a variable `pi` is used, it is always initialized to the natural constant `pi`, after which it behaves like any other local variable and hence can be overwritten.  Behavior will be limited by static typing: if used first on RHS, it will be a scalar double and can only be over-written as such.  Three tests were added to `test-misc.R`.

**Note that `pi` still does not work in model code.** We can open that as a separate issue if desired: it involves model processing and was not handled as part of Issue #92.


packages/nimble/inst/CppCode/Utils.cpp
```diff
@@ -67,3 +67,9 @@ double pairmin(double x1, double x2) {return(x1 < x2 ? x1 : x2);}
 int nimStep(double x) { return(x >= 0 ? 1 : 0);}
 double cube(double x) {return(x*x*x);}
 double inprod(double v1, double v2) {return(v1*v2);}
+
+double nimble_NaN() {
```
fritzo: Consider making this inline and optimizing
```c++
inline double nimble_NaN() {
  return std::numeric_limits<double>::has_quiet_NaN
    ? std::numeric_limits<double>::quiet_NaN()
    : (0. / 0.);
}
```

packages/nimble/R/RCfunction_compile.R
```diff
@@ -237,26 +237,29 @@ RCfunProcessing <- setRefClass('RCfunProcessing',
...
-                                                                         sep = '\n')
-                                                    }
-                                                    stop(message, call. = FALSE)
-                                                })
+
+                                       ## exprClasses_setSizes contains lots of reticulated error trapping.
+                                       ## If options('error') is not NULL (typically options(error = recover)), let that take precedent for error trapping of exprClasses_setSizes.
+                                       ## Otherwise, wrap the setSizes call in our own error-trapping that outputs: any local message from a stop() inside a size handler,
+                                       ##    a message from here showing the larger block of code in which the error occurred, and, if nimbleOptions('verboseErrors') is TRUE, the call stack. 
+                                       if(!is.null(options('error'))) exprClasses_setSizes(compileInfo$nimExpr, compileInfo$newLocalSymTab, compileInfo$typeEnv)
```
fritzo: Good idea! Maybe we can get rid of `verboseErrors` now.

fritzo:
:+1:

fritzo:
Remember to move `nimble_NaN()` from to the header file if you make it inline.

paciorek:
No concerns on my part; didn't look very deeply.

PR #434, opened 2017-06-14
test_coreRfeature_batch, like test_coreRfeature for batches

perrydv:
Not yet ready to merge.  Still has some cleanup to do but I’m making PR for Travis testing.

This PR quiets some tests and speeds up others.  It is incremental progress on our test suite.

Quieting: The following tests now have opening and closing code to set `options(warn = -1)` and `nimbleOptions(verbose = FALSE)` and restore original settings upon completion:

`graphStructure`, `refactorCompilationSteps`, `returnTypeErrorTrapping`, `coreR`, `numericTypes`, `allocation`

Speeding up: There is a new `test_coreRfeature_batch` that compiles many test cases together as methods of one large `nimbleFunction`.  This is now used for `coreR`, `numericTypes`, and `allocation`.  It appears to yield substantial speedup, but we’ll have to see on benchmark machine.

### DIFF
--- a/packages/nimble/inst/tests/test-allocation.R
+++ b/packages/nimble/inst/tests/test-allocation.R
@@ -3,14 +3,18 @@
 library(nimble)
 source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
 
+RwarnLevel <- options('warn')$warn
+options(warn = -1)
+nimbleVerboseSetting <- nimbleOptions('verbose')
+nimbleOptions(verbose = TRUE)
+
 
 numericTests <- list(
     list(name = 'numeric: length',
          expr = quote(out <- numeric(5)),
          outputType = quote(double(1) ),
          args = list(),
-         setArgVals = quote({}),
-         safeCompilerFail = FALSE
+         setArgVals = quote({})
          ),
     list(name = 'numeric: length 0',
          expr = quote({
@@ -26,8 +30,7 @@ numericTests <- list(
          expr = quote(out <- nimNumeric(5, value = 2, recycle = FALSE, fillZeros = TRUE)),
          outputType = quote(double(1) ),
          args = list(),
-         setArgVals = quote({}),
-         safeCompilerFail = FALSE
+         setArgVals = quote({})
          ),
     list(name = 'numeric: length, scalar value, init=FALSE',
          expr = quote({out <- nimNumeric(5, value = 1.5, init = FALSE);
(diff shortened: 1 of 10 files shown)


perrydv:
I'd like to merge this since it appears to reduce testing time by 20-30 minutes and we need that to finish Travis testing (until we split that grid more) and get other PR's unblocked.  Any suggestions @paciorek @fritzo ?

perrydv:
Oh, I guess I noted it is not ready to merge, but that is only because I wanted to clean up some commenting out of old code, which I will do in any case.

packages/nimble/inst/tests/test_utils.R
```diff
@@ -56,6 +56,109 @@ indexNames <- function(x) {
...
+        seedToUse <- if(is.null(input[['seed']])) 31415927 else input[['seed']]
+        set.seed(seedToUse)
+        eval(input$expr, envir = evalEnv)
+        savedOutputs <- as.list(evalEnv)
+        list2env(savedArgs, envir = evalEnv)
+        ## with R ref classes, lookup of methods via `[[` does not work until it has been done via `$`
+        ## force it via `$` here to allow simpler syntax below
+        forceFind <- eval(substitute(nfR$FUNNAME, list(FUNNAME = as.name(funName))))
+        forceFind <- eval(substitute(nfC$FUNNAME, list(FUNNAME = as.name(funName))))
+        if(nArgs == 5) {
```
fritzo: Can you eliminate this poor man's `switch` using `do.call()`?
```r
set.seed(seedToUse)
out_nfR = do.call(nfR[[funName]], evalEnv)
list2env(savedArgs, envir = evalEnv)
set.seed(seedToUse)
out_nfC = do.call(nfC[[funName]], evalEnv)
```

packages/nimble/inst/tests/test-allocation.R
```diff
@@ -3,14 +3,18 @@
 library(nimble)
 source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
 
+RwarnLevel <- options('warn')$warn
```
fritzo: Long range dependencies like this can be difficult to maintain. Since you've already refactored to use a `test_that()` context elsewhere, this might be a good opportunity to introduce a testing helper that can use `on.exit` to clean up after itself.

As Hadley observed in his [testthat paper](https://journal.r-project.org/archive/2011-1/RJournal_2011-1_Wickham.pdf):

> When you use these actions in tests, you’ll need to clean up after yourself. Many other testing packages have set-up and teardown methods that are run automatically before and after each test. These are not so important with testthat because you can create objects outside of the tests and rely on R’s copy-onmodify semantics to keep them unchanged between test runs. To clean up other actions you can use regular R functions.

fritzo:
:+1: I only had two general cleanliness comments.

PR #435, opened 2017-06-15
Add resetFunctions flag to buildMCEM.

NLMichaud:
If buildMCEM detects that the `model` argument has already been compiled, it will now set `resetFunctions = TRUE` when compiling it's latent state MCMC algorithm.


perrydv:
👍

PR #441, opened 2017-06-15
nimDerivs for uncompiled nimbleFunctions.

NLMichaud:
This PR will add the DSL function `nimDerivs()` for use in uncompiled nimbleFunctions.

`nimDerivs()` takes a function call as an argument (e.g. `nimDerivs(nimFxn(x, y))`, and returns the derivatives of that function with respect to the arguments specified in the `wrt` argument to `nimDerivs()`.  These arguments can include indices (e.g. `wrt = c('x[1:2, 1]')` to take derivatives with respect to the elements in the first two rows and first column of the matrix argument `x` to `nimFxn`).  By default, if no `wrt` argument is provided to `nimDerivs()`, derivatives will be taken with respect to all argument to `nimFxn`.

  The code for `nimDerivs()` currently exists in `nimbleFunction_Rderivs.R`.  Derivatives of uncompiled nimbleFunctions are taken using the `numDeriv` R package.  As such, this package must be installed in order for derivatives of uncompiled functions to be taken.

If the function provided to `nimDerivs` is a call to `calculate(model, nodes)`, the derivative can be taken using the chain rule.  This is controlled by the `chainRuleDerivs` logical argument to `nimDerivs()`.  The code for taking the derivative of a call to `calculate(model, nodes)` via the chain rule is contained in the function `nimDerivs_calculate()` in `nimbleFunction_Rderivs.R`.  Chain rule calculation is facilitated by the creation of a `nimDerivsInfoClass` object, which contains information about the `nodes` argument that was provided in the call to `calculate(model, nodes)`, and how these `nodes` relate to the `wrt` argument provided to `nimDerivs`.  

Below is an example of calling `nimDerivs()` on a `calculate` call:
```
library(nimble)
nimbleOptions(experimentalEnableDerivs = TRUE)

ADtestCode <- nimbleCode({
  a ~ dnorm(c, 1)
  c <- b + 1
  b ~ dnorm(2, 1)
})
ADtestMod <- nimbleModel(code = ADtestCode, inits = list(a = 3, b = 1))

## Calculate using chain rule
derivList <- nimDerivs(calculate(ADtestMod, ADtestMod$getDependencies('a')), wrt = c('a', 'b'))
```
Note that the accuracy of this chain rule derivative can be checked by comparison to a derivative taken without the use of the chain rule:
```
tstFxn <- function(a, b){
  origa <- ADtestMod1$a
  ADtestMod1$a <- a
  origb <- ADtestMod1$b
  ADtestMod1$b <- b
  
  outVal <- calculate(ADtestMod1, ADtestMod$getDependencies('a'))
  
  ADtestMod1$b <- origb
  ADtestMod1$a <- origa
  
  return(outVal)
}
deriv2 <- nimDerivs(tstFxn(3, 1))
```

In fact, the above comparison is the basis for all of the currently implemented testing for chain rule derivatives.  The function `test_ADModelCalculate()` in `test_utils.R` compares derivatives of calls to `model$calculate(nodes)` between two methods: derivatives of a wrapper function, such as the `tstFxn ` above, and derivatives calculated using the chain rule.   The derivatives of this wrapper function (e.g. `nimDerivs(testFxn(3, 1))`) are compared to the derivatives calculated using the chain rule (e.g. `nimDerivs(ADtestMod$calculate(ADtestMod$getDependencies('b')), wrt = c('a', 'b'))`) to make sure that both answers align.


packages/nimble/R/nimbleFunction_Rderivs.R
```diff
@@ -0,0 +1,422 @@
...
+
+
+#' Nimble Derivatives
+#' 
+#' EXPERIMENTAL Computes the value, gradient, and Hessian of a given  \code{nimbleFunction} method.  
+#' 
+#' @param nimFxn a call to a \code{nimbleFunction} method with arguments included.
+#' @param order an integer vector with values within the set {0, 1, 2}, corresponding to whether the function value, gradient, and Hessian should be returned respectively.
+#' 
+#' @export
```
fritzo: Add `@return`

NLMichaud: Just added some more documentation for this function, thanks @fritzo!

packages/nimble/R/nimbleFunction_nodeFunctionNew.R
```diff
@@ -8,13 +8,13 @@ nodeFunctionNew <- function(LHS, RHS, name = NA, altParams, bounds, parentsSizeA
     boundsRep <- lapply(bounds, nndf_replaceSetupOutputsWithIndexedNodeInfo, setupOutputLabels)
     logProbNodeExprRep <- nndf_replaceSetupOutputsWithIndexedNodeInfo(logProbNodeExpr, setupOutputLabels)
     if(nimbleOptions('experimentalEnableDerivs')){
-      parents <- names(parentsSizeAndDims)
```
fritzo: nit: revert this trailing whitespace

fritzo:
@NLMichaud How soon can you merge this branch? Could you merge the R implementation of `nimDerivs()` before the CppAD part?

I ask because I'd like to implement gradient calculations using tensorflow. Basic tensorflow support is already merged into devel, all I would need is the R implementation of `nimDerivs()` to test the tensorflow implementation of `nimDerivs()`. (In theory this could be done on a branch that cherry picked bits of tensorflow and `nimDerivs`; in practice the resulting merge conflicts would weeks of painful surgery to repair.)

packages/nimble/R/cppDefs_ADtools.R
```diff
@@ -349,7 +338,7 @@ makeADargumentTransferFunction <- function(newFunName = 'arguments2cppad', targe
 ## The code this generates is embedded in the ADtapingFunction made by makeADtapingFunction
 ##
 ## Note this does some work similar to BUGScontextClass::embedCodeInForLoop
-makeCopyingCodeBlock <- function(LHSvar, RHSvar, indexList, indicesRHS = TRUE, incrementIndex, isNode) {
+makeCopyingCodeBlock <- function(LHSvar, RHSvar, indexList, indicesRHS = TRUE, incrementIndex, isNode = FALSE) {
```
fritzo: aha, I bet this would fix some of the errors I've seen in the `gradients` branch.

packages/nimble/R/cppDefs_core.R
```diff
@@ -230,6 +230,10 @@ cppCodeBlock <- setRefClass('cppCodeBlock',
                                           cppADCode = 'ANY', generatorSymTab = 'ANY'),			#'logical'),
                             methods = list(
                                 generate = function(indent = '', ...) {
+                                    if(identical(cppADCode, TRUE)){
```
fritzo: Why does it not suffice to simplify to `if (cppADCode)` ?

NLMichaud:
@perrydv @fritzo I believe this PR is ready to be merged, pending review.  Any comments would be very appreciated! I also updated the note that goes along with this PR to reflect the current state of `nimDerivs()`.

packages/nimble/R/cppDefs_core.R
```diff
@@ -230,6 +230,10 @@ cppCodeBlock <- setRefClass('cppCodeBlock',
...
+                                      nimbleUserNamespace$cppADCode <- TRUE
+                                      on.exit(nimbleUserNamespace$cppADCode <- FALSE)
```
fritzo: Will this work when blocks are nested? It seems to me that you'd want to avoid setting if it's already true, otherwise you might erroneously set it to FALSE on exit.
```r
if (!nimbleUserNamespace$cppADCode) {
    nimbleUserNamespace$cppADCode <- TRUE
    on.exit(nimbleUserNamespace$cppADCode <- FALSE
}
```

packages/nimble/R/genCpp_exprClass.R
```diff
@@ -35,7 +34,6 @@ exprClass <- setRefClass('exprClass',
             toEigenize <<- 'unknown'
             assertions <<- list()
             eigMatrix <<- logical()
-            cppADCode <<- FALSE
```
fritzo: Why don't you set the default value any more?

fritzo: Oh, is this because you've moved the state to a global variable? (if so, I'll have to fix my tensorflow code)

packages/nimble/R/nimbleFunction_nodeFunctionNew.R
```diff
@@ -17,12 +17,12 @@ nodeFunctionNew <- function(LHS, RHS, name = NA, altParams, bounds, parentsSizeA
     
     if(nimbleOptions('experimentalEnableDerivs')){
       parents <- names(parentsSizeAndDims)
-      parentIndexInfoList <- nndf_extractNodeIndices(LHSrep, parents)
-      parentIndexInfoList <- nndf_extractNodeIndices(RHSrep, parents, indexExprList = parentIndexInfoList)
-      for(i in seq_along(parentIndexInfoList)){
-        for(j in seq_along(parentIndexInfoList[[i]])){
-          parentsSizeAndDims[[names(parentIndexInfoList)[i]]][[j]]$indexExpr <- parentIndexInfoList[[i]][[j]]$indexExpr 
-        }
+      parentIndexInfoList <- nndf_extractNodeIndices(LHSrep, parents)
```
fritzo: nit: remove trailing whitespace

danielturek:
@NLMichaud out of curiosity, is this PR still only for **un-compiled** functions, only?

Or has it been extended to compiled functions?

Thanks!

packages/nimble/R/genCpp_exprClass.R
NLMichaud: @fritzo That's correct.  A suggestion @perrydv had that I agreed with was to set a global variable instead of a per-expression variable, as the latter seems somewhat more involved than may be necessary.

packages/nimble/R/cppDefs_core.R
NLMichaud: Good logical catch @fritzo!  That makes total sense to me.

packages/nimble/inst/tests/test-ADmodels.R
```diff
@@ -2,18 +2,82 @@ source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
...
-            })
-            
-            ADMod1 <- nimbleModel(code = ADCode1, data = list(y = numeric(2)), dimensions = list(y = c(2)))
-            temporarilyAssignInGlobalEnv(ADMod1)  
-            cADMod1 <- compileNimble(ADMod1)
-            ## derivs(calculate(model, nodes)) not fully implemented yet, this tests only that compilation is successful.
-            }
-)
\ No newline at end of file
+ADCode1 <- nimbleCode({
```
fritzo: Wrap each test in a `test_that()` and avoid the `ADCode<N> for each N` pattern. Otherwise you're leaking into the global namespace and risking copy-and-paste errors. the `temporarilyAssignInGlobalEnv()` helper can make this easier.

packages/nimble/inst/tests/test-ADfunctions.R
```diff
@@ -6,26 +6,51 @@ test_that('Derivatives of dnorm function correctly.',
...
+    expect_equal(cderivs$value, Rderivs$value)
+    expect_equal(cderivs$gradient, Rderivs$gradient)
+    expect_equal(cderivs$hessian, Rderivs$hessian)
   }
 )
+
+
+
+
+ADfun1 <- nimbleFunction(
```
fritzo: Wrap in a `test_that()`.

NLMichaud:
@danielturek This PR largely addresses derivatives of un-compiled `nimbleFunctions`.  There is some support for derivatives of compiled `nimbleFunctions`, but compiled calls to `nimDerivs(model$calculate(nodes))` are not yet supported (but are next to go, so should be soon!)

packages/nimble/inst/tests/test-ADfunctions.R
```diff
@@ -6,26 +6,51 @@ test_that('Derivatives of dnorm function correctly.',
...
+    cderivs <- cADfunInst$run(x)
+    expect_equal(cderivs$value, Rderivs$value)
+    expect_equal(cderivs$gradient, Rderivs$gradient)
+    expect_equal(cderivs$hessian, Rderivs$hessian)
   }
 )
+
+
+
+
```
fritzo: Thanks for adding this. Could you add a couple more examples of `R^n->R` functions and gradient computations, so that I can fork those for testing the Tensorflow version?

packages/nimble/R/genCpp_exprClass.R
fritzo: I wish this thread could be marked 'resolved'. Vote for that feature here: https://github.com/dear-github/dear-github/issues/154

packages/nimble/R/cppDefs_core.R
perrydv: If `cppADCode` is some malformed object, then `if(cppADCode)` will error instead of being handled as `FALSE`.

R provides `isTRUE(cppADcode)` as an option.

packages/nimble/R/BUGS_getDependencies.R
```diff
@@ -62,3 +62,310 @@ gd_getDependencies_IDs <- function(graph, maps, nodes, omit, downstream) {
...
+          }
+          if(length(lineWRTArgsAsCharacters) < i){
+            lineWRTArgsAsCharacters[[i]] <<- NA
+          }
+        }
+      },
+
+      # ### A function that substitutes correct values of unrolledIndicesMatrix
+      # ### into symbolicParentNodesReplaced.
+      recurseReplaceIndices = function(code, unrolledIndicesRow){
```
perrydv: 2 questions:
- could this function be a non-method?
- would `substitute` work?  Is the goal to replace in `code` the values by name in `unrolledIndicesRow`?
```{r}
vals <- data.frame(a = 1, b = 2)
code <- quote(x[a, k[b]])
substitute(code, as.list(vals))
eval(substitute(substitute(code, as.list(vals)), list(code = code)))
```

I am not sure if I have captured the idea correctly.

perrydv:
@NLMichaud I made a few comments on specific lines.  Other comments:

- There are two definitions of `nndf_extractNodeIndices` in `nimbleFunction_nodeFunctionNew.R`.  (As far as I can tell, the function is only called when `enableExperimentalDerivs` is `TRUE`).

- As a general practice, we should try to avoid making reference classes with many and large methods, which unfortunately we do have a bunch of.  This is the R's advice.  Could you look at `nimDerivsInfoClass` and see if some methods, or code within methods, could be migrated to stand-alone functions?  If their use of the class internally is fairly light, they could take a class object as an argument and access its members.

packages/nimble/R/BUGS_getDependencies.R
NLMichaud: Yes, `substitute` works perfectly here, so the method can be removed.  Thanks for the suggestion!

NLMichaud:
- Thanks for catching that, I removed the duplicated function.

- The methods can certainly be migrated to stand-alone functions.  I had used the reference class as more of a data structure for the dependency information and the functions that were required to produce it, but the methods could be made into stand alone functions that would take, e.g., a reference class of dependency data as an argument.  Would we still want to use a reference class to store the dependency data?  Or should we store the data as a simple `list()` object?

perrydv:
@NLMichaud I'm not sure of the reference class vs. list question in this case.  Another option is an R6Class.  If it is passed around a lot and not modified, then a list is fine.  When not modified, R is clever about copy semantics and doesn't actually until it becomes necessary, which won't happen if it's not modified.  If it is passed around a lot and it is helpful to modify it in place (i.e. the changes should persist but without returning the modified version to the calling function), then a reference class works well.  If it's really not clear, let's use a reference class for now.

PR #445, opened 2017-06-17
Fixes #443 by ensuring that use of unique in splitVertices operates on a vector rather than matrix or array

### LINKED ISSUE #443: gap in vertex numbering because of issue in splitVertices
This example reveals that when `pr[1:2,1:2]` is processed in splitVertices(), we have an extra skip in the vertex numbering. Ultimately, this results in numVertices being too large and graphID_2_declID having an extra element (and perhaps other effects on maps -- haven't looked carefully).

```
code = nimbleCode({
    w[1:2] ~ dmnorm(zeros[1:2],pr[1:2,1:2])
})
model = nimbleModel(code)
```

This occurs because in 
```
uniqueCurrentVertexIDs <- unique(currentVertexIDblock)
```
currentVertexIDblock is a 2x2 matrix of NAs and unique returns a 1x2 matrix of NAs instead of a single value. This seems like it would occur any time we have a matrix or array of more than 1-dimension (I guess unless some splitting is occurring). 

I think we simply want 
```
uniqueCurrentVertexIDs <- unique(c(currentVertexIDblock))
```

but would be good for @perrydv to take a second look.

perrydv:
This PR uses @paciorek ’s suggestion for fixing Issue #443 .  This appears to be a harmless bug but would be good to clean up.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -1406,7 +1406,7 @@ splitVertices <- function(var2vertexID, unrolledBUGSindices, indexExprs = NULL, 
         ## construct argList
         for(iRow in iRowRange) {
             currentVertexIDblock <- eval(accessExpr)
-            uniqueCurrentVertexIDs <- unique(currentVertexIDblock)
+            uniqueCurrentVertexIDs <- unique(c(currentVertexIDblock))
             if(length(uniqueCurrentVertexIDs)==1) { ## current block has only 1 ID
                 if( is.na(uniqueCurrentVertexIDs[1]) |      ## It's all unassigned OR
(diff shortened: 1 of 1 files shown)


fritzo:
I just learned that you can automatically close issues by using standard naming of pull request titles:
https://help.github.com/articles/closing-issues-via-commit-messages/
So I think you could title this PR "Fixes #443 by ensuring that unique inside splitVertices operates on a vector rather than a matrix or array". Let's see if it works :smile:

perrydv:
@paciorek Looks like this passes tests.  Thanks for that suggestion for future practice @fritzo

paciorek:
Looks like @fritzo suggestion worked.

PR #447, opened 2017-06-19
add name to every package-defined nimbleFunction

perrydv:
This PR adds a name argument to every `nimbleFunction` defined in the package source code.  The only exception is for conjugate samplers, since even the `nimbleFunction` code for those is code-generated and I did not take the time to ensure a safe name choice.  These names should make generated C++ easier to read.


fritzo:
:+1: Thanks, Perry!

PR #450, opened 2017-06-22
Add --parallel option to run_tests.R

fritzo:
This adds a `--parallel` option to `run_tests.R` and adds a `test` target to the `nimble/packages/Makefile` to run tests in parallel.

### DIFF
--- a/packages/Makefile
+++ b/packages/Makefile
@@ -14,12 +14,17 @@ nimble_$(VERSION).tar.gz: nimble/configure FORCE
 
 build: nimble_$(VERSION).tar.gz
 
+install: nimble_$(VERSION).tar.gz
+	R CMD INSTALL nimble_$(VERSION).tar.gz
+
 check: build FORCE
 	R CMD check --as-cran nimble_$(VERSION).tar.gz
 
 test: FORCE
 	cd .. ; ./run_tests.R --parallel
 
+# We are progressively delegating style to clang-format.
+# Some older C++ files are still styled manually.
 clang-format: FORCE
 	clang-format -style=file \
 	  nimble/inst/include/nimble/NimArrBase.h \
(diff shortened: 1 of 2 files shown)


fritzo:
This is weird, when I run `make test` to run the tests in parallel, a bunch of the tests die with `Error: Failed to create the shared library` and `test-distributions.R` dies with `Error in loadNamespace(name) : there is no package called ‘mvtnorm’`. The same tests run fine when not parallelized :confused:

paciorek:
A few questions:

 - is there a reason you only report total number of test failures if running in parallel?
 - should run_tests.R be in the same directory as Makefile? Not clear to me it should be the next level up. (That said, not clear to me what we want at each of those two levels; perhaps some stuff in `packages` should go one level up.)
 - in run_tests.R, this doesn't look OS-agnostic to me and I think it may fail on Windows: `logDir <- '/tmp/log/nimble'`.  See `tempdir()` and `file.path()`. 
 - same concern about `/usr/bin/time` and (possibly) use of Rscript. Perhaps check for Windows and bail if not running under testthat?
 - Makefile looks good - thanks!  Only comment is that would be nice to have an "install" target.

fritzo:
Thanks for the careful revuew, @paciorek ! Let me address your comments:

1. `run_tests.R` reports the total number of failures at the end, and each individual test reports whether it PASSED or FAILED, and on failure prints the path to the log files the user can read to diagnose errors (hence the avoidance of `tmpdir()`).
2. My original thinking was that `run_tests.R` is called by `.travis.yml` and `appveyor.yml`, and `update_test_times.R` downloads and reads travis log files. Therefore I put those scripts next to the CI config files. That said, I don't have strong opinions, and happy to move around the files or even get rid of the `packages` layer entirely, in a separate PR.
3. `run_tests.R` makes no attempt at working on Windows, but it conditionally set `runner` so that it can work on both OS X and Linux. If there is demand for Windows, we can do that in a follow-up PR.
4. ditto, `runner` avoids `/usr/bin/time` if it cannot be found (e.g. on OS X).
5. I'll add an `install` target.

paciorek:
Ok, that all sounds fine to me.

perrydv:
If it's ok with @paciorek, LGTM too.  Reminder that we should run test content on Windows locally prior to release.

PR #325, opened 2017-05-04
Define -DR_NO_REMAP and prefix R symbols with Rf_

fritzo:
This cleans up our C++ global namespace by instructing R not to export 176 internal functions. It requires
- passing `-DR_NO_REMAP` to all C++ compilers (both during NIMBLE install and during `compileNimble`)
- using the fully qualified forms of R functions in C++ (e.g. `Rf_length` instead of `length`)
- using the fully qualified forms of R functions in R template code used to generate C++

**Fixes:** #307 

**Question for reviewer:** What do we need to do to test this on all platforms?

Issue #126, opened 2016-03-29
look into accessing Eigen via RcppEigen


paciorek:
Can use LinkingTo RcppEigen without dependency on Rcpp.

Email from Dirk:
I noticed (from your very detailed Description) that you embed Eigen.  Just
in case you were unaware: you could use LinkingTo: RcppEigen and also obtain
without replicating all the content.  This does not even a dependency on
Rcpp.   A few packages (OpenMX, TMB, oapackages, ...) already do this.

That said, seeing you RcppUtils.{h,cpp} I would of course be equally happy to
help you convert to Rcpp if that was of interest, but that is an entirely
separate and unrelated discussion :)

Ping me anytime by email if you have questions I could help with.

PR #452, opened 2017-06-23
Call external C functions and call R functions from compiled code

perrydv:
This PR proposes two new features, both refreshed from a draft branch from January 2017:

1. Allow a `nimbleFunction` to wrap a call to user-provided compiled code.  Currently this takes a .o file and the user must provide the function prototype (equivalent to .h content) in `nimbleExternalCall`.  This generates a `nimbleFunction` that includes new keywords `nimbleConvert` and `nimbleUncovert` to convert from `NimArr<>` types to simple C pointer types and back.  Currently that is the only conversion supported, but more (e.g. `std::vector<>`) could be easily added.  A limitation in the current draft is that only a scalar can be returned, but non-scalar return information can be handled through pointer arguments. Are there other useful ways a user would want to provide external code?

2. Allow a `nimbleFunction` to wrap a call to an arbitrary R function, even from compiled code.  This would allow a user to rapidly prototype an idea in R and embed it in compiled nimble code.  It is accomplished by decorating an R function with `nimbleRcall`.  This generates a `nimbleFunction` that includes new keywords (some R internals and a `nimVerbatim` for arbitrary code).  The resulting code constructs an R language object (`LANGSXP`), converts from `NimArr<>`s to `SEXP`s, sets the arguments to the `LANGSXP` from the `SEXP`s, calls the R evaluator, and converts the result back to a `NimArr<>`.

The test file `test-externalCalls.R` is intended as more of a demo than a test.  Please step through it manually if you want to see how these features are set up.  It is on the Travis-CI omission list because I want that testing to see if anything else was broken by these features, but the features themselves are still experimental.

Feedback is needed on both the design and implementation of these features.

### Updated PR

This PR now includes draft roxygen and User Manual documentation.

The original PR had bugs for handling return values of `nimbleExternalCall`.  That has been fixed.  

The tests in `test-externalCalls.R` have been fixed up.  The PR is now ready to go pending review.


packages/nimble/R/externalCalls.R
```diff
@@ -0,0 +1,87 @@
...
+    }
+    allLines <- c(list(as.name("{")), convertLines, list(externalCallLine), unconvertLines, returnLines)
+    body(fun) <- as.call(allLines)
+    ans <- quote(nimbleFunction(run = fun))
+    ans <- eval(ans)
+    ## right here stick header information into the nfMethodRC
+    environment(ans)$nfMethodRCobject$externalHincludes <- paste0('\"',headerFile,'\"')
+    oFile <- gsub('\\.c', '', oFile)
+    oFile <- gsub('\\.o', '', oFile)
+    oFile <- paste0(oFile, '.cpp') ## This later gets stripped and replaced with .o but it must be .cpp to be stripped (until we someday generalize that handling)
```
fritzo: Rename this variable to `cppFile` if it is going to store `something.cpp`.

packages/nimble/R/externalCalls.R
```diff
@@ -0,0 +1,87 @@
+
```
fritzo: Document this function. What are its arguments? What does it return?

fritzo: Documentation is especially important if you are requesting design review from reviewers.

packages/nimble/R/externalCalls.R
```diff
@@ -0,0 +1,87 @@
...
+nimbleExternalCall <- function(fun, returnType = void(), Cfun, headerFile, oFile) {
```
fritzo: 1. Passing in a `fun` seems like a hack. Why not simply pass in `name`, `argTypes`, and `returnType`?

2. Languages like Lua and Python provide a nice ffi libraries ([ffi](http://luajit.org/ext_ffi.html) and [cffi](https://cffi.readthedocs.io/en/latest/overview.html#simple-example-abi-level-in-line), resp), that allow users to specify the *text* of a header file, rather than a filename or even type signature. In our case, we should be able to *create* the proper header from `name`, `argTypes` and `returnType` if a header file is not provided.

3. I assume the `oFile` argument is the name of a shared library to link in? I believe this should a list `dllFiles` of multiple libraries. We may also want a `cppFiles` in case the user wants us to compile their code, so that they can leave this empty if they just want to call some internal R function or a C function that's part of another R package's internal implementation (in which case the package would already have been compiled and linked in), right?

packages/nimble/inst/tests/test-externalCalls.R
```diff
@@ -0,0 +1,98 @@
...
+
+void my_internal_function(double *p, double *ans, int n) {
+  printf("In my_internal_function\\n"); /* cat reduces the double slash to single slash */ 
+  for(int i = 0; i < n; i++) 
+    ans[i] = p[i] + 1.5;
+}
+')
+sink()
+
+## A major limitation is that we cannot have compiled code return anything except a scalar.  So returned non-scalars will have to be via arguments.  That meansto use this in BUGS code we'll have to wrap it in another nimbleFunction.  It will make sense below.
```
fritzo: Sounds like an opportunity for a convenience wrapper or an optional argument `returnArg` or somethting :grinning:

NLMichaud:
Thanks Perry!  I haven't looked over the code outside of the testing file yet, but here's a quick question on the way users will interact with the `nimbleExternalCall` function:

Would it be possible to avoid the creation of a return wrapper function (such as `wrappedRadd1p5` in the `test-externalCalls.R` file) by the addition of another argument to `nimbleExternalCall`?  This argument could be called `returnArg` (or something else).  If `returnArg` was left unspecified, the external call (here `Radd1p5`) would return exactly what the external function returns, which must be void or scalar.  If a user set `returnArg = ans`, for example, then the code generated by `nimbleExternalCall` could return the value of `ans` after the external function has been called (the type of which we would already know from the `fun` argument).  If this were possible, does it seem as though it would make things any simpler for a user?

packages/nimble/inst/tests/test-externalCalls.R
NLMichaud: Oops this was my thought exactly below!

perrydv:
Thanks for all the comments @NLMichaud and @fritzo.  It's definitely rough and these suggestions are helpful.  I'll have to pick it up in some weeks. A few quick responses:

Yes there is an [Rffi](http://www.omegahat.net/Rffi/) that we could use.  I was drafting the quickest possible approach but we could discuss this enhancement.  But maybe your suggestion is simpler than that, just that we generate the header text if necessary.  That seems feasible.

The design of passing in a `function` object itself was inspired by things like `match.call`, but it would be easy enough to provide alternative ways to pass the information.

The other modes of using external code (having nimble compile it, or indicating it is already compiled into a shared library) make sense.

The difficulty of non-scalar return objects is that it would be unusual for external code to return a `NimArr<>`.  It would seem dangerous for it to return a pointer to memory it has allocated that would likely be leaked.  Current the mapping between nimble and C types for argument passing is basically from `NimArr<double>` to `double *`.  We could expand that to go from `NimArr<double>` to `std::vector<double>`, although that would require forcing a copy.  But it would make more sense for return objects: we could copy a returned `std::vector` to a `NimArr<>`.  But now I think I see that your suggestion is actually simpler than this, just that a user can indicate which input argument to the C function should be automatically wrapped to behave as the return value.  That seems feasible, but it may also be necessary to indicate what the expected size(s) of the object will be.

perrydv:
I have partway revised this PR but have some more to do.  It is turning out some suggestions are challenging in the immediate short-term.

It is not so simple to automatically create the wrapper that treats an input argument as the return object, because one must correctly allocate it, such as by a call to `numeric`, `matrix`, or `array`.  I don't have the time to automate that before 0.6-6 and am not sure it is worth it.  So I suggest we leave it to the user to do so and provide a good example in the Rd and the user manual.

For the suggestion that we handle different ways to link to or compile the external code, this runs into limitations in triggering compilation dependencies in the nimble compiler.  We need to decide whether it is worth releasing a version that is really limited in this regard (to using a `.o`) or wait until we put together something more thorough.

The updated branch does allow the `prototype` (formerly `fun`) to be provided as a list of `nimbleType` objects (similarly to for `nimbleList`).

perrydv:
@NLMichaud @fritzo @paciorek This PR has been updated.  See description.

packages/nimble/R/cppDefs_cppProject.R
```diff
@@ -143,7 +143,7 @@ cppProjectClass <- setRefClass('cppProjectClass',
                                        selfCPP <- if(is.character(con)) paste0('"', con, '.cpp"') else '"[FILENAME].cpp"'
                                        CPPincludes <- CPPincludes[ CPPincludes != selfCPP ]
 
-                                       ## Eigen must be included before any R header files because they both define "length"
+                                       ## Eigen must be included before any R header files because they both define "length" (this may be moot, 7/17)
```
fritzo: As of #325, this should no longer be an issue. Consider changing this comment to
```r
## TODO Simplify Eigen include logic now that Nimble defines `R_NO_REMAP`.
```

packages/nimble/R/cppDefs_cppProject.R
```diff
@@ -232,15 +229,16 @@ cppProjectClass <- setRefClass('cppProjectClass',
...
-                                       cppPermList <- c('RcppUtils.cpp',
-                                                        'Utils.cpp',
-                                                        'NamedObjects.cpp',
-                                                        'ModelClassUtils.cpp',
-                                                        'accessorClasses.cpp',
-                                                        'predefinedNimbleLists.cpp',
-                                                        'nimOptim.cpp'
-                                                        )
-                                       if(getNimbleOption('includeCPPdists')) cppPermList <- c(cppPermList, 'dists.cpp', 'nimDists.cpp')
+                                       ## cppPermList appears to be deprecated.  Will remove this commented-out block if no problems arise.
```
fritzo: nit: Remove commented-out code. Here are [some tips](https://github.com/nimble-dev/nimble/blob/devel/CONTRIBUTING.md#avoid-submitting-commented-out-code) on using `git grep`.

packages/nimble/R/cppDefs_cppProject.R
```diff
@@ -268,17 +266,18 @@ cppProjectClass <- setRefClass('cppProjectClass',
                                        if(is.null(nimbleUserNamespace$sessionSpecificDll)) compileDynamicRegistrations(showCompilerOutput = showCompilerOutput)
 
                                        origSHLIBcmd <- SHLIBcmd
-                                       if(isTRUE(nimbleOptions()$stopCompilationBeforeLinking)) {## used only for testing, when we want to go quickly and skip linking and bail out
+                                       if(isTRUE(nimbleOptions('stopCompilationBeforeLinking'))) {## used only for testing, when we want to go quickly and skip linking and bail out
```
fritzo: Is there an intended difference between `getNimbleOption('myOption')` and `nimbleOptions('myOption')`? They are both used in this file.

packages/nimble/R/externalCalls.R
```diff
@@ -0,0 +1,216 @@
+#' Create a nimbleFunction that wraps a call to external compiled code
```
fritzo: Consider marking this experimental like
```r
#' EXPERIMENTAL Create a nimbleFunction that wraps a call to external compiled code.
#' 
...
```

fritzo:
@perrydv Could you explain why your proposed solution better than integrating with Rcpp? From what I can tell, `nimbleExternalCall` is a reinvention of `Rcpp::cppFunction` plus some plumbing to integrate with our compiler. I'm inclined to offload half of that work to Rcpp/RcppEigen so that we can focus more effort on compiler integration. Note that Dirk already suggested #126 that we use RcppEigen's version of Eigen.

Also, this seems like a big enough feature to warrant a design doc.

packages/nimble/R/genCpp_sizeProcessing.R
```diff
@@ -1,5 +1,6 @@
-assignmentAsFirstArgFuns <- c('nimArr_rmnorm_chol', 'nimArr_rmvt_chol', 'nimArr_rwish_chol', 'nimArr_rinvwish_chol', 'nimArr_rcar_normal', 'nimArr_rmulti', 'nimArr_rdirch', 'getValues', 'getValuesIndexRange', 'initialize', 'setWhich', 'setRepVectorTimes', 'assignVectorToNimArr', 'dimNimArr', 'assignNimArrToNimArr')
-setSizeNotNeededOperators <- c('setWhich', 'setRepVectorTimes')
+assignmentAsFirstArgFuns <- c('nimArr_rmnorm_chol', 'nimArr_rmvt_chol', 'nimArr_rwish_chol', 'nimArr_rinvwish_chol', 'nimArr_rcar_normal', 'nimArr_rmulti', 'nimArr_rdirch', 'getValues', 'getValuesIndexRange',  'initialize', 'setWhich', 'setRepVectorTimes', 'assignVectorToNimArr', 'dimNimArr', 'assignNimArrToNimArr')
```
fritzo: nit: Split this long line into a tall line, so that it is easier to read diffs.
```r
assignmentAsFirstArgFuns <- c(
    'nimArr_rnorm_chol',
    'nimArr_rvmt_chol',
    ...)
```

packages/nimble/R/genCpp_sizeProcessing.R
```diff
@@ -207,6 +214,22 @@ sizeProxyForDebugging <- function(code, symTab, typeEnv) {
     return(ans)
 }
 
+## This is used by nimbleExternalCall
```
fritzo: nit: Punctuate sentences in comments. As yourself "Is this how I would write in a preprint"?

packages/nimble/inst/include/nimble/NimArr.h
```diff
@@ -8,6 +8,26 @@
...
+  } else {
+    return(orig.getPtr());
+  }
+}
+
+template<int nDim, class T>
+  void nimArrPtr_copyBackIfNeeded(T* tptr, NimArr<nDim, T> &orig, NimArr<nDim, T> &possibleCopy) {
+  if(tptr == orig.getPtr()) return;
+  if(tptr != possibleCopy.getPtr()) {
+    PRINTF("Problem in unconverting from an external call\n");
```
fritzo: Should this be `NIMERROR` instead of `PRINTF`?

packages/nimble/R/cppDefs_cppProject.R
perrydv: I left it in this one PR while I wanted to be sure it passed testing. Was planning to remove it!

perrydv:
@fritzo In response to your question about `Rcpp::cppFunction`: The "hard" part about `nimbleExternalCall` is generating code to map from `NimArr<>` objects to C/C++ objects and back again, so that the external function can be called from a compiled `nimbleFunction`.  That seems totally distinct from `Rcpp::cppFunction`, which is a handy way to manage C++ code directly from R.  It is easily conceivable that we could make some good use of `Rcpp`, but for this particular purpose I don't see redundancy.  When I'm back in person let's talk about some of these ideas.

Making a design doc for this feature is a good idea.  For context, we have not viewed this as our highest priority, so let's be careful about elevating it.  It was an experimental feature of interest to some users.  Does it seem worth including in experimental status in 0.6-6 or does it make more sense to wait?

Thanks for the code review!  I'll go over those items soon.

fritzo:
@perrydv Thanks for the clarification re: `Rcpp::cppFunction`, that makes sense. I wish I could give a less superficial code review, but this feature is quite tricky! I'd be happy to chat in person when you get back.

packages/nimble/R/cppDefs_cppProject.R
perrydv: Done - the comments even indicated I was planning to take it out :)

packages/nimble/R/cppDefs_cppProject.R
perrydv: Hmmm, not sure.  It looks like `getNimbleOption` will be more efficient and will error if `'myOption'` doesn't exist.  `nimbleOptions` can get or set.  It will be slower and it will return `NULL` if the option doesn't exist.  If we want to revamp this system let's do it via a different Issue & PR.  Am going to ignore for this unrelated PR.

perrydv:
Thanks for the review @fritzo!  I've modified following most suggestions.

PR #455, opened 2017-06-27
Move test requirements from .travis.yml to install_requirements.R

fritzo:
This fixes issues with `run_tests.R --parallel` and enables parallel testing by default.
The two issues were:

1. Missing dependencies.
  This is fixed by moving requirements into a script `install_requirements.R` which is used by both `.travis.yml` and `make test`.
2. Conflict between GNU `make` and R's `parallel::mclapply`.
  This is fixed by setting `MAKEFLAGS=-j1` in child processes that run tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,12 +39,7 @@ addons:
       - time  # For /usr/bin/time profiling.
 
 install:
-  - Rscript -e "install.packages('igraph', repos = 'http://cran.us.r-project.org')"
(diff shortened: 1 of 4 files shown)


fritzo:
@paciorek `make test` now works on my system, and completes in under 22 minutes!
```sh
$ time make test
cd .. ; ./install_requirements.R
cd .. ; ./run_tests.R --parallel
...
PASSED all tests

real	21m7.525s
user	48m12.332s
sys	1m29.872s
```

PR #461, opened 2017-07-06
Add xfail arguments to math tests

fritzo:
This adds support for an optional `xfail` field of the parametrized tests for math, and avoids reliance on commenting-out-code in tests.

## Why?

We have a nice bank of math examples in [`mathTestLists.R`](https://github.com/nimble-dev/nimble/blob/devel/packages/nimble/inst/tests/mathTestLists.R) that are currently used in `test-math.R`, and we're starting to use those examples in other branches. A few those fail, and our current practice is to comment-out the failing tests. This is problematic when different examples fail on different branches or files: we would currently have to comment out an example if it fails in any branch or file.

## How?

This PR adds an `xfail` field to the example lists (the name is borrowed from py.test, dejagnu, and php). This field specifies a regular expression of circumstances under which the test can fail, using the convention that `test_math()` will now take a second argument as the context name (currently only 'math'), and can fail in three circumstances:

- general compile error" `xfail = 'math.*runs'`,
- disagreement between R and the Nimble DSL: `xfail = 'math.*DSL'`
- disagreement between R and compiled code: `xfail = 'math.*Cpp`

Regular expressions seem like the simplest way to specify the sparse tensor of expected failures (contexts x expectations). For example we can in the future specify `xfail = '(math|tensorflow).*Cpp'`.

### DIFF
--- a/packages/nimble/inst/tests/test-math.R
+++ b/packages/nimble/inst/tests/test-math.R
@@ -9,16 +9,16 @@ context("Testing of math functions in NIMBLE code")
 source(system.file(file.path('tests', 'mathTestLists.R'), package = 'nimble'))
 
 set.seed(0)
-ans1 <- sapply(testsVaried, test_math)    ## 12
-ans2 <- sapply(testsBasicMath, test_math) ## 70
+ans1 <- sapply(testsVaried, test_math, 'math')    ## 12
+ans2 <- sapply(testsBasicMath, test_math, 'math') ## 70
 if(.Platform$OS.type == 'windows') {
     message("Since you are running on Windows, tests stopped prior to reaching max DLL limit.  Please use test-math2 to continue")
     stop()
 }
-ans3 <- sapply(testsMoreMath, test_math)  ## 41
-ans4 <- sapply(testsReduction, test_math) ## 13
-ans5 <- sapply(testsComparison, test_math)## 6
-ans6 <- sapply(testsMatrix, test_math)    ## 19
+ans3 <- sapply(testsMoreMath, test_math, 'math')  ## 41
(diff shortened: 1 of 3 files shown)


packages/nimble/inst/tests/test-math.R
```diff
@@ -9,16 +9,16 @@ context("Testing of math functions in NIMBLE code")
 source(system.file(file.path('tests', 'mathTestLists.R'), package = 'nimble'))
 
 set.seed(0)
```
fritzo: In other branches, I will add e.g. the argument 'tensorflow' or 'eigenize-v2'

packages/nimble/inst/tests/test-math.R
```diff
@@ -9,16 +9,16 @@ context("Testing of math functions in NIMBLE code")
...
-ans1 <- sapply(testsVaried, test_math)    ## 12
-ans2 <- sapply(testsBasicMath, test_math) ## 70
+ans1 <- sapply(testsVaried, test_math, 'math')    ## 12
```
fritzo: In other branches, I will add the argument 'tensorflow' or 'eigenize-v2'

packages/nimble/inst/tests/test_utils.R
```diff
@@ -292,16 +292,16 @@ test_coreRfeature_internal <- function(input, verbose = TRUE, dirName = NULL) {
   invisible(NULL)
 }
 
-gen_runFun <- function(input) {
+gen_runFun <- function(param) {
```
fritzo: I've renamed the argument from generic `input` to specific `param` to match standard terminology of "parametrized tests".

perrydv:
LGTM

fritzo:
Thanks for reviewing, Perry!

PR #464, opened 2017-07-06
Fixes #77 by updating to Eigen 3.3.4

### LINKED ISSUE #77: update version of Eigen we provide with NIMBLE

fritzo:
The only behavior change is:

- Dropped support for determinants of integer matrices


perrydv:
Have you confirmed that removing the `::internal` layer is done correctly?  Specifically, it seems that previously `Index` was a type defined by template as a trait of arbitrary Eigen types, and now it is simply a type in the Eigen namespace.  Did Eigen consolidate to have a single `Index` type throughout?

fritzo:
Yes, Eigen did consolidate to have a single index type throughout. That type is now `ptrdiff_t`.

PR #467, opened 2017-07-07
Remove EXPERIMENTAL tag from optim documentation.

fritzo:
Note that I've set the merge target to the `make-man` branch to make reviewing tractable. We hadn't run `make man` in a long time, so the diff wrt `devel` is huge.


perrydv:
LGTM

PR #471, opened 2017-07-08
MCMC sampler control defaults are internal to sampler algorithms

danielturek:
DO NOT MERGE.

Reasonably large change in MCMC system.

All developers are encouraged to test their MCMCs & custom sampler functions.


perrydv:
@danielturek LGTM.  We are moving away from leaving commented-out code as a record of changes and relying on git instead if we need to refer back to old versions.

Issue #15, opened 2015-05-01
allow user-defined link functions


paciorek:

perrydv:
moved to wiki

PR #475, opened 2017-07-11
Migrate User Manual from nimble-docs to nimble

### LINKED ISSUE #472: Move User Manual into nimble repo?
@EmmanuelCharpentier suggested moving the User Manual into the main Nimble repo. @fritzo agrees this would be helpful. What are everyone's thoughts?

The `#1` finding of GitHub's [2017 Open Source survey](http://opensourcesurvey.org/2017/) was that

> Documentation is highly valued, frequently overlooked, and a means for establishing inclusive and accessible communities.

## Pros

- It's generally easier to keep documentation in sync with code if they live in the same repo.
  - PRs modifying a function's behavior can update both code and documentation.
  - New features can be released with immediately updated documentation.
  - Branches can contain edits to both code and docs, thus encouraging developers to write documentation as they write code (and before they've forgotten the code!).
- It's easier for users to find the manual, in case they want to fix a bug or add documentation.
- It may be easier to html-ize the manual #468 if the `man/*.Rd` function docs and the tutorial-like UserManual live in a single repo. E.g. [pkgdown](http://hadley.github.io/pkgdown) may be able to build a site that incorporates both types of documentation.
- ???

## Cons

- It will take effort to make the move.
- ???

fritzo:
This fixes #472 by migrating the Nimble User Manual from https://github.com/nimble-dev/nimble-docs to this repo. See also [nimble-docs #15](https://github.com/nimble-dev/nimble-docs/pull/15).

I've also added a documentation section to [`CONTRIBUTING.md`](CONTRIBUTING.md). After recent fixes to our `UserManual/Makefile` and `packages/Makefile`, it should be easier for developers to regenerate documentation in their branches, so as to keep documentation as close as possible to the code.

**Note to reviewers:** This is a simple cross-repo `copy` operation. There is no need to review the copied code, but I would like a :+1: if you think this is a reasonable action. Thanks!

Thanks to @EmmanuelCharpentier for the original suggestion.


paciorek:
LGTM.

fritzo:
Thanks, Chris. Do you think Perry is OK with this change?

paciorek:
@perrydv can you give thumbs up on this?

PR #476, opened 2017-07-11
Dynamic indexing fix typos

fritzo:
If tests were quicker, then developers would be more likely to run tests before committing.


paciorek:
I had fixed these typos (except for the optional inclusion of {} for a one-line body in an if statement already but hadn't pushed.  So please pull my changes and go from there. I think this PR can be deleted.

PR #479, opened 2017-07-12
Fixes #478 by documenting run.time in the UserManual

### LINKED ISSUE #478: run.time() is not documented
@perrydv filed this issues on `nimble-docs`

fritzo:


perrydv:
This looks fine and thanks @fritzo for cleaning up my lapse on this.  I would have made fewer assumptions about different timing needs and when they might arise, and instead simply have presented how `run.time` works and let different programmers determine their needs for it, e.g. timing different steps of a function.  E.g. we've used it to collect times of each sampler in an MCMC.

PR #483, opened 2017-07-12
Use Eigen::internal::aligned_malloc for NimArr memory

fritzo:
This replaces `NimArr`'s memory functions `new[]`/`delete[]` with compile-time conditional versions that can optionally use Eigen's internal aligned versions `aligned_malloc`/`aligned_free`. While this PR makes no changes to compiled code, it does allow us to experiment with different allocators by switching a single byte of source code.

## Why?

Speed. Aligned memory gives CPUs more opportunity to vectorize operations using SIMD operations (mmx, sse, avx, fma, ...). Eigen can make use of aligned memory internally. Alignment mainly affects medium sized matrix and vector operations (4 to 100 elements). Below 4 there is little room to parallelize, and above 100, compilers can deal separately with the unaligned ends and the aligned middle. 

## How?

This first PR simply replaces NimArr's alignment methods so that the memory aligned. This PR does not take the next step of informing `Eigen::Map<>`s about whether their memory is aligned (see [documentation](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html)). We should still see some benefit, since Eigen does check for aligned memory.

## Tested

- [x] Runs on Linux.
- [x] Runs on OS X.
- [x] Runs on Windows.

## Benchmarks
There doesn't seem to be a significant change on my linux system:
```
                       Before this PR | After this PR
--------------------------------------+--------------------------
Benchmarking Matrix Arithmetic        |
  K   M   N DSL calls/ms C++ calls/ms | DSL calls/ms C++ calls/ms
  1   1   1       1077.6      36224.5 |       1079.9      35894.9
  2   2   2        902.5      20764.5 |        900.9      21215.1
  3   3   3        905.0      13886.2 |        892.1      13212.6
  4   4   4        858.1      11064.0 |        762.2      10422.5
  5   5   5        684.5       8021.2 |        629.2       7711.7
  8   8   8        529.4       5298.8 |        524.9       5227.4
 10  10  10        456.4       4146.6 |        452.1       4038.3
 16  16  16        258.3       2365.8 |        210.9       2336.1
 32  32  32         86.0        855.0 |         60.8        838.0
 64  64  64          9.4        245.2 |         12.1        222.3
100 100 100         10.9         97.4 |          9.9         93.0
128 128 128          2.1         48.7 |          2.1         49.4
256 256 256          0.5         13.3 |          0.5         12.5
--------------------------------------+--------------------------
Benchmarking Matrix Multiplication    |                       
  K   M   N DSL calls/ms C++ calls/ms | DSL calls/ms C++ calls/ms
  1   1   1       4048.6      20185.5 |       3906.2      19431.5
  2   2   2       3086.4      15775.8 |       3030.3      15168.0
  3   3   3       2645.5      10733.1 |       2577.3      10366.9
  4   4   4       2611.0       6678.2 |       2538.1       6428.1
  5   5   5       1886.8       4223.7 |       1824.8       4096.3
  8   8   8       1042.8       2206.6 |       1044.4       2200.1
 10  10  10       1004.2       1495.3 |        572.9       1454.4
 16  16  16        407.9        708.1 |        404.0        680.8
 32  32  32         75.0        139.3 |         72.3        130.3
 64  64  64          7.2         21.0 |          7.2         20.1
100 100 100          3.8          5.9 |          3.5          5.7
128 128 128          1.0          2.9 |          0.9          2.7
256 256 256          0.3          0.4 |          0.3          0.4
--------------------------------------+--------------------------
Benchmarking Special Functions        |
          N DSL calls/ms C++ calls/ms | DSL calls/ms C++ calls/ms
          1        971.8       4688.6 |        914.9       4479.7
          2        503.8       2376.9 |        455.9       2358.9
          3        429.4       1671.7 |        413.1       1560.0
          4        387.4       1220.8 |        368.7       1215.0
          5        371.5       1001.7 |        352.5        992.5
          8        285.1        638.5 |        279.3        628.9
         10        249.7        507.2 |        234.3        496.7
         16        192.9        297.1 |        182.0        319.4
         32        101.8        150.1 |        104.2        158.1
         64         55.4         72.6 |         58.9         78.1
        100         37.5         43.2 |         39.5         49.5
        128         29.5         36.1 |         31.2         38.3
        256         15.4         18.1 |         16.2         18.9
        512          7.7          8.9 |          7.9          9.3
       1000          3.9          4.6 |          3.9          4.6
       1024          3.8          4.2 |          3.8          4.5
       2048          1.8          2.1 |          1.8          2.2
       4096          0.9          1.0 |          0.9          1.0
       8192          0.3          0.4 |          0.5          0.6
      10000          0.3          0.3 |          0.4          0.5
--------------------------------------+--------------------------
```


perrydv:
LGTM

fritzo:
Thanks, Perry! I'll wait to merge until a windows user (@NLMichaud ?) has verified that this works for them.

NLMichaud:
@fritzo This PR installs successfully and tests run successfully on windows.

fritzo:
Thanks for testing, @NLMichaud !

Before I merge this PR, I'd like to add some standard benchmarks to Nimble, say next to `inst/test` as `inst/benchmark`. It's good practice to A/B test benchmark results when merging PRs that affect performance, and I'd like to volunteer this PR as the first to go through this more rigorous process.

paciorek:
sounds like a good plan to me -- do we know if other R packages use the
inst/benchmark location?

fritzo:
@paciorek Great question! There is precedent for an `inst/benchmarks` directory. I see on my local mirror of 533 CRAN packages:
```sh
$ ag -l microbenchmark \
    | sed 's/^[^/]*\///' | sed 's/inst\//inst-/' | sed 's/\/.*//' | sed 's/-/\//g' \
    | sort | uniq -c | sort -nr
     54 inst/doc
     51 man
     50 R
     46 src
     30 DESCRIPTION
     28 vignettes
     26 tests
     10 README.md
      7 NEWS.md
      5 inst/benchmarks     # <--------------- This seems most appropriate to me.
      3 NAMESPACE
      3 demo
      2 NEWS
      2 MD5
      2 inst/tests
      2 inst/examples
      2 ChangeLog
      1 tools
      1 inst/test.R
      1 inst/README.brew
      1 inst/docker
```

fritzo:
Hmm, it looks like package `h5` uses `inst/benchmark`, `igraph` uses `inst/benchmarks`. A few packages use `R/benchmark.R`, but that feels more like benchmarking tools and less on par with tests.

perrydv:
Yes, I also like the idea of integrating benchmarks into the package structure and am happy for you to do so following a combination of common practices and what suits our needs as you see fit.

perrydv:
👍 The updated PR with new benchmarks looks great.  Improvements aren't huge but it seems clearly a good move and we have yet to modify construction of `Eigen::Map<>`s to harness the aligned allocation.

perrydv:
Oh, now that I've looked at the benchmarks I realized I was making an inverted interpretation without thinking carefully.  So there is if anything a minor regression, perhaps due to the overhead of the actual allocation steps?  It still seems like a good move on basic principles.

perrydv:
Another comment for future consideration: We may want at a later stage to avoid a copy between R and compiled nimble by making Eigen maps directly into R-allocated memory.  For that purpose we will need to know R's allocation behavior and when we modify how `Eigen::Map` code is generated we'll want to maintain options for both aligned and unaligned rather than assuming we always will have aligned memory in every future use case.

fritzo:
@perrydv Since there's no conclusive evidence that aligned memory helps, I suggest we not switch yet. I've changed this PR to conditionally use either `new`/`deletete` or `Eigen::internal::aligned_*`, defaulting to our current behavior. This will be a no-op to our compiled code, but will make it easier to test subsequent PRs that deal with `Eigen::Map<>` alignment hinting. Sound reasonable?

packages/nimble/inst/include/nimble/NimArrBase.h
```diff
@@ -28,6 +29,33 @@
 
 using std::vector;
 
+// This allows us to experiment with different allocators. In theory aligned
+// memory should improve Eigen performance, but microbenchmarks have failed to
+// substantiate this.
+#if !defined(NIMBLE_ALIGNED_MALLOC)
+#define NIMBLE_ALIGNED_MALLOC 0
```
fritzo: We can change this one line to experiment with different allocators.

perrydv:
👍

PR #484, opened 2017-07-13
Add tests/test-benchmarks.R for simple math functions

fritzo:
This adds a `test-benchmarks.R` to our test suite. The test is run automatically. You can run it manually using `make benchmark`.

## Question for reviewers

Are there any other benchmarks you would like me to add?

## Benchmark Results
<details><pre><code>
---------------------------------------
Benchmarking Matrix Arithmetic
  K   M   N DSL calls/ms C++ calls/ms
  1   1   1       1077.6      36114.9
  2   2   2        914.1      21279.5
  3   3   3        910.7      13732.0
  4   4   4        858.1      10849.8
  5   5   5        699.0       7743.5
  8   8   8        323.8       5439.1
 10  10  10        464.3       4135.2
 16  16  16        323.9       2373.8
 32  32  32         87.6        861.3
 64  64  64          8.1        219.2
100 100 100         10.6        100.5
128 128 128          2.2         52.2
256 256 256          0.5         13.3
---------------------------------------
---------------------------------------
Benchmarking Matrix Multiplication
  K   M   N DSL calls/ms C++ calls/ms
  1   1   1       4056.4      20180.5
  2   2   2       3105.6      15315.2
  3   3   3       2652.5      10724.0
  4   4   4       2527.0       6685.9
  5   5   5       1876.2       4250.0
  8   8   8        481.3       2198.7
 10  10  10       1063.8       1512.8
 16  16  16        422.8        706.1
 32  32  32         75.0        136.1
 64  64  64          7.3         21.0
100 100 100          3.8          5.9
128 128 128          1.0          2.9
256 256 256          0.4          0.4
---------------------------------------
-------------------------------
Benchmarking Special Functions
    N DSL calls/ms C++ calls/ms
    1        529.9       4429.4
    2        486.9       2400.4
    3        446.0       1603.0
    4        407.2       1222.6
    5        366.8        982.5
    8        304.2        624.6
   10        247.5        506.8
   16        188.9        318.4
   32        103.4        158.3
   64         57.1         77.8
  100         39.3         48.9
  128         32.2         38.2
  256         16.6         18.6
  512          8.1          9.2
 1000          2.0          4.6
 1024          2.6          3.2
 2048          1.3          1.6
 4096          0.7          0.8
 8192          0.3          0.4
10000          0.3          0.3
-------------------------------
</code></pre></details>

### DIFF
--- a/packages/Makefile
+++ b/packages/Makefile
@@ -31,6 +31,9 @@ test: FORCE
 	cd .. ; ./install_requirements.R
 	cd .. ; ./run_tests.R --parallel
 
+benchmark: FORCE
+	cd .. ; ./run_tests.R benchmarks
+
 # We are progressively delegating style to clang-format.
 # Some older C++ files are still styled manually.
 clang-format: FORCE
--- a/packages/nimble/inst/tests/test-benchmarks.R
+++ b/packages/nimble/inst/tests/test-benchmarks.R
@@ -1,0 +1,105 @@
+## This runs benchmarks of compiled NIMBLE code. It is located in tests/ so
+## that it can continue to be tested, but it most useful to run this manually.
+
+source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
+
+context('Benchmarking Nimble code')
+cat('\n')
+
+## Computes number of iterations per second.
+## This increase iterations until test takes around 1 sec.
+benchmarkIters <- function(fun, minIters = 100L, minRunTimeSec = 0.1) {
+    iters <- minIters
+    elapsed <- fun(iters)
+    while (elapsed < minRunTimeSec) {
+        iters <- as.integer(iters * max(2, (minRunTimeSec / (elapsed + 1e-6)) ^ 0.8))
+        elapsed <- fun(iters)
+    }
+    return(iters / elapsed)
+}
+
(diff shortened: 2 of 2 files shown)


perrydv:
The basic design looks good to me.  Note that R and C/C++ can perform (very) differently for basic iteration.  Given `for(i in 1:bigNumber)`, uncompiled execution involves construction of the large object returned by `1:bigNumber` and then iteration over its values ignoring the simplicity of their pattern.  Compiled execution makes a C `for` loop that increments `i` one by one.  So if the goal is truly to isolate the arithmetic operations, we may want to accumulate `run.time` results from within the `for` loop, rather than including the `for` loop inside the `run.time`.

Future benchmarks of interest could include:

- copying and math with slices and arbitrary integer or logical indices
- `nimCopy` (see test-copy), known to be inefficient.
- model computations via direct calls to nodeFunctions and via `calculate`.
- end-to-end MCMCs or other algorithms.

But in the interest of moving forward I suggest we merge these initial tests and add to them when we work on relevant parts of the system.

fritzo:
Thanks, Perry! That's a good point about overhead of the for loop. I was also worried about overhead of `run.time` itself, so maybe we can experiment to see which is more accurate.

I'll merge this as-is and start adding additional benchmarks.

PR #492, opened 2017-07-18
Fixes #488: un-reference scalar arguments to rankSample

### LINKED ISSUE #488: rankSample is tricky to use
Not sure if I'm using it wrong, but I had to do some gymnastics to use rankSample in the new categorical sampler I just wrote for categorical nodes.  Advice or discussion on modifications to rankSample() welcome:
```
    run = function() {
        currentValue <- model[[target]]
        for(i in 1:maxval) {
            model[[target]] <<- i
            probs[i] <<- exp(calculate(model, calcNodes))
        }
     ## can't use model[[target]] directly in rankSample as it expects int NimArr for 3rd arg.
        newValue <- integer(1)
   ## for second arg, rank sample wants an int&; I need these gymnastics instead of simply doing rankSample(probs, 1, newValue)
        size <- integer(1,1)
        rankSample(probs, size[1], newValue)
        model[[target]] <<- newValue[1]  
        calculate(model, calcNodes)
```

perrydv:
There were three cases of scalar arguments unnecessarily passed by reference in the two versions of `rankSample`.  Until recently,  `rankSample` had not been used outside of its original need, so this oversight in style had gone unnoticed.  This PR removes the reference `&` from those arguments.

### DIFF
--- a/packages/nimble/inst/CppCode/RcppNimbleUtils.cpp
+++ b/packages/nimble/inst/CppCode/RcppNimbleUtils.cpp
@@ -764,12 +764,12 @@ double nimMod(double a, double b) {
 //   return(a.value < b.value);
 // }
 
-void rankSample(NimArr<1, double> &weights, int &n, NimArr<1, int> &output) {
+void rankSample(NimArr<1, double> &weights, int n, NimArr<1, int> &output) {
   bool silent = false;
   rankSample(weights, n, output, silent);
 }
 
-void rankSample(NimArr<1, double> &weights, int &n, NimArr<1, int> &output, bool& silent) {
+void rankSample(NimArr<1, double> &weights, int n, NimArr<1, int> &output, bool silent) {
   //PRINTF("in VOID rankSample\n");
   output.setSize(n);
   int N = weights.size();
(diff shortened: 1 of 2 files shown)


danielturek:
Once this is merged, we can simplify the new categorical sampler implementation.  I'm happy to do that, once it's merged into devel, then devel is merged into branch dynamic-indexing-modelDef

fritzo:
:+1:

perrydv:
OK @danielturek go ahead and simplify the new categorical sampler's use of `rankSample`.

danielturek:
I have merged these changes into `dynamic-indexing-modelDef` branch, and simplified the new `categorical` sampler accordingly.

@paciorek these changes are pushed on branch `dynamic-indexing-modelDef`.  I'm not sure the status there otherwise, but as far as with the new sampler, I'd say it's ready to merge.

paciorek:
Thanks all. Though I did just realize (disturbingly belatedly) that we
could just use rcat in the categorical sampler. I think we probably want to
do that for code readability/style.

DT: thanks for all the various cleanup to sampler_categorical - all
appreciated.

danielturek:
No problem.  Makes sense about `rcat()`, I didn't think of it either.  I would have re-invented the wheel and written a while-loop....

Do you want to change to using `rcat()`, or want me to?

paciorek:
if you're up for it and have a few moments, please do change to rcat.

danielturek:
@paciorek I modified `sampler_categorical` to use `rcat()`, and tested it for correctness.

Changes pushed to branch `dynamic-indexing-modelDef`.

PR #506, opened 2017-07-21
Improve usability of run_tests.R

fritzo:
1. Add a --stop option to stop tests after the first failure.
2. Support CTRL+C interrupt via `sys::exec_wait` if the `sys` package is installed.


paciorek:
LGTM presuming Travis errors are resolved.

PR #510, opened 2017-07-24
Revert breaking change to test-mcmc.R

fritzo:


fritzo:
@paciorek This reverts the small breaking change you had made to test-mcmc.R in 46a92fc1b0f7

paciorek:
Ok, so I just ran the jaw-linear stanza in test-mcmc.R where the stanza _did_ include the "breaking" changes I had made, and the test completed without error (I'm guessing/hoping that is what I had done before and therefore why I did not expect a travis problem). 

Any of the Travis experts out there have any suggestion as to why the test is fine on my Linux box but Travis is unhappy?

Issue #517, opened 2017-07-26
warning because of "long long" on Ubuntu


paciorek:
We're getting this warning under Rhub on Ubuntu.
```
  ../inst/include/Eigen/src/Core/arch/CUDA/Half.h:122:46: warning: ISO C++ 1998 does not support ‘long long’ [-Wlong-long]
```
See
https://builder.r-hub.io/status/original/nimble_0.6-6.tar.gz-3f66989dbba748058f2b6fe95b2fa349

perrydv:
It is just a warning.  The configure output above the warnings contains:

"checking whether the C++ compiler supports the long long type... yes"

Not sure if that is ok for CRAN or not.

PR #518, opened 2017-07-27
Substitute u?int64_t for (unsigned)?long long in Eigen files

fritzo:
Fixes: #517 

**Status:** DO NOT MERGE - for testing only

Disclaimer: I have no idea what I'm doing.


fritzo:
@paciorek Can we test this attempted fix on R-hub?

paciorek:
Yes, I will do so. And, thanks.

paciorek:
@fritzo That seems to have silenced the long long errors and I don't see based on rhub that there were any additional bad consequences.

PR #520, opened 2017-07-27
revert -g avoidance

paciorek:
I will merge this.


fritzo:
@paciorek FYI in case you hadn't seen it, there is a `git revert` command that sometimes works. In this case we may have been able to simply
```sh
git checkout devel
git checkout -b revert-g-flag
git revert 7f3553fa
```

PR #526, opened 2017-08-02
Fix nim eigen

### LINKED ISSUE #523: Disagreement between R's eigen() and nimEigen() ?
There appears to be disagreement between the numerical results of R's eigen(), and our nimEigen().

Is this a problem?

Reproducible example:

```
M <- 1:4
M1 <- diag(M^-0.5)
M2 <- diag(M^ 0.5)

Cmatrix <- t(array(c(0,1,0,0,  .5,0,.5,0,  0,.5,0,.5,   0,0,1,0), c(4,4)))

x <- M1 %*% Cmatrix %*% M2

eigen(x)

library(nimble)
nimEigen(x)
```

NLMichaud:
Fixes #523 by providing correct eigendecompositions for non-symmetric matrices, as long as those decompositions contain no complex numbers.  Decompositions with complex numbers will have those numbers set to `NaN`.   Additionally adds a `symmetric` argument (default `FALSE`) to `nimEigen` that is used to signify the matrix being decomposed is guaranteed to be symmetric.  

Matching between R's eigen and NIMBLE's eigen for non-symmetric matrices should now happen, with the possible exception of eigendecompositions of non-symmetric matrices which have multiple eigenvalues with the same absolute value.  It is possible that the ordering of these same valued eigenvalues, and their corresponding eigenvectors, could be permuted between R and NIMBLE.


danielturek:
Sounds perfect @NLMichaud.  Thanks for the quick attention.

Offhand, I can't think of any logical way to prevent the re-ordering of eigenvectors (corresponding to duplicated eigenvalues), either.  I guess go with this for now, and cross that bridge if it ever becomes an issue.

danielturek:
@perrydv @NLMichaud What are the chances of this PR #526 getting reviewed and merged before the weekend?

NLMichaud:
@danielturek @perrydv  I believe this is ready to merge, but if anyone wants to check out the revised c++ code for the `EIGEN_EIGEN_INTERNAL` function [here](https://github.com/nimble-dev/nimble/blob/ddacbd1130dc78c0a540df229ac16fbeae6235d9/packages/nimble/inst/include/nimble/EigenTypedefs.h), specifically the logic for sorting the eigenvectors by absolute value in the non-symmetric case to make sure there are no bugs there, that would be great!  Otherwise, this can be merged at will, so anyone should feel free to do so!

packages/nimble/inst/include/nimble/EigenTypedefs.h
```diff
@@ -85,22 +85,69 @@ nimSmartPtr<EIGEN_SVDCLASS_R>   EIGEN_SVD_R(const Eigen::MatrixBase<Derived> &x,
...
 	}
     Eigen::DecompositionOptions eigOpts = valuesOnly ? EigenvaluesOnly : ComputeEigenvectors;
-	SelfAdjointEigenSolver<MatrixXd> solver(x, eigOpts); // The MatrixXd here doesn't seem generic, but I couldn't get it to work otherwise and it would be odd to do an Eigen decomposition on anything else. -Perry
-	Eig_eigVals = solver.eigenvalues().reverse();
-	if(!valuesOnly){
-	  returnClass->getVectors().initialize(0, 0, x.rows(), x.cols());
-	  Map<MatrixXd> Eig_eigVecs(returnClass->getVectors().getPtr(),x.rows(),x.cols());
-	  Eig_eigVecs = solver.eigenvectors().rowwise().reverse();	
+	if(!symmetric){
+		symmetric = true;
```
fritzo: nit: The symmetry check would cleaner as a helper function. E.g. you could `return false` rather than using two break statements. It would improve readability of `EIGEN_EIGEN_INTERNAL()`.

packages/nimble/inst/include/nimble/EigenTypedefs.h
```diff
@@ -85,22 +85,69 @@ nimSmartPtr<EIGEN_SVDCLASS_R>   EIGEN_SVD_R(const Eigen::MatrixBase<Derived> &x,
...
-	//	return(returnClass);
+	else{ 
+		EigenSolver<MatrixXd> solver2(x, eigOpts); 
+		vector<pair<double,int> > sortIndices;
+		for(int i = 0; i<x.rows(); i++){
+			sortIndices.push_back(make_pair(abs(solver2.eigenvalues().real()(i)),i));
+		}
+		sort(sortIndices.begin(),sortIndices.end());
+		for(int i = 0; i < x.rows() ; ++i){
+			if(solver2.eigenvalues().imag()(sortIndices[x.rows() - (i+1)].second) != 0){
```
fritzo: nit: Extract repeated subexpression
```c++
const int j = sortIndices[x.rows() - (i+1)].second;  // Is j the right variable name?
```

packages/nimble/inst/include/nimble/EigenTypedefs.h
```diff
@@ -85,22 +85,69 @@ nimSmartPtr<EIGEN_SVDCLASS_R>   EIGEN_SVD_R(const Eigen::MatrixBase<Derived> &x,
...
+			}
+			else{
+				Eig_eigVals(i) = solver2.eigenvalues().real()(sortIndices[x.rows() - (i+1)].second);
+			}
+		}
+		if(!valuesOnly){
+			Map<MatrixXd> Eig_eigVecs(returnClass->getVectors().getPtr(),x.rows(),x.cols());
+			MatrixXd sorted_eigVecs(x.rows(), x.rows());
+			for(int i = 0; i<x.rows(); i++){
+				sorted_eigVecs.col(i) = solver2.eigenvectors().real().col(sortIndices[x.rows() - (i+1)].second);
```
fritzo: ditto: extract repeated subexpression

UserManual/chapter_DataStructures.Rnw
```diff
@@ -333,7 +333,7 @@ eigenListFunctionGenerator <- nimbleFunction(
   },
   run = function(){
     ## we will take the eigendecomposition and svd of a symmetric matrix
-    eigenAndSvdList$demoEigenList <<- eigen(demoMatrix, only.values = TRUE)
+    eigenAndSvdList$demoEigenList <<- eigen(demoMatrix, symmetric = TRUE, only.values = TRUE)
```
fritzo: Consider moving `symmetric` to the third argument. It is generally safer to add parameters to the end of a parameter list, so as not to break existing user code with positional arguments like `eigen(X, FALSE)`.

packages/nimble/R/distributions_implementations.R
```diff
@@ -148,14 +152,14 @@ nimDerivs <- function(nimFxn = NA, order = nimC(0,1,2)){
...
 #'    },
 #'    run = function(){
-#'      eigenvalues <- eigen(demoMatrix)$values
+#'      eigenvalues <- eigen(demoMatrix, symmetric = TRUE)$values
 #'      returnType(double(1))
 #'      return(eigenvalues)
 #'  })
 #' 
-nimEigen <- function(x, only.values = FALSE) {
+nimEigen <- function(x, symmetric = FALSE, only.values = FALSE) {
```
fritzo: ditto: Consider moving `symmetric` to the third position

packages/nimble/R/distributions_implementations.R
```diff
@@ -148,14 +152,14 @@ nimDerivs <- function(nimFxn = NA, order = nimC(0,1,2)){
...
   ## placeholder list with correct names of elements, will be populated in C++
-  .Call(C_nimEigen, x, as.logical(only.values), eigenNimbleList$new())
+  .Call(C_nimEigen, x, as.logical(symmetric), as.logical(only.values), eigenNimbleList$new())
```
fritzo: Note that this ordering is not as important, since it is an internal function.

packages/nimble/R/nimbleFunction_keywordProcessing.R
```diff
@@ -776,7 +776,7 @@ matchFunctions[['double']] <- function(nDim, dim, default, ...){}
 matchFunctions[['int']] <- function(nDim, dim, default, ...){}
 matchFunctions[['nimOptim']] <- nimOptim
 matchFunctions[['nimOptimDefaultControl']] <- nimOptimDefaultControl
-matchFunctions[['nimEigen']] <- function(squareMat, only.values = FALSE){}
+matchFunctions[['nimEigen']] <- function(squareMat, symmetric = FALSE, only.values = FALSE){}
```
fritzo: ditto: Consider moving `symmetric` to third argument.

packages/nimble/inst/tests/test-nimbleList.R
```diff
@@ -1296,13 +1306,68 @@ expect_equal(RnimbleList$v, CnimbleList$v)
 expect_equal(testInst$testList27$testSvd$d, CtestInst$testList27$testSvd$d)
 expect_equal(testInst$testList27$testSvd$u, CtestInst$testList27$testSvd$u)
 expect_equal(testInst$testList27$testSvd$v, CtestInst$testList27$testSvd$v)
+test_that("return object (from c++) is nimbleList.",
```
fritzo: nit: (and beyond scope of this PR) It is safer to wrap the entire test in a `test_that`, e.g.
```r
test_that('nimbleList Test 27', {
  nlTetsFunc <- nimbleFunction(
    ...
  )
  testInst <- ...
  ...
  expect_equal(diag(1), ...)
  ...
  expect_identical(is.nl(CnimbleList), TRUE, info = 'Return object (from c++) is nimbleList')
})
```
This pattern has the following advantages:
- avoids copy-and-paste errors
- avoids polluting global environment
- is compatible with both `Rscript test-nimbleList.R` and `test_package('nimble', 'nimbleList')`. (The current pattern will stop at the first error in the `Rscript ...` version, and will ignore some failed expectations).
- Correctly attributes errors in setup (e.g. if nlTestFunc27 errors, it will be reported correctly).

packages/nimble/inst/include/nimble/EigenTypedefs.h
```diff
@@ -85,22 +85,69 @@ nimSmartPtr<EIGEN_SVDCLASS_R>   EIGEN_SVD_R(const Eigen::MatrixBase<Derived> &x,
...
+		}
 	}
-	//	return(returnClass);
+	else{ 
+		EigenSolver<MatrixXd> solver2(x, eigOpts); 
+		vector<pair<double,int> > sortIndices;
+		for(int i = 0; i<x.rows(); i++){
+			sortIndices.push_back(make_pair(abs(solver2.eigenvalues().real()(i)),i));
+		}
+		sort(sortIndices.begin(),sortIndices.end());
```
fritzo: nit: The logic below would be a little simpler if you reverse-sorted; then you could use `i` instead of `x.rows() - (i+1)`. The easiest way to reverse-sort is to negate the sorting index, as in `make_pair(-abs(...), i)` above. Alternatively, you can define your own sorting operator; this is verbose in our C++03, but would be cleaner once we switch to C++11.

fritzo:
:+1: Sorting logic looks correct, but could be simplified as described in comments.

packages/nimble/inst/include/nimble/EigenTypedefs.h
fritzo: Another simple way to reverse sort is to use [`std::reverse`](http://en.cppreference.com/w/cpp/algorithm/reverse) after sorting:
```c++
std::sort(sortIndices.begin(), sortIndices.end());
std::reverse(sortIndices.begin(), sortedIndices.end());
```

danielturek:
@fritzo @NLMichaud Is this ready to merge?

If so, then per Fritz's guidelines on "ownership of PRs", it's up to @NLMichaud to merge this in.

NLMichaud:
Thanks a lot for the code review @fritzo.  I've made most of the suggested changes - one question though. In terms of the ordering of the `symmetric` argument:  I chose to insert `symmetric` in between the matrix argument `x` and the `only.symmetric` argument to mirror the ordering of arguments in R's `eigen` function.  However, I realize this has the potential to break someone's code (or at least make it less efficient) if they currently have something like `eigen(x, FALSE)`.  I'm guessing we want to keep our argument ordering aligned with R's, so inserting the new argument may be the lesser of two evils.

@danielturek Sorry for the delayed response!  I'm traveling currently, and would have no problem with someone else merging this. Otherwise, I'll let testing run and should have time to merge sometime later tonight.  Thanks again!

danielturek:
No problem @NLMichaud.  Definitely smart to let the testing finish.  I'll watch for the merge (hopefully) later today.  Thanks!

fritzo:
@NLMichaud Your ordering decision sounds good to me!

packages/nimble/inst/include/nimble/EigenTypedefs.h
```diff
@@ -85,22 +85,72 @@ nimSmartPtr<EIGEN_SVDCLASS_R>   EIGEN_SVD_R(const Eigen::MatrixBase<Derived> &x,
 }
 
 template<class Derived>
-void EIGEN_EIGEN_INTERNAL(const Eigen::MatrixBase<Derived> &x, bool valuesOnly, EIGEN_EIGENCLASS_R *returnClass) {
-  // nimSmartPtr<EIGEN_EIGENCLASS> returnClass = new EIGEN_EIGENCLASS;
-  returnClass->getValues().initialize(0, 0, x.rows());
+bool EIGEN_CHECKSYMMETRY(const Eigen::MatrixBase<Derived> &x) {
```
fritzo: nit: In C++ a common convention is to reserve ALL_CAPS_SNAKE_CASE identifiers for macros. I would recommend renaming to something like `is_symmetric`.

packages/nimble/inst/include/nimble/EigenTypedefs.h
```diff
@@ -85,22 +85,72 @@ nimSmartPtr<EIGEN_SVDCLASS_R>   EIGEN_SVD_R(const Eigen::MatrixBase<Derived> &x,
...
+	for(int i = 0; i<x.rows(); i++){
+		for(int j = i + 1; j<x.rows(); j++){
+			if(x(i,j) != x(j,i)){
+				return(false);
```
fritzo: nit: In C++, return is a keyword rather than a function, so `return false;` is more idiomatic.

danielturek:
Great, thanks @NLMichaud !

PR #527, opened 2017-08-03
Refine WAIC algorithm and add more involved example.

NLMichaud:
This PR revises the `calculateWAIC()` method of objects created from a call to `buildMCMC()`.  The new algorithm for WAIC calculation should be orders of magnitude faster, and includes more stable handling of log probabilities.

Additionally, another test example of WAIC calculation is included, this time for a model with 900+ data points.  The WAIC calculated by NIMBLE is compared to results obtained using the `stan` and `loo` packages.  @perrydv The new test requires data to be loaded from a website to create the testing model.  Specifically the data for the model, named `wells`, is obtained using the lines:

```
  url <- "http://stat.columbia.edu/~gelman/arm/examples/arsenic/wells.dat"
  wells <- read.table(url)
```
Will it be a problem to have urls accessed from testing?


fritzo:
Reading urls during testing can make the tests flaky (fail due to network error). I would recommend testing the result and using `testthat::skip('No network connection)` on error. Indeed this is the example use case given for `testthat::skip()`: https://github.com/hadley/testthat/blob/master/R/skip.R#L32

danielturek:
@NLMichaud I was trying out `MCMC$calculateWAIC`, and also the speed-ups in this PR.

Below is a model (not minimalistic, sorry about that), for which the previous `calculateWAIC()`, which is on devel, runs ok, but running on this branch (`fasterWAIC`) has a heap of problems.

Without looking into it at all, perhaps it has to do with the multivariate data nodes?

```
    library(nimble)
    code <- nimbleCode({
        x ~ dgamma(1, 1)
        a  ~ dnorm(0, x)
        a2 ~ dnorm(0, tau = 3*x+0)
        b  ~ dpois(0+5*x)
        b2 ~ dpois(1*x*1)
        c ~ dgamma(1, 7*x*5)
        for(i in 2:3) {
            jTau[i] <- 1
            jNorm[i] ~ dnorm(c * (a+3) - i, var = jTau[i])
            kTauSd[i] <- 2
            kLogNorm[i] ~ dlnorm(0 - a - 6*i, kTauSd[i])
        }
        z[1:3] ~ dmnorm(mu0[1:3], prec = ident[1:3,1:3])
        mu_y2[1:2] <- asCol(A[1:2]) + B[1:2,1:3] %*% asCol(z[1:3])
        mu_y3[1:3] <- asCol(A[1:3]) + B[1:3,1:3] %*% asCol(z[1:3])
        mu_y5[1:5] <- asCol(A[1:5]) + B[1:5,1:3] %*% asCol(z[1:3])
        y2[1:2] ~ dmnorm(mu_y2[1:2], prec = prec_y[1:2,1:2])
        y3[1:3] ~ dmnorm(mu_y3[1:3], prec = prec_y[1:3,1:3])
        y5[1:5] ~ dmnorm(mu_y5[1:5], prec = prec_y[1:5,1:5])
    })
    mu0 <- rep(0,3)
    ident <- diag(3)
    A <- 11:15
    B <- matrix(1:15, nrow=5, ncol=3, byrow=TRUE)
    prec_y <- diag(1:5)
    constants <- list(mu0=mu0, ident=ident, A=A, B=B, prec_y=prec_y)
    data <- list(y2=1:2, y3=1:3, y5=1:5)
    inits <- list(a = 0, z=rep(0,3))
    conf <- configureMCMC(Rmodel)
    Rmcmc <- buildMCMC(conf)
    Cmcmc <- compileNimble(Rmcmc, project = Rmodel)
    niter <- 10000
    set.seed(0)
    Cmcmc$run(niter)
    waic <- Cmcmc$calculateWAIC(burnIn = 2000)
    print(waic)
    ## 20.46023

Warning: in nimArr_2_ManyModelAccess, accessor larger than NimArr!
Warning: after completing ManyModelAccess_2_nimArr, nimCurrent != nimEnd. Perhaps the NimArr was longer than the accessor?
Warning: in nimArr_2_ManyModelAccess, accessor larger than NimArr!
Warning: after completing ManyModelAccess_2_nimArr, nimCurrent != nimEnd. Perhaps the NimArr was longer than the accessor?
Warning: in nimArr_2_ManyModelAccess, accessor larger than NimArr!
Warning: after completing ManyModelAccess_2_nimArr, nimCurrent != nimEnd. Perhaps the NimArr was longer than the accessor?
Warning: in nimArr_2_ManyModelAccess, accessor larger than NimArr!
Warning: after completing ManyModelAccess_2_nimArr, nimCurrent != nimEnd. Perhaps the NimArr was longer than the accessor?
Warning: in nimArr_2_ManyModelAccess, accessor larger than NimArr!
Warning: after completing ManyModelAccess_2_nimArr, nimCurrent != nimEnd. Perhaps the NimArr was longer than the accessor?
```

danielturek:
@NLMichaud Confirmed, the problem is arising because of multivariate data nodes.

danielturek:
@NLMichaud Is it ok if I push stylistic changes (only) to this `fasterWAIC` branch?

NLMichaud:
Sure thing @danielturek!  Make any changes as you see fit.  And thank you for finding and investigating that multivariate data node bug!

danielturek:
@NLMichaud I tested the fixes on this branch, and it seems to work great with multivariate data nodes.  LGTM.

danielturek:
@NLMichaud I tested the fixes on this branch, and it seems to work great with multivariate data nodes.

danielturek:
@NLMichaud One more question.  I'm sorry that I didn't raise this earlier, but how set are we on the argument name `burnIn` that's used by `calculateWAIC()`?  Is it too late to change the name of that argument?

I ask because:
- `MCMCsuite()` uses an argument called `burnin`, and it's much, much too late to change that one!
- `runMCMC()` uses an argument called `nburnin`, which was selected to be consistent with the other arguments, `niter` and `nchains`.

So, I'm trying to avoid adding **another** spelling of a "burn-in" argument.  In my ideal world, I think I would mimic the `nburnin` name, that's used by `runMCMC()`.

What do you think?

NLMichaud:
Good point @danielturek.  I definitely agree that adding another spelling of "burn-in" is one too many, and support the changing the argument name to `nburnin`.  Would it make sense to also detect whether a user has provided a `burnIn` argument, and if so give them a warning that the `burnIn` argument name is deprecated and will no longer be supported in future versions of NIMBLE?

danielturek:
@NLMichaud Yes, I guess it makes sense.  Do you know how much usage `calculateWAIC()` has yet?  I really don't know.  Also, I'm not sure how easy it would be, but I suppose you could add the new argument `nburnin`, and then if the old argument `burnIn` has a value other than the default value (like -100 or something), then issue a warning.  What do you think?  Are you up for making the change?

NLMichaud:
@danielturek I'm definitely up for making the change!  I have no good sense for how much use `calculateWAIC()` has, but I know at least one user has used it (the user who provided the impetus for the refined algorithm).   I just pushed a commit that should implement the naming changes.  One quirk: I tried using -100 as the default value for the `burnIn` argument, but found that currently negative numbers can't be used as default values (this is filed as issue #562).  Thus `burnIn` now defaults to 0, and issues a warning if a non-zero number is detected.

danielturek:
@NLMichaud Thanks for jumping on this change.  That certainly is an oddity about negative default values for arguments!

Only wanted to mention that aside from the naming of `burnIn` (which I know you're addressed), I'm happy with the state of the `calculateWAIC()` function.

NLMichaud:
Great, thanks @danielturek and @fritzo for the code review!

PR #536, opened 2017-08-10
Add default value for name arg of nimbleModel

fritzo:
## Why?

RStudio warns about missing values. This PR pacifies RStudio.


perrydv:
Simple and clear.  👍

PR #541, opened 2017-08-15
Add first unit test of exprClass transformations

fritzo:
**Status:** Do not merge, for review only.

## User Story

A user files a bug like "`step(inprod(x,y))` does not compile". As a developer, I want to write a failing unit test for this bug. This test should exercise only the compiler stage responsible for this bug. But the only test I can currently write is an end-to-end test that parses, generates C++, and compiles. 

## Description

This PR starts to create infrastructure for **unit tests** of Nimble's compiler, i.e. small fast tests that exercise only one part of the compiler chain. This approach contrasts many of the existing **end-to-end tests** of the compiler, such as those in `test-math.R`.

The goal of these unit tests is to achieve more complete testing and reduced test latency. These goals can be achieved by factorizing our tests: Rather than running `N1 x N2 x ... x Nk` end-to-end tests of a `k`-stage compiler, we can run `N1 + N2 + ... + Nk` unit tests over components. However this forces us to specify the `k-1` intermediate representations between each stage. Most of these intermediate representations are just `exprClass` instances with various fields initialized.

### DIFF
--- a/packages/nimble/inst/tests/test-genCpp.R
+++ b/packages/nimble/inst/tests/test-genCpp.R
@@ -3,22 +3,41 @@
 
 source(system.file(file.path('tests', 'test_utils.R'), package = 'nimble'))
 
-skip_if_not_installed('pryr')
-
 context("Unit tests of code transformations")
 
-RparseTree2ExprClasses <- nimble:::RparseTree2ExprClasses
-nimDeparse <- nimble:::nimDeparse
+# These wrappers help convert between exprClass and R parse trees.
+code_to_expr <- nimble:::RparseTree2ExprClasses
+expr_to_code <- function(expr) {
+    if (!inherits(expr, 'exprClass')) stop(paste('Expected an exprClass, got', class(expr)))
+    deparsed <- nimble:::nimDeparse(expr)
+    # nimDeparse returns a list rather than a character vector,
+    # and after parsing, the code is wrapped in expression(-).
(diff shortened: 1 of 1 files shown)


fritzo:
@perrydv Last week I suggested that we could try to "write R like we're writing Haskell", since Haskell is such a great language for writing compilers. Let me explain what that might mean in the context of this PR.

First, Haskell is strongly typed, and programmers practice type-driven-development, where they first specify exactly what values can be realized (e.g. for an `exprClass` object), and then they write recursive algorithms against that datatype. As we make Nimble more modular, it will help us to specify the types of intermediate representations, both as documentation, and for unit testing. One benefit of type specifications is that it allow us to use QuickCheck-style libraries like [quickcheck](https://github.com/RevolutionAnalytics/quickcheck) or [chekr](https://github.com/peterhurford/checkr) (unfortunately neither are on cran, but we could still use that style of testing).

Second, Haskell prefers pure functions, like R without reference classes. This makes it easy to test just one stage of a compiler, since you can easily create inputs and outputs. By contrast, as I try to write unit tests for Nimble's compiler stages, I'm having difficulty setting up enough contextual state in `RCfunProcessing` etc. It would be nice to factor out some of `RCfunProcessing`'s logic to be purely functional: stages should need only inputs and outputs. That would allow us to more exhaustively test compiler components (with very low computational cost).

PR #542, opened 2017-08-15
fix to autoBlock using new sampler defaults

danielturek:
Updated autoBlock system to correctly use new system with sampler control defaults.

Also, removed printing the MCMC progress bar, during the autoBlock preliminary runs.

Running this fix through all testing before merging.


perrydv:
👍

PR #545, opened 2017-08-16
Add model macros as experimental feature

perrydv:
[Design Doc](https://github.com/nimble-dev/nimble/wiki/Model-Macros)

This pull request would add model macros to `nimbleModel` processing as an experimental feature (enabled by `nimbleOptions(enableModelMacros = TRUE)`).

In early discussion, these were referred to as "BUGS modules".  At this point they are tentatively called "model macros".  The term "BUGS" is omitted because this is not part of the original BUGS language.  The term "macro" seems more appropriate than "module" because "macro" typically refers to code expansion and replacement, which is what this feature does.

Based on recent discussions, this feature would allow a line of model code to be replaced by one or more new lines.

The macro processing can be triggered either by a RHS label, such as
`x ~ macro1(y)` 
or by using the macro label for an entire line, such as
`macro1(x, y)`.

To create a macro processor, a programmer must write a function that takes inputs from a model line and returns (a list with) new code to replace the original line.  

It became apparent that there are multiple different ways a programmer might wish to take the input information as arguments to their macro processor, including:

1. As an entire line, such as `x ~ macro1(y)` as a single argument.
2. Decomposed into LHS, RHS, and an logical variable indicating whether the declaration was stochastic (used `~`) or deterministic (used `<-`).
3. Possibly further decomposed, with the RHS unpacked into (possible named) arguments, each passed separately.
4. If using the second format above (without a `~` or `<-`), one might want to take the entire line as one argument or to take arguments separately.

Two arguments to `make_model_macro`  allow control over how arguments from the model line are composed or decomposed and passed to the user-provided function for the macro processor.   The default is option 3 above.

Examples can be found in `test-macros.R` and in the draft roxygen entry for `make_model_macro` in `BUGS_macros.R`. (Although we decided not to name them "BUGS macros", our file naming convention uses the "BUGS_" prefix for all code files primarily associated with model processing.  Hence the filename does contains "BUGS".)

---
This change is [Reviewable](https://reviewable.io/reviews/nimble-dev/nimble/545)


paciorek:
I look briefly and don't have particular suggestions except that it occurs to me that in some cases a macro processing function will generate new variable names and perhaps we should provide guidance on this so that a macro developer can avoid having their variables conflict with the names of user-defined variables in BUGS code. So perhaps something like having "__varName" or the like. Thoughts?

packages/nimble/R/BUGS_macros.R
```diff
@@ -0,0 +1,126 @@
+
+#' EXPERIMENTAL: Turn a function into a model macro builder
+#'
+#' A model macro expands one line of code in a nimbleModel into one or more new lines.  This supports compact programming by defining re-usable modules.  \code{model_macro_builder} takes as input a function that constructs new lines of model code from the original line of code.  It returns a function suitable for internal use by \code{nimbleModel} that arranges arguments for input function.  Macros are an experimental feature and are available only after setting \code{nimbleOptions(enableModelMacros = TRUE)}.
```
fritzo: nit: To ease the job of github reviewers, please limit line length to 80 characters. Emacs even has a macro to wrap long lines into paragraphs.

packages/nimble/R/BUGS_macros.R
```diff
@@ -0,0 +1,126 @@
...
+#' model1 <- nimbleModel(
+#'     nimbleCode(
+#'     {
+#'         ## Create a "for" loop of dnorm declarations by invoking the macro
+#'         x ~ all_dnorm(mu, start = 1, end = 10)
+#'     }
+#'     ))
+#' 
+#' ## show code from expansion of macro
+#' model1$getCode()
```
fritzo: nit: It would be helpful to show the expanded code here.

packages/nimble/R/BUGS_macros.R
```diff
@@ -0,0 +1,126 @@
...
+#' 
+#' model2 <- nimbleModel(
+#'     nimbleCode(
+#'     {
+#'         flat_normal_priors(mu, beta, gamma)
+#'     }
+#'     ))
+#' 
+#' ## show code from expansion of macro
+#' model2$getCode()
```
fritzo: ditto: show the expanded code

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -199,6 +200,58 @@ codeProcessIfThenElse <- function(code, constants, envir = parent.frame()) {
...
+    ## Check if this line invokes a submodel.
+    ## This can be done in two ways:
+    ## (i) node1 [<- | ~] <macro name>(...)
+    ## or
+    ## (ii) <macro name>(args)
+    ##
+    ## The first version is more BUGS-like.
+    ## The second version allows more full control.
+
+    ## Check for first version:
```
fritzo: This confused me a little, since at first glance I didn't see that both forms are supported. Maybe it would clearer to comment like
```r
possibleMacroName <- deparse(code[[1]])
## Reduce first form to second form.
if(possibleMacroName %in% c('<-', '~')) {
  possibleMacroName <- deparse(code[[3]][[1]])
}
## Check for second form.
...
```

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -199,6 +200,58 @@ codeProcessIfThenElse <- function(code, constants, envir = parent.frame()) {
...
+                            possibleMacroName,
+                            " should return a list with an element named ",
+                            "'code'.  It did not return a list."))
+            if(!is.call(expandedInfo[['code']]))
+                stop(paste0("submodel ",
+                            possibleMacroName,
+                            " should return a list with an element named ",
+                            "'code' that is a call."))
+            ## possibly extract other content in the future
+            recursedNewCode <- codeProcessModelMacros(expandedInfo$code, envir)
```
fritzo: It seems a little more robust to simply assign `code <- codeProcessModelMacros(...)` here, rather than `return()` early. That way if someone adds postprocessing at the end of this function, it will apply to both code paths.

packages/nimble/inst/tests/test-macros.R
```diff
@@ -0,0 +1,222 @@
...
+                list(LHS = LHS, RHS = RHS,
+                     OP = if(stoch) as.name('~') else as.name('<-'))
+            )
+            list(code = ans)
+        },
+        use3pieces = TRUE, ## default
+        unpackArgs = FALSE  ## NON-default
+    )
+    
+    temporarilyAssignInGlobalEnv(testMacro2)
```
fritzo: Do I understand correctly that the `temporarilyAssignInGlobalEnv` is required by `nimbleModel` but not by `nimble:::codeProcessModelMacros`, since the latter searches in its parent frame? If that is the case, it might be a good global fix to modify `nimbleModel` to search in its parent frame as well. Do you think that would work?

packages/nimble/inst/tests/test-macros.R
```diff
@@ -0,0 +1,222 @@
...
+            RHS[[1]] <- as.name('step') ## something valid
+            ans <- substitute(
+                OP(LHS, dnorm(RHS, 1)),
+                list(LHS = LHS, RHS = RHS,
+                     OP = if(stoch) as.name('~') else as.name('<-'))
+            )
+            list(code = ans)
+        },
+        use3pieces = TRUE, ## default
+        unpackArgs = FALSE  ## NON-default
```
fritzo: nit: These "default" and "NON-default" comments will be silently invalidated if you ever change the default behavior.

packages/nimble/inst/tests/test-macros.R
```diff
@@ -0,0 +1,222 @@
...
+
+nimbleOptions(enableModelMacros = TRUE)
+
+context("Testing model macros")
+
+test_that('Macro expansion 1',
+{
+    ## This converts a ~ testMacro1(b)
+    ## to inputs as (stoch = TRUE, LHS = a, b)
+    testMacro1 <- nimble:::model_macro_builder(
```
fritzo: nit: Since you've encapsulated these macros in a `test_that('', {})`, it is clearer to avoid the numeric suffix in `testMacro1`, `testMacro2`, etc. That way it is easier to copy-and-modify tests and to remove or reorder existing tests without breaking some vestigial numerical ordering.

packages/nimble/inst/tests/test-macros.R
```diff
@@ -0,0 +1,222 @@
...
+        nimbleCode({
+            testMacro5(a[1], b[2], c[3])
+        })
+    )
+    expect_identical(
+        model$getCode()[[2]],
+        quote(a[1] <- b[2] + c[3])
+    )
+})
+
```
fritzo: Add a test for recursion by defining one macro that uses another. As far as I can tell, none of the existing tests exercise the recursive expansion logic.

packages/nimble/R/BUGS_macros.R
```diff
@@ -0,0 +1,126 @@
...
+#' @examples
+#' nimbleOptions(enableModelMacros = TRUE)
+#' nimbleOptions(verbose = FALSE)
+#' 
+#' ## Example 1: Say one is tired of writing "for" loops.
+#' ## This macro will generate a "for" loop with dnorm declarations
+#' all_dnorm <- model_macro_builder(
+#'     function(stoch, LHS, RHSvar, start, end, sd = 1) {
+#'         newCode <- substitute(
+#'             for(i in START:END) {
```
fritzo: As @paciorek mentions, variable collision is a major problem. In this example, `i` can collide with bound variables, e.g.
```r
for (i in 1:10) {
  x[i,1:10] ~ all_dnorm(mu, start = 1, end = 10)
}
```
(there's probably also an indexing issue)

Some possible solutions include:
- [alpha renaming](https://en.wikipedia.org/wiki/Lambda_calculus#Alpha_equivalence) bound variables to avoid conflict (in this case, renaming the macro's `i` to avoid conflict). I believe this could be implemented within `model_macro_builder` in a three step process: (1) rename all variables in the arguments; (2) expand macro; (3) revert all variable names in arguments and resolve name conflicts, e.g. by adding `_1`, `_2` suffices.
- provide a helper `fresh_variable()` to create a variable with no collisions. This would involve either (1) lots of wiring to pass around the set of variables to avoid, or (2) creating a global fresh variable id (which makes testing difficult).

packages/nimble/inst/tests/test-macros.R
fritzo: Also add a test of variable name conflict. It's fine if that test fails for now (so that you can merge a partial implementation).

fritzo:
I agree with Chris that variable binding and naming will be tricky. Two questions that are unaddressed so far are:
- How to deal with variable name collision?
- How to deal with indexing, e.g. is `x[i,1:10][j]` equivalent to `x[i,j]`?

packages/nimble/inst/tests/test-macros.R
fritzo: Add a test of multiple indexing `x[i,1:10][j]`. It is fine if the test fails for now.

perrydv:
I've been delayed revising this PR, and I want to incorporate some of the suggestions - thanks.

For now I want to follow up on discussion of the design questions that arose:

1. Name collision: I see clear explanation of macro-generated names (including possibly options built by a macro-developer for alternative name-generation schemes) as the responsibility of a macro developer, who should also document their macro's options and behavior.  And I see avoiding name conflicts as the responsibility of a nimbleModel programmer, after reading documentation provided by any macro developer (whether in part of the nimble package or provided from elsewhere).  We could build error-trapping that would tell a user if they have stumbled into a name conflict.  But I wouldn't want to force all macro names to have leading underscores or other such solution, because a responsible nimbleModel programmer might often want to do something with those variables (either use them in another part of the model, or look at their posterior from MCMC, or whatever), so the macro-generated variables should not be forced into awkward names.  An example of macro-generated names might be that if there is a factor "A" used in the macro, then the macro might generate a vector "A.effects".  Finally, the expanded code is available for inspection, so a user can easily determine the source of a conflict.  In summary, I think it's ok for the design of the macro-replacement system to not be clever about name conflicts other than finding them and warning the user.
     If others think I am being too cavalier about this, we could make it an option *chosen by the macro developer* when they call `make_model_macro` to have NIMBLE automatically avoid name conflicts by inserted characters as suggested.  In this case, the macro developer should state this behavior in their documentation, so a nimbleModel programmer can find the explanation if macro-generated names are not what they expected.

2. @fritzo's second question is a good one.  I think the idea is that if a macro developer writes a macro that takes a variable and uses it component-wise in some way, then how would it use a variable provided by a slice?  E.g. if `x[1:10]` as input as some particular macro argument means that `x[j]` will be used in a for loop from 1 to 10, then how will `x[i, 1:10]` be used?  `x[i, j]` with `for(j in 1:10)` would make sense, but must we formalize some logic around that for universal application to all macros?  Again my thought here is that it is up to the macro developer to implement and document as flexible (or as narrow) a system as they choose.  If they want to implement and document their macro's behavior with input provided as a slice, they can.  In the other hand, if they define a slice expression to be invalid input, they can error-trap it if they choose to.  Again, we could add an error-trapping system that would warn them if a particular macro-expansion fails.

**The feature in this pull-request is simply the macro-expander**: It sees when there is a macro, calls the macro, puts the resulting code in place, and recurses if necessary.  The macro programmer has *full flexibility* beyond that to design and document a useful macro for their needs.

Further discussion welcome.

fritzo:
Re: name collision. The current design decision is to burden macro authors with avoiding name conflicts. It seems like this choice will inflict pain on our future users and will sign us up for difficult debugging work.

I don't have a complete solution to avoiding name collisions, but here are some possible short-term solutions that could save our users future pain:

1. Hide the fully general macro builder as internal to nimble, and expose instead a few safer more limited versions. For example, we could provide simple safe substitution via a function like
    ```r
    #' Create a model macro from a fragment of BUGS code.
    #' @param code a fragment of BUGS code.
    #' @param ... a list of \code{name}s to be safely substituted.
    make_model_macro_from_fragment <- function(code, ...) {
        # Extract variable names to substitute.
        vars_to_substitute <- as.character(c(...))
        return(function(...){  # Result inputs name=var parts.
            env <- as.environment(...)
            # TODO Detect name conflict and error for variables bound by for().
            return(substutute(code, env))
        })
    }
    ```
    Indeed this may suffice for our GLM and NBP examples.

2. Provide helper functions like `fresh_variable()` and `safe_substitute()` that help macro authors write robust macros. This doesn't really help beginner users, since it only adds complexity to the macro mechanism interface. But this does let us move forward quickly: we can release the unsafe `make_model_macro()` now and release safety measures later.

3. Expose this general `make_model_macro()` function now to learn how users use it, then replace it with a few safer special purpose macro builders once we've learned what users want.

If these seem too difficult, it probably suffices to address conflicts in bound variables first.
Bound variables in model code are variables that that are introduced as the loop variable in a `for()` inside that model code. I would guess that half of all macros will contain loops like `for (i in 1:N)`; these bound loop variables will conflict as soon as they are embedded in other loops. Bound variables are special since they can be automatically renamed ("alpha converted" in CS jargon) without changing the meaning of code. At the very least we should detect these collisions.

perrydv:
I have updated this pull request for model macros.  Thanks @fritzo and @paciorek for the comments.  Rather than respond to each, I summarize changes as follows:

- All new code did respect 80-character width.  A comment requested that in addition the roxygen entry do the same.  This is a bit of a pain in aquamacs (the #' roxygen annotation is not replicated by fill-paragraph) but I have done it.

- The results of the `model1$getCode()` and `model2$getCode()` in the roxgen @examples entry have been included as comments, as suggested.  This is not necessarily standard in roxygen examples, but it was requested and does make the example more readable without executing it oneself.

- Comments around `possibleMacroName` clarified per request (but slightly different).

- An indexed naming scheme (`testMacro1`, `testMacro2`, etc) across multiple tests in `test-macros.R` was removed (`testMacro` in first test, `testMacro` in second test, etc.).

- Added error-trapping of failed macro expansions (for any reason), including a message showing the levels of macro recursion, if any.

- Error-trapping of duplicate variable declarations in (any) model code already exists.  The message was expanded to explain how to inspect code post-expansion in case this is the culprit.

- Error-trapping of duplicate nested for loop indices (for(i in 1:10{ for(i in 1:5) <...>})) was added (whether from macros or not), with a message explaining how to inspect code post-expansion in case this is the culprit.

- Tests of correctness of the above three kinds of error-trapping were added.  These use the dangerous `try()`-within-`test_that` scheme, but they are safe because immediately following the `try()` I included an expectation that an error occurred.  This allows a subsequent expectation to check that the correct error message was generated.  That latter check can't be done with `expect_failure`.

- Utility for generating unique variable names was *not* added to this PR.  It could be added later.

Thanks @fritzo for trying the Reviewable tool.  I was not taken by it as a big improvement, but please let me know if I should give it a second chance.

If tests pass, and if no major concerns arise, I'd like to get this PR merged without delaying it for more core macro features.  Additional error-trapping, tests, and macro-programming tools can be added later, as long as the core functionality is ok from this PR.  Points from comments that I have not addressed seem to me ok to skip for now.

PR #549, opened 2017-08-17
initial benchmarking of building and compiling models and MCMCs

perrydv:
This PR adds inst.test-benchmark-building-steps.R.

It benchmarks the steps of:

- building model (with check = FALSE and calculate = FALSE in first cases)
- compiling model
- configuring MCMC
- configuring MCMC without conjugate
- building MCMC (not allowing conjugacy, in the initial cases)
- compiling MCMC

It outputs a table with a row for each case and a column for the time used by each step.

Currently there are two cases: 

- a model with 1 top-level parameter, 1000 random effects, and 1000 data (each depending on 1 random effect).

- a model with 1 top level parameter, 100 random effects, and 2000 data (20 data depending on each random effect).

This is in the testing directory but is not needed as a test, so it is on the omission list of run-tests.R

It will be easy to expand to other cases and consideration of `check = TRUE`, `calculate = TRUE`, and use of conjugacy.

### DIFF
--- a/packages/nimble/inst/tests/test-benchmark-building-steps.R
+++ b/packages/nimble/inst/tests/test-benchmark-building-steps.R
@@ -32,7 +32,7 @@ timeSteps <- function(code, data = list(), constants = list(), inits = list()) {
     )[3]
     times['Build MCMC (No Conj)'] <- system.time(
         RMCMC <-
-            buildMCMC(MCMCconf)
+            buildMCMC(MCMCconfNoCon)
     )[3]
     times['Compile MCMC'] <- system.time(
         CMCMC <-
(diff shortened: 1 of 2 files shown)


fritzo:
If you exclude this from the test suite, it will eventually break. If you're worried about tests taking too long, you can e.g. parameterize them with an environment variable that defaults to something small, or you can use small numbers based on `Sys.getenv('CI')`.

perrydv:
BTW the reason I turned off the "test" was that it is not intended to test a feature.  Instead it is a benchmark that uses basic features that *should* be probed by other tests.  However I get your point and followed your suggestion and parameterized it (directly in the test file) to be always on but small during routine use.  When working on features to be evaluated by the benchmarks, larger parameters (sizes of model variables) can be used.

Also I messed up the branch flow for this PR.  A subsequent PR that received 👍 *included* the changes in this PR and hence have been merged already, a few moments ago.  I will close this PR and make a new PR to turn the "test" on and parameterize it to be small in routine use.

PR #551, opened 2017-08-18
NodeFunctionVector as list

perrydv:
This pull request changes `nodeFunctionVector` from a reference class to a list.  Its uses are simple, so the change was very simple to make.  Using a list avoids overhead of reference class object construction.  There is a `nodeFunctionVector` made every time a model method needs a vector of nodes.  Evaluation of the `newSetupCode` making these does not happen until compilation of MCMCs (or other algos).  On the two cases in benchmarking of model and MCMC building a compiling, this change reduces MCMC compilation time by an average of about 10%.


fritzo:
Did you forget to push? It looks like this is only benchmarking code.

perrydv:
Yes @fritzo, you're right.  Now I've pushed the main change.

PR #553, opened 2017-08-23
Turn on model building benchmark during routine testing

perrydv:
Per comments on closed PR #549, the current PR turns on `test-benchmark-building-steps.R` but parameterizes it to use small model variables during routine use.  When using these benchmarks to test efficiency of new changes, larger model variable sizes can be used.

### DIFF
--- a/packages/nimble/inst/tests/test-benchmark-building-steps.R
+++ b/packages/nimble/inst/tests/test-benchmark-building-steps.R
@@ -45,24 +45,40 @@ test_that('Benchmarking model and MCMC building and compiling steps',
 {
     caseNames <- character()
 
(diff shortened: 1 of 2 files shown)


packages/nimble/inst/tests/test-benchmark-building-steps.R
```diff
@@ -45,24 +45,40 @@ test_that('Benchmarking model and MCMC building and compiling steps',
 {
     caseNames <- character()
 
-    caseNames[1] <- 'theta->mu[1:1000]->y[1:1000]'
+    ## 1000 is a good size for full benchmarking
+    Benchmark1length <- 10
```
fritzo: Do I understand correctly that your typical workflow is to modify this code and run from within emacs?

fritzo:
:+1:

PR #554, opened 2017-08-23
Fix copy-and-paste error in mathTestLists.R

fritzo:


perrydv:
👍 Whoa, egg on my face.  Thanks for finding this.

PR #555, opened 2017-08-23
warning instead of error for extra variables in data

lponisio:
when debugging models it is often necessary to drop variables from models to find problems but it was difficult to then need to create a new data structure each time. This change allows the user to pass in extra data without breaking the model building, but instead just get a warning that extra variables were passed in. The extra variables are then removed from the data as a cleaning step (in case having extra data downstream) is a not great in the future.


fritzo:
:clap: thank you for removing trailing whitespace!

perrydv:
@lponisio Oops it looks like we left in the failing version of the new test.  It should have `data = data` instead of `data = 123`.  (We used the latter locally to check how the test would fail, then forgot to put in a working version before the PR.)

lponisio:
okay unit test functioning now!

perrydv:
@paciorek and/or @danielturek will you take a look at this proposal to warn instead of error if unnecessary variables are provided in `data`.  Thanks.

paciorek:
LGTM. There are a bunch of formatting changes in the diff so a bit hard to look at everything, so I'm not sure if there's any actual code changes other than the change from error to warning.

perrydv:
Yes I believe all other changes were formatting, so I'll merge this.

PR #556, opened 2017-08-23
format BUGS_BUGSdecl.R

perrydv:
This PR reformats `BUGS_BUGSdecl.R` for readability.

### DIFF
--- a/packages/nimble/R/BUGS_BUGSdecl.R
+++ b/packages/nimble/R/BUGS_BUGSdecl.R
@@ -1114,23 +1114,19 @@ stripDynamicallyIndexedWrapping <- function(expr) {
 }
 
 usedInIndex <- function(expr)
-    if(length(expr) > 1 && expr[[1]] == ".USED_IN_INDEX")
-        TRUE
-    else
-        FALSE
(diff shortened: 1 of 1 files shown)


fritzo:
:+1:  BTW did you do this using formatR or an Emacs script or RStudio's formatter? It would be nice to find an autoformatter that the whole team could use (i.e. works in emacs, rstudio, vim, etc.).

paciorek:
Yeah, I'd like to know for myself what a good tool is.

perrydv:
I did it manually, with Emacs Speaks Statistics determining indentation.  I manually inserted line breaks such that ESS indentation would keep lines <= 80 characters whenever possible.  On occasion with expressions involving long variable names and quoted character strings, lines go > 80 characters.

I might experiment with formatR, but some of its ["Further Notes"](https://yihui.name/formatr/) may imply that we'll need to make a manual pass over files anyway.  I agree it would be nice to have a shared tool.  ESS seems opinionated about formatting (and I don't want to take some dive into its configurations).  But I guess that's ok if we have a standard formatter to use prior to pull requests.

In any case, I'll merge the current formatting PR.

PR #561, opened 2017-08-24
Document nimbleOptions actual behavior

### LINKED ISSUE #558: nimbleOptions(nimbleOptions()) does not work
# Expected Behavior
A common pattern is to save-and-restore options
```r
old <- nimbleOptions()                  # Save original options.
nimbleOptions(foo = TRUE, bar = FALSE)  # Set new options.
# ... do stuff ...
nimbleOptions(old)                      # Restore old options.
```
This should work, since as `help(nimbleOptions)` states, "Options can also be passed by giving a single unnamed argument that is a named list."

# Actual Behavior

```
> nimbleOptions(old)
Error in as.list(.nimbleOptions)[args] : invalid subscript type 'list'
1: nimbleOptions(old)
```

# Workaround

```r
do.call(nimbleOptions, old)
```

fritzo:
Fixes #558


paciorek:
I'd rather we just put the logic into nimbleOptions() that detects if the input is a list and if so does the do.call.

I guess since it's possible to use nimbleOptions with a single name=value pair where 'value' is a list itself, we need to detect if we are getting a list of name=value pairs or a single name=value with value as a list.

If you want to merge as is, feel free, but let's leave this open so that someone can circle back to modify the behavior.

fritzo:
@paciorek That sounds reasonable. How about something like this?
```r
nimbleOptions <- function(...) {
    args <- list(...)
    if (!length(args)) {
        # Get all nimble options.
        return(as.list(.nimbleOptions))
    }
    if (length(args) == 1 && is.null(names(args))) {
        # Set all nimble options from a list.
        args <- args[[1]]
    }
    # ...continue processing args...
}
```

paciorek:
Yep, I think that looks good.

packages/nimble/R/options.R
```diff
@@ -112,26 +112,53 @@ getNimbleOption <- function(x) {
...
+#' str(nimbleOptions(), max.level = 1)
+#'
+#' # Save-and-restore options:
+#' old <- nimbleOptions()                    # Saves old options.
+#' nimbleOptions(showCompilerOutput = TRUE,
+#'               verboseErrors = TRUE)       # Sets temporary options.
+#' # ...do stuff...
+#' nimbleOptions(old)                        # Restores old options.
 nimbleOptions <- function(...) {
     args <- list(...)
```
fritzo: @paciorek PTAL, I think I've implemented this correctly.

paciorek:
LGTM

PR #565, opened 2017-08-31
Fixes #563: Incorrect handling of non-trivial expression in while() condition

### LINKED ISSUE #563: problem with using a function within condition of a while loop in compiled code
This example gives the wrong answer -- the loop is not stopping at the appropriate place. I can look deeper, but wanted to record this for the moment.

```
mynf = nimbleFunction(
    run=function(xi = double(1)) {
        returnType(double(0))
        newInd <- 1
        n <- length(xi)
        while(sum(xi == newInd) > 0 & newInd <= n){ 
            newInd <- newInd+1
        }
        return(newInd)
    })

cmynf=compileNimble(mynf)

cmynf(c(1,1,3,3,4))
# should be 2
> 6 
```
If one uses
`mySum <- sum(xi==newInd) ` and has `while(mySum > 0 & newInd <= n)`, things are fine.

perrydv:
The problem in Issue #563 is the following oversight about handling `while`.  If we have `while(foo(x) > 0) {}`, and `foo(x)` turns out to be an expression for which we generate an intermediate variable in C++, then that intermediate variable was set once prior to the `while` rather than each time the `while` condition is evaluated.  In this example, we would get

```
Interm1 = foo(x);
while(Interm1 > 0) {
... // x might be updated, leaving Interm1 invalid during next iteration
}
```

The proposed solution is to insert all code for intermediates both before the `while` and at the end of the `while` body.  In this case:

```
Interm1 = foo(x);
while(Interm1 > 0) {
...
Interm1 = foo(x);
}
```

It seems a little clunky, but it should yield correct behavior.  Any cases where it won't?

Implementing this required a small utility function (`copyExprClass`) for copying `exprClass` objects while updating the upward link (`caller`) of their doubly-linked structure.

A regression test was added to `test-misc.R` for Issue #563.  A test of the new `copyExprClass` was also added.

### DIFF
--- a/packages/nimble/R/genCpp_insertAssertions.R
+++ b/packages/nimble/R/genCpp_insertAssertions.R
@@ -69,13 +69,16 @@ exprClasses_insertAssertions <- function(code) {
                 if(length(beforeContents) > 0) {
                     ## The "while" is now embedded in newExpr, so it is in a different place in the AST.
                     newWhileBody <- newBracketExpr(
-                        args = list(code$
-                                    args[[i]]$ ## entry in original {
-                                    args[[length(beforeContents)+1]]$ ## entry in newExpr {
-                                    args[[2]],  ## body of while clause
-                                    beforeContents) ## new piece
+                        args = c(code$
+                                 args[[i]]$ ## entry in original {
+                                 args[[length(beforeContents)+1]]$ ## entry in newExpr {
+                                 args[2],  ## body of while clause, as list
+                                 lapply(beforeContents, copyExprClass)) ## new piece
                     )
-                    setArg(code$args[[i]], 2, newWhileBody)
+                    setArg(
+                        code$args[[i]]$args[[length(beforeContents)+1]],
+                        2,
+                        newWhileBody)
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


packages/nimble/R/genCpp_insertAssertions.R
```diff
@@ -31,19 +33,53 @@ exprClasses_insertAssertions <- function(code) {
...
+            }
+            ## For 'if' or 'while', recurse into the if/while clause and (possibly) else clause.
+            ## 'while' causes a special challenge.  Any assertions must go before the condition
+            ## and also at the end of the clause for execution.  Otherwise the assertions will be
+            ## done only once on the first time through.
+            if(handleWhile) {
+                if(length(afterContents) > 0)
+                    stop('Invalid conditional in while()')
+                if(length(beforeContents) > 0) {
+                    ## The "while" is now embedded in newExpr, so it is in a different place in the AST.
```
fritzo: It's good to stand back and observe that this logic might be simpler in a [nanopass compiler](https://www.cs.indiana.edu/~dyb/pubs/nano-jfp.pdf), where there might be an entire isolated compiler stage devoted to rewriting
```r
while (complexExpression(x)) {
    doStuff()
}
```
to
```r
temp_123 <- complexExpression(x)
while (temp_123) {
    doStuff()
    temp_123 <- complexExpression(x)
}
```
Wouldn't that be easy to test!

perrydv:
@paciorek Can you give this PR a quick look?

paciorek:
Ok, I didn't fully wrap my head around how you were managing the exprClasses, but overall LGTM.

PR #566, opened 2017-08-31
Fix #538 - step(inprod(a, b))

### LINKED ISSUE #538: Error compiling step(inprod(x, y)) in BUGS model
Compilation fails on the following model.  I believe the usages are all correct.

Come to think of it, this is probably the same issue as #530.

@perrydv or @fritzo is this something we could get looked at?  I'm guessing / hoping it's an easy fix related to `inprod()`.  This fix would help me moving forward.  Thanks!

```
library(nimble)

nfDef <- nimbleFunction(
    setup = function() {},
    run = function() {
        a <- rep(0, 5)
        b <- rep(0, 5)
        c <- step(inprod(a, b))
    }
)

Rnf <- nfDef()
Cnf <- compileNimble(Rnf, showCompilerOutput = TRUE)

compiling... this may take a minute. On some systems there may be some compiler warnings that can be safely ignored.
clang++  -I/Library/Frameworks/R.framework/Resources/include -DNDEBUG -DR_NO_REMAP -I  -DEIGEN_MPL2_ONLY=1 -I"/Library/Frameworks/R.framework/Versions/3.4/Resources/library/nimble/include" -Wno-misleading-indentation -Wno-ignored-attributes -Wno-deprecated-declarations  -I/usr/local/include   -fPIC  -Wall -g -O2  -c P_6_nfRefClass_R_GlobalEnv48.cpp -o P_6_nfRefClass_R_GlobalEnv48.o
warning: unknown warning option '-Wno-misleading-indentation'; did you mean '-Wno-missing-declarations'? [-Wunknown-warning-option]
P_6_nfRefClass_R_GlobalEnv48.cpp:25:13: error: no matching function for call to 'inprod'
c = nimStep(inprod(a, b));
            ^~~~~~
/Library/Frameworks/R.framework/Versions/3.4/Resources/library/nimble/include/nimble/Utils.h:181:8: note: candidate function not viable: no known conversion from
      'NimArr<1, double>' to 'double' for 1st argument
double inprod(double v1, double v2);
       ^
1 warning and 1 error generated.
make: *** [P_6_nfRefClass_R_GlobalEnv48.o] Error 1
Error: Failed to create the shared library
```

perrydv:
The problem with Issue #538 was in size processing of `step` and other unary functions that can handle scalar or non-scalar input (and operate component-wise on non-scalar input).  

For background, in a case like `step(foo(x))`, where `foo(x)` *cannot* be turned into eigen code, we put the result of `foo(x)` into an intermediate and then proceed to generate eigen code for `step(<Intermediate>)`.  That works fine.  

The missing case was somewhat opposite: `step( inprod(a, b))`, where `inprod(a, b)` can and should be turned into eigen code but then the scalar version of `step`, not its `eigen` version, should be used.  This was not being identified in `sizeUnaryCwise`, resulting in failure to eigenize `inprod(a, b)`.  

This bug impacted all unary functions. E.g., `exp(inprod(a, b))` also did not work.

In this pull request, logic has been added to `sizeUnaryCwise` to create an intermediate if the argument to the unary is marked for eigenization but returns a scalar.

A regression test has been added to `test-coreR.R`.

### DIFF
--- a/packages/nimble/R/genCpp_sizeProcessing.R
+++ b/packages/nimble/R/genCpp_sizeProcessing.R
@@ -2574,11 +2574,26 @@ sizeUnaryCwise <- function(code, symTab, typeEnv) {
     asserts <- recurseSetSizes(code, symTab, typeEnv)
     ## lift intermediates
     a1 <- code$args[[1]]
+    
     if(inherits(a1, 'exprClass')) {
         if(!nimbleOptions('experimentalNewSizeProcessing') ) {
-            if(a1$toEigenize == 'no') {
-                asserts <- c(asserts, sizeInsertIntermediate(code, 1, symTab, typeEnv))
-                a1 <- code$args[[1]]
+            if(a1$nDim == 0) {
+                ## Argument is scalar.
+                ## If it results from vector operation (e.g. inprod)
+                ## lift that to an intermediate
+                if(a1$toEigenize == 'yes') {
+                    asserts <- c(asserts, sizeInsertIntermediate(code, 1, symTab, typeEnv))
+                    a1 <- code$args[[1]]
+                }
+            } else {
+                ## Argument is non-scalar.  In this case, the
+                ## expression will be eigenized, so we must lift the
+                ## argument to an intermediate if it *can't* be
+                ## eigenized.
+                if(a1$toEigenize == 'no') {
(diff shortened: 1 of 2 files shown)


danielturek:
LGTM.  Passes my local test of the failure case.  Thanks @perrydv

PR #570, opened 2017-09-05
created runNimbleMCMC function to compile and run a model return samples and summary stats like JAGS/BUGS

lponisio:
This function parallels JAGS/BUGS implementation where given a model, it compiles the model, configures the MCMC using nimble defaults, and runs the model, or returns the model components without running. I realized the user can then use runMCMC() to run the compiled model outside the function, so I did not give the option of taking the model components it returns as inputs to run the model, because it seemed simpler to just use runMCMC(). I followed the format for the documentation from other scripts. I could not think of a good test other than a "smoke test", i.e., yes this function works and returns the right classes of things, so there is no unit test yet.


danielturek:
Moving our thread of comments to github, for posterity.

danielturek:
After thinking and digesting this for a bit, I think I'm in favor of a separate function as Lauren suggested, called `nimbleMCMC()`.  In my mind, this would be a drop-in replacement for `bugs()` or `jags()` functions.  User supplies the model "code", some version of "data" and "inits", and it proceeds to (internally) build a model object, build an MCMC, compile them, run the MCMC, and return samples.

Other arguments could include:

- `niter`, `burnin`, `nchains`, to control the MCMC.
- `monitors`, to control what's monitored
- `setseed`, to allow setting a fixed RNG seed.
- `waic`, to allow returning WAIC value

How are we on the issues of 
- always returning all the samples?
- returning summary statistics as well?  maybe controlled by an option?
- maybe by default returning all samples, but by setting `returnSamples = FALSE`, only returning summary statistics?

Thoughts?

I kind of feel that if someone wants to interact with the model object in other ways, or customize the MCMC beyond the default spec, or do more complex interactions with the MCMC (e.g., timing, or using `reset = FALSE` to continue a run), then they really should be using the `nimbleModel() -> buildMCMC() -> compileNimble() -> runMCMC()` workflow.

The `nimbleMCMC()` function wouldn't try to replace that workflow, but rather, just provides a one-liner substitute for the simple `bugs()` and `jags()` style calls.  Does that sound reasonable?

danielturek:
@perrydv suggests that `nimbleMCMC()` might itself call `runMCMC()`.

danielturek:
@lponisio I'm keen to keep this moving forward.  One issue, this PR seems to be failing current tests, I think because the newly created source file `runNimbleMCMC.R` isn't included in the `packages / nimble / DESCRIPTION` file.

One option is to add it to the `DESCRIPTION` file.

Another would be to move the function definition of `runNimbleMCMC()` into the file `MCMC_run.R`, which is my personal vote.  What do you think?

danielturek:
@paciorek suggests: we might want our argument names to be similar to the names in the jags/bugs functions but of course we also want them to be similar to argument names we've used in other nimble mcmc-related functions. Just a thought of something to look into.

NLMichaud:
@danielturek I like the option  `returnSamples`, defaulting to `TRUE`, but if set to `FALSE` only having summary statistics returned.  Not clear to me if the `waic` argument (or maybe `returnWAIC` ?) should default to `TRUE` or not, I'm a little bit leaning towards yes though. 

One other alternative would be to return from `runNIMBLE()` an R class (maybe a reference class?) with a method to calculate the WAIC after the MCMC has been run, another method to return the posterior samples to the user, and possibly other methods that could be of interest.  Of course, this returned class is beginning to sound pretty similar to the functionality of a compiled nimble MCMC algorithm,  which is what we're planning to take out of the user's hands with `runNIMBLE()`, so this may be a bad idea.

danielturek:
@lponisio Suggest moving the function definition of `runNimbleMCMC()` into the exisiting MCMC file: `MCMC_run.R`.  Then delete the new file that's been created here, `runNimbleMCMC.R`.  Then try pushing that, and see if it passes travis testing.  Let me know if that's not clear.

danielturek:
Passing tests, merging this in.

Thanks again, @lponisio!

PR #571, opened 2017-09-07
Integrate tensorflow gradients with CppAD

fritzo:
**Status:** Ready to merge when reviewed. This is still buggy but failing tests have been wrapped in `expect_error`, and debugging is blocked by underlying bugs in AD.

This PR attempts to integrate Fritz's Tensorflow machinery with Nick's CppAD machinery using CppAD's [User Defined Atomic Op](https://www.coin-or.org/CppAD/Doc/atomic_base.htm) interface.

## Tasks

- [x] Add R code to add gradient ops to the Tensorflow graph before serializing it.
- [x] Expose gradient computations in the `NimTf_Runner` C++ class.
- [x] Create a wrapper class `NimTf_op` to be used in AD computations (like `NimTf_Runner`'s but for `CppAD::AD<double>` types.)
    - [x] Implement `NimTf_op::forward()` and `::reverse()` functions.
    - [x] Implement `NimTf_op::NimTf_run()` interface.
- [x] Use `NimTf_op` in the generated code for AD taping. (or merge it into `NimTf_runner`)
- [x] Get everything to compile.
- [ ] Get tests to pass.

PR #574, opened 2017-09-09
Work around broken compiler flag plumbing of -DR_NO_REMAP

fritzo:
This forces a line `-DR_NO_REMAP` at the top of each generated C++ file. This works around broken plumbing to pass the `-DR_NO_REMAP` flag to the compiler. `-DR_NO_REMAP` is needed for some of our new `Tensorflow`+`CppAD` integration, e.g. for #571 .


fritzo:
bump

PR #578, opened 2017-09-12
merge in DP-related fixes

paciorek:
This PR:

1. adds the dcat-ddirch conjugacy, which we were missing. It does not add dcat-dbeta as that is more involved because the beta is a probability only on p and not on (p, q = 1-p).

2. fixes conjugacy detection when there is random indexing in one parameter and the conjugacy being checked is on the other parameter (this was overlooked in my original work on dynamic indexing for 0.6-6). Also adds a bunch of tests of such conjugacies to tests-dynamicIndexing.R to see if we detect conjugacy.

Note: in the branch I'm having trouble with test-mcmc - it's giving weird errors that don't occur if I run each test one by one. I'm hoping that when run through Travis it will be fine. Given that dynamic indexing is turned off in the branch for the test-mcmc tests, there is no reason the test-mcmc results should have changed.

As an example of what was broken in 0.6-6 and that this fixes:
```
code = nimbleCode({
   y ~ dnorm(mu[k+1], var = s0)
   for(i in 1:2)    
     mu[i] ~ dnorm(0,1)
   k ~ dbern(p)
   s0 ~ dinvgamma(1,1)
})

### CODE REVIEW

File: packages/nimble/inst/tests/test-dynamicIndexing.R [hunk 1]
```diff
@@ -19,54 +19,209 @@ ans3 <- sapply(testsInvalidDynIndexValue, test_dynamic_indexing_model)
...
-              for(j in 1:5) 
-                  mu[j] ~ T(dnorm(0, 1), 0 ,1)
-          })
-          m = nimbleModel(code, data = list(y = rnorm(4)),
-                          inits = list(k = rep(1,4)))
-          conf <- configureMCMC(m)
-          expect_equal(length(grep("conjugate", conf$getSamplers()[[1]]$name)), 0,
-                       info = "incorrectly detected conjugacy")
+
+    code = nimbleCode({
```
fritzo: nit: Ideally these could each go in a separate `test_that("...", {...})` so that they could fail independently. When I'm debugging, I find it really helpful to have as many as possible independent pass/fail variables to help me infer where the error might be.

packages/nimble/inst/tests/test-mcmc.R
```diff
@@ -1175,6 +1209,34 @@ test_that('dcar_normal sampling', {
 })
 
 
+## testing dmnorm-dnorm conjugacies we don't detect
+
+test_that('dnorm-dmnorm conjugacies NIMBLE fails to detect', {
+    code = nimbleCode({
```
fritzo: ditto: consider splitting into multiple `test_that()`s.

paciorek:
@perrydv @fritzo do the test-mcmc failures in Travis look like they might be in terms of the comparison with the gold file?  The error log text seems to be indicating something about formatting (there is language like "dots[[2L]][[2990L]] not identical to dots[[3L]][[2990L]]"

Anyway, if that rings any bells, let me know.

danielturek:
I reviewed this now, and also in detail previously, and everything looks in good order -- both the changes to conjugacy system, and also the new `dcat-ddirch` conjugacy.  This also passed several local tests of my own.

@paciorek Regarding `dcat-dbeta` conjugacy, are we suggesting (possibly in the future) supporting conjugacy of models such as:
```
p[1] ~ dbeta(a, b)
p[2] <- 1 - p[1]
y ~ dcat( p[1:2] )
```
I ask in the context of, this seems convoluted to me, and not like something worth trying to detect and support.  But perhaps I'm missing what was meant, or an actual use case for this?

paciorek:
@danielturek Yes, the dbeta-dcat conjugacy is the one you indicated. Not sure there is a compelling use case.

paciorek:
@fritzo At various times I've gotten tired of having an initial character string as the first arg to test_that and then also an "info" argument string, so grouping tests seemed like it made sense and followed the test_that help page, which describes a test as encapsulating a series of expectations. 

When a test with multiple expectations fails, one gets the "info" information about each failed expectation so grouping seemed ok to me. Can you say more about what is obscured by grouping?

That said, in this case, the cases being tested may be distinct enough anyway that conceptually it makes sense to separate them.

fritzo:
@paciorek I believe you are correct that your usage does show multiple individual failed expectations, but if the test fails due to an error rather than an expectation, then you will only see the first error, since testthat will stop after the first error.

fritzo:
@paciorek @perrydv Probably only the first one or two lines of the gold file mismatch are meaningful. I don't know what would cause the difference, but if there are no obvious errors or quantitative differences then maybe we should just update the gold file.

perrydv:
@paciorek Those error reports are differences in gold-file comparisons.  I don't know if they are problematic or not.  I don't think we have a way to get the test output file from travis, so I suggest you run the test locally and then save the output file (which is compared to the gold file) somewhere safe and manually see what the discrepancies are about.

(Side note: I had an updated version of gold file comparison that used `diff` as a comparison tool.  The messages generated make me think perhaps the old version (compare line by line) is being used.  I can look at that sometime later.)

PR #579, opened 2017-09-12
proper CAR distribution: dcar_proper()

danielturek:
The proper CAR distribution `dcar_proper()` is now in a reasonable place, including density evaluation and simulation, component-wise MCMC support, Roxygen and User Manual documentation, and testing.  I've checked everything for correctness, and was pleased to see that results agree with those from the WinBUGS manual's example using `car.proper()`.  This gives validation of both the CAR density implementation, and the MCMC sampling.

I've implemented the density evaluation `dcar_proper()` in an efficient manner using pre-computed eigenvalues of the adjacency matrix.  These eigenvalues are calculated only once upfront, and are reused thereafter for density evaluation during MCMC sampling.

To achieve this, I had to extend the modelDef processing code, again, to allow for a special case of lifted nodes, where the indexing on the lifted node must be determined by the indexing on a particular subset of its arguments.  This is manifested in a new named list in the modelDef processing, which specifies calls for which particular argument(s) should give the indexing of the final lifted node.  At present, this list has only has one entry, the function used for pre-calculating the eigenvalues of the adjacency matrix.  It appears as below.  This new logic has no effect on any other lifting or model processing, and has passed all our testing.

```
liftedCallsGetIndexingFromArgumentNumbers <- list(CAR_calcEVs = c(3))
```

The implementation of random draws, `rcar_proper()`, is not nearly as efficient, which constructs the full precision matrix, then performs a Cholesky factorization.  This can be improved significantly once we handle sparse matrix operations, but will suffice for the moment, since it's not used in the course of MCMC sampling.  It is, however, functional, and can be used to generate random draws from the proper CAR distribution.

I also added a convenience function `CAR_calcC()` to generate a default value of the parameter `C`, the sparse representation of the adjacency matrix.  Following the WinBUGS specification, there are symmetry constraints on `C` which are not immediately easy to satisfy (as the constraint intertwines `C` with another parameter, `M`, the vector of conditional variances of each dimension).  The Roxygen example for `dcar_proper()` demonstrates how `CAR_calcC()` can be used, which can automatically generate a suitable value for the `C` argument.  It's worth noting, this function was inspired by the **only** example of using `dcar_proper` I have available, taken from the WinBUGS manual, where they go through gymnastics inside the WinBUGS model code, to deterministically construct the `C` argument to `dcar_proper()`.  The implementation of `CAR_calcC()` mimics this deterministic construction.  Anyway, if WinBUGS is our only guide, then this auto-generation of `C` appears to be the **only** practical way to generate the `C` parameter, which is why I streamlined it for users.  Again, usage of this function is shown in the Roxygen example, and also in the User Manual example.

I changed the names of the WinBUGS functions for calculating the bounds on the `gamma` parameter, from `min.bound()` and `max.bound()`, into `CAR_calcMinBound()` and `CAR_calcMaxBound()`, so the compiler works on them (as the NIMBLE compiler would not allow "periods" in the nimbleFunction name).  I also provided aliases using the original WinBUGS names, for compatibility with WinBUGS models.  Again, the Roxygen example and the User Manual example both show how to use these correctly.

I've updated the User Manual chapter on Spatial models, to include information on `dcar_proper()`.  @paciorek , I'll ask you to read through that, make sure it makes sense, and also hopefully add a short paragraph giving a bit more technical background, motivation, and potential uses of the proper CAR distribution.  You'll see a note in there where I'm hoping you can add something.


fritzo:
It looks like you've accidentally added a bunch of latex temp files like NimbleUserManual.log, NimbleUserManual.aux, .... (Maybe your .gitignore is malfunctioning?) Anyway I'd recommend removing them from the PR and squash-and-merging so that these files are never added to the nimble repo.

danielturek:
Nice eye @fritzo.  Definitely slipped by me.

I think I cleaned it up now, but please do take another look.  Thanks!

paciorek:
Some initial comments (we'll probably want to iterate on this PR):

1) I think section 9.3.2 on zero-neighbor regions needs to be updated w.r.t. car_proper since one can update the zero-neighbor regions (but is it the case that the variance is zero, so the update is just a deterministic update to set the value to mu_i?).

2) I'm confused about C (perhaps discussion is best via Skype?) -- much of which is probably just that it takes me a while to wrap my head around the proper CAR propriety conditions whenever I come back to it. 

a) it confused me on p. 81 of Banerjee what 'alpha' is and how it is any different than 'gamma'. It strikes me if we can understand that then that provides simpler conditions on the parameter? Might be worth me looking at the newer edition of Banerjee to see if it is clearer there.
b) If a user just wants to provide an adjacency matrix with weights of 1, and the M matrix with values one over the number of neighbors, then does the C calculation give what we would expect, which should be elements of the ith row that are 1/w_i+  ?
c) another way to put it is whether our specification allows a user to easily specify the most simple CAR model with weights of one and simple neighborhood structure. 

3) Minor requests - can you put in the roxygen formatting for code in the CAR-Proper roxygen, make sure it's clear in manual that one would need help("CAR-Proper") (I think), change from British spelling and check typos.

4) how about a test of dcar_proper, rcar_proper that does the matrix calculations in the most straightforward way in R and checks that the efficient C implementation matches.

danielturek:
The following updates are available on the `car` branch, addressing comments from @paciorek:

- I've added a new test of density evaluation of `dcar_proper` node, comparing against density evaluation using matrix calculation and `dmnorm_chol()`.
- I've added a new test of random simulation from `dcar_proper` node, comparing it against simulation using `rmnorm_chol()` using the same RNG seed.
- The Uwer Manual is more clear about how zero-neighbor regions are handled, specifically for `dcar_normal` vs. `dcar_proper` distributions.  See updated section 9.3.2.
- Clarified in the User Manual about how to find help about `dcar_normal` and `dcar_proper` distributions, using R `help()`.
- Removed British spellings from User Manual, and did a complete spell check.
- Added `\code{}` formatting to the `dcar_proper` roxygen documentation.

@paciorek I'm happy to chat about the most simple model specification using `dcar_proper`.  It can be done, without too much difficulty. Hopefully, we can discuss all this during tomorrow's meeting.

Also, I'm still hoping you can add a few sentences to the User Manual describing technical details or use cases of `dcar_proper`.  There's a placeholder there, in red text.

paciorek:
Some comments for our call tomorrow now that I've spent a bit of time pulling my hair to wrap my head around the various conditions on the proper CAR model:

1) If C is normalized, then the symmetry condition requires that M be the diagonal of 1/number of neighbors (i.e., inverse of D_w in the Banerjee notation), or a constant scaling of that. In that case gamma must be in (-1,1), so open question of whether we even need to calculate the bounds.

2) WinBUGS makes a big deal out of C being normalized, but I believe the only case where gamma bounds would not be (-1,1) are if C is not normalized (such as the Cressie and Chan example that WinBUGS gives, or the example in CAR-proper where we generate C). 

3) I think the most natural things a user might supply are:
a) C based on a normalized W, i.e., C=\tilde{W}
b) an unnormalized,symmetric W that they want to be normalized
c) an arbitrary C for which they might want us to determine M based on the symmetry condition

It's not clear to me a user would want to provide M but not C. Also of note, your scheme to generate C does not produce a normalized matrix. Let's talk more about calcC().

4) Let's talk about if we are checking the symmetry CM=MC condition.

danielturek:
@paciorek I made some significant changes to handling for `dcar_proper`.

In short, the options available for users are:

- Supply both `C` and `M`.
- Supply *neither* `C` or `M`, and it will use all weights = 1
- Begin with your own vector of symmetric unnormalized `weights` (exactly the same argument that's used by `dcar_normal`), and pass it into the `as.carCM` function, which will generate the equivalent `C` and `M` for use in the `dcar_proper` specification.

Please take a look at the updated User Manual, and the roxygen documentation.

Thoughts welcomed.

paciorek:
I'll probably have some rewording to put in for the manual and roxygen (and I know you want some technical text for the manual), but I'll leave that aside for now. 

I'd prefer we not export calcM or other internal functions, unless we have to. Do we?

I've tried in my error messages to have more informative messages, and to have the first part of the message be the name of the function in which the error is occurring. Not sure if others are doing something similar, but can you make the 'something went wrong' errors more informative?

Otherwise LGTM.

danielturek:
@paciorek For what it's worth, the "something went wrong" messages are for internal errors only -- they technically should never, ever, occur, and if they do, it signals an error in the logic of my code, i.e. what should be an impossible situation.

All of my error messages that are user facing (i.e., giving bad arguments, or misuse of a function) try to be helpful, e.g.:
```
stop('length of adj argument to dcar_normal() must be equal to total number of neighbors specified in num argument')
stop('num argument to dcar_normal() can only contain non-negative integers')
```

If you think having "something went wrong in dcar_proper" would be helpful, I can add that, but again, a user should *never ever* get that message.  I'm inclined to leave them, if you don't mind too much.  Thoughts?

paciorek:
ok, makes sense. Perhaps the message could be: "dcar_proper: internal error."

danielturek:
@paciorek 

I've discovered that we can't remove the exports of `CAR_calcC`, `CAR_calcM`, `CAR_calcNumIslands`, etc, because then the building of R models which use `dcar_proper` or `dcar_normal` will fail, since the dynamically generated class objects (not in the NIMBLE namespace) will try calling these functions (when they're necessary to fill in default values of parameters), and won't be able to find them.

I left the error messages as-is.  All the ones that users could encounter give helpful information.

Unless I hear otherwise, I'm leaving this in your hands to add to / revise the roxygen documentation and the User Manual Spatial chapter, as you see fit.

Let me know when you're happy with any revisions you make, and I'll take a look, and we can merge this in.

danielturek:
@paciorek I just saw your other comment (I think it was emailed), suggesting messages:

"Perhaps the message could be: "dcar_proper: internal error."

That's fine with me.  I'll make that change now.

paciorek:
We might be able to use nimble:::foo for those functions, but at the same time I think CRAN has balked at referring to package functions via :::. I can look into this when I am doing the packaging up for the release.

paciorek:
I believe the zero-neighbor case applies to dcar_proper too, so I modified a bit of the language in the user manual about that. @danielturek can you confirm that the code handles 0-neighbors for dcar_proper? 

Also @danielturek can you add explicit argument info to the roxygen for carBounds(), car{Min,Max}Bound for consistency with function documentation in general?

I moved exporting of non-user-facing functions to prep_pkg.R

Otherwise, added a bit of overview text on proper CAR models in manual and did some fairly minor wording editing. 

Apart from comments above, LGTM to merge.

Note I did not regenerate the User Manual pdf or the roxygen rd files (via 'make man'). Probably makes sense if @danielturek does this just before merging.

danielturek:
@paciorek Would it be ok to leave regenerating the manual and roxygen files to when you're building the package for release?  Is there any reason to do that now, before merging this?

danielturek:
@paciorek Confirming that zero-neighbor cases are also handled in `dcar_proper`.

Also, I just added `@param` and `@return` roxygen information for the `carBounds` functions, as suggested.

Pending your thoughts about rebuilding the manual and .rd files, I think this is ready to merge.

paciorek:
Fine by me to leave rebuilding till later.

danielturek:
This PR has passed tests, and I've looked over the changes to the Spatial chapter.

Everything looks good to me, merging in.

PR #580, opened 2017-09-12
Add withNimbleOptions() helper function

fritzo:
This adds a helper `withNimbleOptions()` to make it easier to temporarily set nimble options, e.g.
```r
cnf <- withNimbleOptions(list(experimentalUseTensorflow = TRUE), {
    compileNimble(nf)
})
```

**Note to reviewers:** This PR includes `make man` noise from stale man pages that weren't updated on other PRs.

## Why?

We often temporarily set nimble options in tests, and we often forget to unset those options. The `withNimbleOptions(list(...), {...})` pattern will help prevent these errors. This pattern also makes it easier to compose nimble options (e.g. to temporarily disable tensorflow in one part of a helper function, regardless of whether it was previously enabled).

## How?

The `withNimbleOptions()` function saves a temp copy of the old options. When multiple `withNimbleOptions()` are nested inside each other, these temp copies constitute a stack, and options are push onto and popped off of this stack. This pattern is very common, e.g. used in Python's [`numpy.errstate`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.errstate.html).


fritzo:
@paciorek Does this look ok to merge?

paciorek:
LGTM

fritzo:
Thanks, Chris!

PR #584, opened 2017-09-12
Support chol() in tensorflow backend

### LINKED ISSUE #583: Support Cholesky in tensorflow backend

fritzo:
Fixes #583 

Heinously complex. Reviewer beware.

### DIFF
--- a/packages/nimble/R/genCpp_TFize.R
+++ b/packages/nimble/R/genCpp_TFize.R
@@ -256,6 +256,7 @@ tfTranslate <- function(name) {
             }
             tf$matmul(y, x)  ## Note the transpose.
         },
+        'chol' = tf$cholesky,
         'backsolve' = function(r, x) {
             ## TODO Decide whether to transpose.
             tf$matrix_triangular_solve(r, x, lower = TRUE)
--- a/packages/nimble/inst/tests/mathTestLists.R
+++ b/packages/nimble/inst/tests/mathTestLists.R
@@ -186,7 +186,7 @@ testsMatrix = list(
     list(name = 'backsolve matrix-vector amid expr', expr = quote(out <- arg2[1:2] + backsolve(arg1[1:2,1:2], arg2[1:2] + arg2[1:2])), inputDim = c(2, 1), outputDim = 1, xfail = 'tensorflow.*runs'),
     list(name = 'backsolve matrix-matrix amid expr', expr = quote(out <- arg2[1:2,1:2] + backsolve(arg1[1:2,1:2], arg2[1:2,1:2] + arg2[1:2,1:2])), inputDim = c(2, 2), outputDim = 2, xfail = 'tensorflow.*runs'),
 
-    list(name = 'chol', expr = quote({ A <- arg1; for(i in 1:dim(A)[1]) A[i,i] <- A[i,i] + 10; out <- chol(A) }), inputDim = c(2), outputDim = 2, xfail = 'tensorflow.*runs'),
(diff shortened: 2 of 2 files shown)


paciorek:
Indeed, convoluted.

Do you know if the factor is upper or lower?  We want it to be upper triangular for consistency with R and our Eigen implementations. Otherwise LGTM.

fritzo:
Yes, there is a transpose between R and Tensorflow (one is row major and
the other column major; we simply reverse all matmul operations to avoid
the transpose work). So I since Tensorflow is (transposed) lower-triangular
and R is upper-triangular, everything should match.

On Sep 12, 2017 1:38 PM, "Christopher Paciorek" <notifications@github.com>
wrote:

Indeed, convoluted.

Do you know if the factor is upper or lower? We want it to be upper
triangular for consistency with R and our Eigen implementations. Otherwise
LGTM.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nimble-dev/nimble/pull/584#issuecomment-328977502>, or mute
the thread

paciorek:
I'm not quite following the bit about reversing all matmul operations. If a user writes a nimbleFunction and they are expecting to have the upper triangle when they call chol(), then their subsequent lines of code may fail if we compile to TF. E.g.,
```
factor <- chol(Sigma)
tmp <- backsolve(factor, vec)
```

fritzo:
Not sure how to explain more clearly. If a user puts data in the upper
triangle, then indeed tensorflow will see that data in the lower triangle:
the difference is that R and Tensorflow differ in row- vs column-major
orderings. When a Nimble user views a piece of memory as a coumn-major
upper triangular matrix, Tensorflow sees the same piece of memory as a
row-major lower triangular matrix. Anyway it should work as currently
implemented.

paciorek:
What I wasn't sure of from the documentation I looked at quickly was whether the _output_ is stored in the lower triangle (with zeros or nonsense in the upper) or in the upper, or if both the upper and lower are stored in the output matrix.

fritzo:
Oh that makes sense. You're right that the tensorflow docs are ambiguous. But `tf.cholesky` is used in `tf.cholesky_solve()`, which does specify that the the output is stored in the row-major-lower-triangle (= the column-major-upper-triangle): https://www.tensorflow.org/api_docs/python/tf/cholesky_solve

paciorek:
Ok, but I still think there may be issues when embedded in a nimbleFunction -- e.g., a user would expect to be able to use backsolve and forwardsolve in combination with the result of the chol(), but at the moment it looks like the backsolve and forwardsolve we have for tensorflow do not call tf.cholesky_solve(), so I don't think (on quick glance) that everything fits together.

So presuming you won't have time to think/work through this, that suggests we should leave this PR open until I have time to work on this.

fritzo:
SGTM. I wasn't sure how `backsolve` and `forwardsolve` are used, and and our tensorflow wrappers may incorrectly handle transposes (there are TODOs in the code, but behavior may silently fail).

One thing that would make this linear algebra stuff easier to test would be some custom test tools for linear algebra. We're currently using `test_math()` and `mathTestLists.R` for linear algebra, but those tools are better suited to element-wise ops and reductions. It would be nice to have some tooling to set up various matrices with e.g. NANs in the strict lower diagonal, or random test matrices that were positive definite.

fritzo:
Oh, this PR was already merged, so I'll create an issue instead.

fritzo:
See https://github.com/nimble-dev/nimble/issues/586

PR #587, opened 2017-09-13
Sketch support for tensorflow.SparseTensor

fritzo:
**Status:** Do not merge. For review only.

This sketches a possible implementation of sparse tensor math using the Tensorflow back-end. Note that this only implements the back-end part, and does not register operations like `make_sparse` and `add_sparse_dense` with the Nimble compiler front-end.

For the full range of Tensorflow's sparse operations, see the [Tensorflow Sparse docs](https://www.tensorflow.org/api_guides/python/sparse_ops).


perrydv:
@paciorek Shall we also close this one?  I don't think having it sit as a PR changes its visibility or priority right now.

PR #589, opened 2017-09-14
One-line invocation of MCMC engine: nimbleMCMC()

danielturek:
This PR introduces a new function, `nimbleMCMC()`, a one-liner for invoking the MCMC engine.  This function provides a range of options:

- Executing multiple MCMC chains
- Burn-in
- Thinning
- Specifying monitors
- Providing initial values as a list, a list of lists, or as a function
- Returning any combination of:
  - Posterior samples
  - Posterior summary statistics
  - WAIC values

The entry point for using `nimbleMCMC()` is either:
- The usual `code`, `constants`, `data`, `inits` arguments, or,
- A compiled or un-compiled `model` object.

This PR also includes some minor updates to the existing `runMCMC()` function, the major change being the ability to return posterior summary statistics, and /or WAIC values, in addition to the posterior samples themselves.

In addition, a semi-major refactoring of the MCMC chapter in the User Manual.  This now presents `nimbleMCMC()` as the most direct and user-friendly approach to invoking the MCMC engine, and subsequently the `configureMCMC() -> buildMCMC() -> compileNimble() -> runMCMC()` approach as a longer and more customizable option.

I welcome reviews from @perrydv, @paciorek, and @NLMichaud on the changes to the MCMC chapter, and the new function `runMCMC()` itself.

`NimbleUserManual.pdf` has been rebuilt to reflect these changes.

Also added utility functions `is.model()`, `is.Rmodel()`, and `is.Cmodel()`.


perrydv:
Two general comments / questions, not based on detailed code review:

1. In `runMCMC`, should the handling of `burnin` be to call the `<compiled mcmc>$run()` twice, rather than calling it once and then discarding the first `burnin` samples?  The difference in behavior would be that in the former version, the adaptation parameters would be reset after the burnin.  This could potentially be beneficial if the burning was long and had guided the adaptation parameters to bad values while using up a lot of adaptation clock.  OTOH, it could potentially be harmful if the adaptation parameters turned out well from burnin and then end up discarded.

2. In addition to accepting an uncompiled or compiled model, shouldn't it also accept an uncompiled or compiled MCMC?  I guess the idea is that if a user is advanced enough to re-use an MCMC object for multiple runs, they should go through the more granular workflow.  But I imagine allowing MCMCs to be returned and/or accepted as an argument to `nimbleMCMC` would provide maximal flexibility in workflow.

paciorek:
It would be nice to know more about how JAGS and BUGS handle adaptation to help us here. We could reach out to Andrew again to get his insights.

danielturek:
@perrydv, in response:

1. I'm less inclined to second-guess whether resetting adaptation parameters would be beneficial, and instead stick with what I thought to be standard MCMC burn-in practice: discarding the first `nburnin` samples, thus leaving the parameter values reasonably in the posterior, and also leaving the samplers as being well-adapted, and operating at reasonably high efficiency.  Unless we hear a compelling argument otherwise, I'm inclined to leave it as written.
1. As far as accepting an `mcmc` object as an argument to `nimbleMCMC()` -- if someone is advanced enough to get through `nimbleModel() -> buildMCMC() -> compileNimble()`, they really should be using `runMCMC()` at that, point, which is designed specifically for that purpose: to accept an `mcmc` argument, and run that.  Furthermore, if we modified `nimbleMCMC()` to take an `mcmc` argument, then it would (silently) be ignoring the `thin` argument, which doesn't seem good.  Again, I'm inclined to leave it as written.

If you feel very strongly otherwise on either point, then let me know.

danielturek:
This PR now also includes a very brief introduction to using `nimbleMCMC()` in Chapter 2 of the User Manual, which is the Lightning Introduction -- specifically, Section 2.4.

This is reflected in the re-built `NimbleUserManual.pdf` document.

Thanks for the suggestion, and edits, @NLMichaud !

danielturek:
As discussed during development meeting, merging this in.

For the record, there was discussion about at a later point, modifying the behaviour of `nimbleMCMC()` such that after the `nburnin` samples, the **adaptation clock** for all samplers is reset to its initial state.  The values of model nodes, and sampler tuning parameter values (e.g, scale and proposal covariance) would not be changed.  Only the **clock** governing the decaying adaptation of tuning parameters (internal to the samplers) would be reset to its original state.

PR #597, opened 2017-09-18
Remove calls to browser() from some R files

NLMichaud:
This PR removes a number of the calls to `browser()` that occur after an error is found during compilation.  
Previously, many calls to `browser()` were inserted directly after error messages, in the following form:

```
if(error){
    cat('Informative error message here.')
    browser()
}
```
 In this PR, the `browser()` calls are replaced with calls to `stop()`, with the informative error message included. 

 In some appearances of `browser()`, no error message exists.  I attempted to add informative messages for some of these appearances.  For a few others (e.g.  line 1656 of `bugs_modelDef.R`), the call to `browser()` has been left in place, as it's not clear what error message would be appropriate there.


paciorek:
For the former browser call in sizeAssignAfterRecursing, is there a reason we need the try() and the second stop()? @perrydv can we just have the first stop()?
```
    test <- try(if(inherits(RHStype, 'uninitializedField') | length(RHStype)==0) {
        stop(exprClassProcessingErrorMsg(code, paste0("In sizeAssignAfterRecursing: '",RHSname, "' is not available or its output type is unknown.")), call. = FALSE)
    })
    if(inherits(test, 'try-error')) stop('"In sizeAssignAfterRecursing: Problem with RHStype.')
```

Side note: I'm running into this because surprisingly when I am trying test-math, the norm() test is putting me into a browser, so I was looking into this code. Not sure why that browser was not happening in previous versions when running tests.

perrydv:
@NLMichaud I agree with @paciorek: in the code he referenced, we've ended up with something that doesn't make sense.  I think it could just be:
```{r}
if(inherits(RHStype, 'uninitializedField') | length(RHStype)==0) {
        stop(exprClassProcessingErrorMsg(code, paste0("In sizeAssignAfterRecursing: '",RHSname, "' is not available or its output type is unknown.")), call. = FALSE)
    }
```

In other words, no `try`, and only one `stop`.

Also I like using `call.=FALSE` on all `stop()`s.  I think it makes the messages emitted to users easier to read.

If you make those changes, LGTM and we can clean up remaining `browser()`s in a subsequent PR.

NLMichaud:
Thanks @paciorek, and @perrydv, I've made the suggested changes and will merge once testing is finished.

PR #598, opened 2017-09-19
only add check for zero contribution in conj sampler if variable is dynamically indexed

paciorek:
We have code like this in conjugate samplers:
```
   for (iDep in 1:N_dep_dnorm) {
        if (dep_dnorm_coeff[iDep] != 0) 
            contribution_mean <<- contribution_mean + dep_dnorm_coeff[iDep] * 
                (dep_dnorm_values[iDep] - dep_dnorm_offset[iDep]) * 
                dep_dnorm_tau[iDep]
        if (dep_dnorm_coeff[iDep] != 0) 
            contribution_tau <<- contribution_tau + dep_dnorm_coeff[iDep]^2 * 
                dep_dnorm_tau[iDep]
    }

```

This PR makes it so the check for != 0 is only added if the node is in a variable that is dynamically indexed.

Two shortcomings:
1) code above has two if() statements as I wasn't sure how to wrap multiple lines in an if in 'addForLoopBody'
2) in MCMC_configuration.R I didn't look deeply enough to see how to update the addConjugateSampler call that is part of the new dynamic conj rules. It still needs the 'dynamicallyIndexed' argument that one can see in the standard conjugacy processing code.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -5,7 +5,8 @@ varInfoClass <- setRefClass('varInfoClass',
                                 mins = 'ANY',
                                 maxs = 'ANY',
                                 nDim = 'ANY',
-                                anyStoch = 'ANY'))
+                                anyStoch = 'ANY',
+                                anyDynamicallyIndexed = 'ANY'))
 
 ## A small class for information about a node in the igraph
 graphNode <- setRefClass(
@@ -32,7 +33,7 @@ modelDefClass <- setRefClass('modelDefClass',
                                  ## the following are all set in setupModel()
                                  BUGScode = 'ANY',  ## original BUGS code, set in assignBUGScode()
                                  constantsEnv = 'ANY', ## environment with constants, set in assignConstants()
-                                 constantsList = 'ANY',  ## named list with constants, set in assignConstants()
+                                 constantsList = 'ANY',  ## named list with constants, set in assignConsstants()
                                  constantsNamesList = 'ANY', ## list of constants name objects, set in assignConstants()
                                  constantsScalarNamesList = 'ANY', ## could eventually replace constantsNamesList. added for newNodeFxns
                                  dimensionsList = 'ANY',		#list		   ## list of provided dimension information, set in assignDimensions()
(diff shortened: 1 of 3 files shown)


packages/nimble/R/BUGS_modelDef.R
```diff
@@ -32,7 +33,7 @@ modelDefClass <- setRefClass('modelDefClass',
                                  ## the following are all set in setupModel()
                                  BUGScode = 'ANY',  ## original BUGS code, set in assignBUGScode()
                                  constantsEnv = 'ANY', ## environment with constants, set in assignConstants()
-                                 constantsList = 'ANY',  ## named list with constants, set in assignConstants()
+                                 constantsList = 'ANY',  ## named list with constants, set in assignConsstants()
```
danielturek: Wow, this typo probably dates back a *very* long time...

danielturek:
@paciorek Noting that this PR is currently failing travis tests.

danielturek:
@paciorek Another question, I'm probably missing something obvious here... but why would we only check for coefficient != 0, when the node is in a variable that is dynamically indexed?  Why not **always** check for that, and skip the unnecessary addition, anytime that the coefficient = 0?

danielturek:
@paciorek When the following code is added to the `initialize()` function of the MCMC configuartion (in two places):
```
varName <- model$getVarNames(nodes = node)
if(length(varName) > 1) stop("MCMCconf: conjugate sampler for more than one variable: ", node, ".")
```
I think this check for `length(varName) > 1` is unnecessary, since the way this part of the system works, each `node` here is necessarily either a scalar node, or a single multivariate node; nothing else is possible.

paciorek:
Some responses:

1) The idea of only checking for 0 when we are working with dynamic indexing is that we would only expect a zero multiplier (a _coeff) of 0 if we have something like `y[i] ~ dnorm(mu[1], 1)` with the target being, say, mu[2]. I think the only other time this would occur is if someone had a statement like `y ~ dnorm(a + 0*mu, 1)` the only case I can think of is if there is a variable multiplied by mu and it is discrete and therefore might be zero. Could be the case in model selection or similar, but seems like a special case. 

2) fixed the typo

3) I've taken out the check. I wasn't looking more broadly -- just was using varName to do indexing so thought I should have the check.

4) test failure appears to be a misplaced parenthesis. 

Waiting on Travis to merge this in.

danielturek:
@paciorek I pushed one commit, to update the samplerAssignmentRules system with these changes.  If travis passes, then LGTM.

PR #599, opened 2017-09-19
more efficient dcat-ddirch conjugacy

paciorek:
This creates 'ncat=length(prob)' as altParam for dcat so that ncat can be used instead of length(prob) in dcat-ddirch conjugate sampler. This should save some copying of the prob values in the conjugate sampler code, though as the code stands in the PR we will still dynamically calculate length(prob) when getParam(..., 'ncat') is called. 

To hard-code in the actual length(prob), I think I could intervene in buildNodeFunctions just after:
```
     altParams <- BUGSdecl$altParamExprs
     altParams <- lapply(altParams, nimble:::insertSingleIndexBrackets, modelDef$varInfo)
```
and simply do 'altParams$ncat <- eval(altParams$ncat)'.  However, before I do that, could either @perrydv or @danielturek fill me in on what the 'insertSingleIndexBrackets' is doing here?


danielturek:
@perrydv can certainly comment with more authority, but per my understanding, *all* model variables are internally stored as vectors, or arrays with dimension >= 2.  Meaning, there's nothing (in a model) that's actually a scalar.  Accordingly, when we have 
```
a <- ...
b ~ dxxx(...)
```
in model code, both `a` and `b` will eventually be represented as length-1 vectors in the underlying C++.  So for consistency, we alter the model code, so that it's:
```
a[1] <- ...
b[1] ~ dxxx(...)
```
to agree with this from the beginning of the model processing stages.  This holds for any appearances of `a`, `b`, and any such (originally declared) scalar variables.  So we need to process all the model `code` to insert `[1]` after such variable names.  This happens a lot of places, including in the declaration and processing of `altParam` code.

Not sure I'm totally following your suggestion about "intervening in the buildNodeFunctions function", but I would be extremely hesitant to insert anything in there that's specific to the processing of `dcat` distribution.  Right?

Also, this is merely a suggestion on your PR.  Happy to go with whatever you prefer, but thoughts on re-naming `ncat` to `k`?  Up to you.  I understand `ncat` is for "number of categories", but seems confusingly similar to `dcat`, `rcat`, etc.  `k` seems like a nice variable name to indicate this fixed number.  But like I said, merely a thought, I'm ok either way.

paciorek:
Thanks I had forgotten about the [1] insertion. Clear now.

Yes, I was proposing an "if(dist == 'dcat') {} ' insertion in buildNodeFunctions, and agree it is ugly.

Yes, 'k' is probably better.

paciorek:
This is passing locally on my machine so not sure what is up with Travis.

I changed 'ncat' to 'k' and hoping that the new Travis build will pass tests.

PR #605, opened 2017-09-22
FixIssue529

perrydv:
This PR is a repeat of 1 of 2 PRs that had passed tests and then appeared to break devel a couple of days ago.  That test failure on devel may have been due to a Travis failure, but we immediately reverted to be sure devel is working.  I have now updated this branch from devel and am making a new PR to go through the process again.


perrydv:
The apparent failure of this merge to devel is only because a subsequent merge arrived before this one completed testing.  Apparently that's what it does.

PR #606, opened 2017-09-22
Turn on C++11

perrydv:
This PR is for the second of two recent merges that were reverted because they appeared to break devel despite passing tests prior to merge.  For this one too I have updated from devel and am creating this new PR to initiate a fresh round of testing.


perrydv:
Here goes... this again passed Travis testing.  Let's see if devel breaks again when this is merged, or if last time was a fluke.

PR #607, opened 2017-09-22
Use R6 classes for exprClass

perrydv:
NIMBLE generates a very large number of exprClass objects, each representing one vertex in a syntax tree.

The overhead of creating so many reference class objects is one of many small computational costs contributing to net `compileNimble` time.  

`R6` classes have performance improvements over reference classes.  There is a good case they are more suited to our needs.  For example, they don’t hard-type member data, which we skip for reference classes (via `”ANY”`) anyway because of the unacceptable type-checking overhead.  They are faster to instantiate and use less memory.

On the simple benchmarks in `test-benchmark-building-steps.R`, this change saves about a second each for compiling a tiny model and an MCMC on my Mac.

This PR would add a new package requirement of `R6`, but that does not chain to additional required packages.


paciorek:
LGTM

PR #608, opened 2017-09-22
Do some of `copyFromRobject` in a single new compiled method

perrydv:
One of the big costs of `compileNimble` with thousands of nimbleFunctions (for thousands of MCMC samplers, say) is in populating all the member data  of all the C++ objects from R objects.  This is done by `copyFromRobject`.  This PR contains some steps towards consolidating that work on the C++ side.  It also contains some related fixes. 

C++ for nimbleFunctions gains a `copyFromRobject` method that handles numeric types and nodeFunctionVectors.  The latter are important because they represent the operations of a `calculate`, `simulate`, etc, so they are very common.

I also noticed that the `.Call` used to instantiate new C++ nimbleFunction objects lacked cached `nativeSymbolInfo`, so I added that.  

The new `copyFromRobject` method could work for nimbleLists as well, but currently there is an issue with predefined nimbleLists.  I decided to move forward with a PR for nimbleFunctions and fix the nimbleLists issue as a separate piece of work.

One drawback of this PR is that the net amount of code involved in interfacing between nimble's R objects and C++ objects has increased.  It is hoped that eventually this path will led to removal of some old code as it becomes unnecessary.

All together, the changes speed up compileNimble for benchmarks in `test-benchmark-model-building.R` (with parameters set at large values suggested in the comments) by a second or two on my Mac.


NLMichaud:
This LGTM.  I like the creation and use of the C++ macros to define the `copyFromRObject` methods.  And I agree that even though this creates another pathway for the R to C++ interface for nimbleFunctions, eventually this will make for a cleaner system (not to mention the already present speed increases!)

Issue #516, opened 2017-07-26
Rhub check on CentOS fails because of new compilation flags


paciorek:
Previously I don't think -fstack-protector-strong or -Wdate-time were being injected and now that they are, CentOS gcc is having problems with them.

Here's the 0.6-6 build:
https://builder.r-hub.io/status/original/nimble_0.6-6.tar.gz-4846e2718c13468c909879b318ca4d15
For comparison here is 0.6-5:
https://builder.r-hub.io/status/original/nimble_0.6-5.tar.gz-44433fd76b554bde984c96d94dbbb9c6

paciorek:
Per Duncan's email info, this was caused by issues introduced with trying to avoid the -g flags.

Issue #515, opened 2017-07-26
Rhub check issue on Windows with libnimble_x64.a


paciorek:
On Windows, Rhub check is complaining about not finding /bin/bash and does not create libnimble_x64.a.
Here is the 0.6-6 build result
https://builder.r-hub.io/status/original/nimble_0.6-6.tar.gz-e788db6a60c0457090ec58e1dbdbffa3
and here is the 0.6-5 build result where libnimble_x64.a is created.
https://builder.r-hub.io/status/original/nimble_0.6-5.tar.gz-19a7f6a22e6445e196691adbe8ea9aef

paciorek:
Per Duncan's email info, this was caused by issues with the attempted avoidance of -g flags.

PR #609, opened 2017-09-22
Reduce size of libnimble (2nd try)

perrydv:
Prior to release of 0.6-7, we experimented with a script called from `CppCode/GNUmakefile` that copied the R installation's `Makeconf`,  modified it by removing `-g` flags, and then used it for building libnimble.  This massively reduced the size of libnimble (.a or .dll) and hence the whole package binary.  

Unfortunately the modified `Makeconf` remained in `CppCode` when a binary was built for distribution, which means that R's configure-generated system-specific `Makeconf` from the *build* machine would have persisted in `CppCode` in our binary distribution.  Although this was easy to solve, no one had bandwidth to deal with it at the time, so we did not include any of this in 0.6-7. 

Now I'm making a try at cleaning up the approach.  In each of the two make targets in `GNUmakefile`, the last line now invokes another script to remove the local `Makeconf`.  This prevents it from being included in a binary built for distribution.

@paciorek will you give this a try?


perrydv:
Apparently debugging information *can* reduce performance if the compiler withholds optimizations in order to keep debugging information sane: https://stackoverflow.com/questions/24117063/is-a-program-compiled-with-g-gcc-flag-slower-than-the-same-program-compiled-wit

paciorek:
Was Dirk's suggested approach relevant here? 
http://dirk.eddelbuettel.com/blog/2017/08/14/#009_compact_shared_libraries
http://dirk.eddelbuettel.com/blog/2017/08/20/#010_stripping_shared_libraries

I'm putting on my todo to do various checks on this.

perrydv:
I'm not sure if Dirk's approaches are relevant or not.  Our problem is `libnimble`, which is built in `CppCode` by our config process outside the typical R steps that do compilation in `src`.  We could try the `-Wl,-S` in the Makevars in `CppCode`.  I don't know if that would succeed more universally or if stripping `-g` manually would do so.  I think in a normal R build process one wouldn't want to consider the latter because it would involve a customized set of build steps managed by a `configure` which would be a pain to set up... except we are in that situation anyway.  But by all means feel free to experiment with other solutions.

paciorek:
Just tried rhub. I seem to get the same Windows and CentOS issues as before when we tried this July 26 (issue #516, issue #515).

https://builder.r-hub.io/status/nimble_0.6-7.tar.gz-b842992dc7144d93b67c1738775fe861
https://builder.r-hub.io/status/nimble_0.6-7.tar.gz-6473f13259a645c0aa49e73c61fbc53c

```

paciorek:
One other comment, in the case where we don't strip out -g flags, we are seeing a growth in package size that seems primarily related to libnimble.

0.6-5: 12 MB on my Ubuntu machine
0.6-6: 14 MB
devel: 22 MB

Stripping the -g of course helps a lot, but just wanted to flag this in case worth further investigation.

perrydv:
@paciorek I made one small change that is worth trying.  Instead of replacing "-g" with "", we should replace "-g " with " ", because there could be an option "-g\<more_text\>".

It is possible this will alleviate an unintended corruption of Makeconf.

Beyond that, I don't know what to do about the problems if they are caused by removing "-g".  I am not sure but I think Dirk's suggestion of adding a flag (-Wl,S) is a platform-specific flag, but we need a platform-independent solution.  The "-g " might not be platform-independent, but my hope is that if it appears it means the same thing, and if it doesn't appear, then the stripping operation will do nothing.

Otherwise I'm not sure what to do.  Maybe check for certain common compilers for which we can safely manipulate the Makeconf and otherwise do nothing?  But I hate to start building in compiler-specific behavior.

paciorek:
The errors I saw yesterday seemed to be the same issue as we saw in July - errors resulting from the build machine's Makeconf being used on the machine where the install was being done. But of course avoiding that situation was what your recent fixes were supposed to deal with. So I'm confused.

Could you and @NLMichaud test the build process on a Windows machine? That and CentOS were where the problems were arising.

perrydv:
I see at least one source of why this caused problems on Windows.  There may be more to it, but anyway what I did is disable the removal of "-g" on Windows.  It doesn't seem to make a ridiculously big libnimble_x64 anyway. @NLMichaud can you try installing from this branch on your Windows machine?

paciorek:
I got intrigued so spent a bit of time on the CentOS issue. Somehow a Makeconf is being put into CppCode at the time of R CMD build. So that explains why the rhub problem is happening - it's getting the Makeconf from my Ubuntu machine. 

I just spent a chunk of time figuring out what Duncan told us back in July:
"The call to R CMD build calls the cleanup script. That in turn calls make in inst/CppCode and the GNUmakefile automatically creates the Makeconf.
I believe if inst/CppCode/Makeconf is actually wanted, it should be created at configure time."

Alternatively Duncan's suggestion 7/26/17 was:
`file.copy(packageMakeconfFile, 'Makeconf', overwrite = TRUE)`

So I think we should do one or the other. The idea of creating the Makeconf via customizeMakeConf.R at configure time seems cleaner than passing around a problematic Makeconf with the package tarball and then overwriting. I think we could do it at line 199 of configure.ac. Have to run for now...

perrydv:
Thanks.  I'll try setting it up do be done at configure time.  That point about `R CMD build` invoking `make clean` had slipped past me.

paciorek:
This might reveal my ignorance about configure...

Why don't we just run the customizeMakeConf.R directly from configure rather than invoking Make, which then runs customizeMakeConf.R?

perrydv:
OK, I've implemented a system driven from configure.  One part is yes, as you suggested, running `customizeMakeconf.R` from configure.  I did this via setting up a separate `GNUmakefile_customizeMakeconf` to run this script and then later to remove the local `Makeconf`.  But there is another step, which is that a line in the regular `CppCode/GNUmakefile` needs to include either the local `Makeconf` or R's default `Makeconf`.  So I made a `CppCode/GNUmakefile.in` from which configure generates `CppCode/GNUmakefile` with the proper line.

I added an option `--enable-debugging` to allow regular debugging flags to go through as provide by R.  This is ignored on Windows.

To test it: `R CMD install --configure-args="--enable-debugging" nimble_0.6-7.tar.gz`

I also modified `configure.win` so it will generate `CppCode/GNUmakefile` but never use the customization step.  Actually, that customization step might work on Windows now, but it doesn't seem necessary since debugging information appears to be disabled by default on Windows, at least on mine.

paciorek:
Ok, making progress. After fixing a typo (so please pull the latest on the branch), the CentOS install works via rhub.

On Windows we're still seeing a failure - it fails to build libnimble_x64.a:
https://builder.r-hub.io/status/nimble_0.6-7.tar.gz-9fbd5c64c48e4308af36e54b5ccc6a2a
In contrast under devel, libnimble_x64.a is built:
https://builder.r-hub.io/status/nimble_0.6-7.tar.gz-145db7ab79be41d8a932329a6a4be20c

perrydv:
I'm fairly puzzled by the Windows failure.  The `configure.win` script should render the `GNUmakefile` and subsequent processing essentially identical to what was on devel.  Also I can't reproduce the error.  I did remove one possibly extraneous `/` from a path.  Maybe that mattered.  If it passes tests, do you want to try it on rhub?

paciorek:
Well, the Windows build on r-hub looks ok now. So I think we're ok now.

paciorek:
Well, the Windows build on r-hub looks ok now. So I think we're ok now.

PR #610, opened 2017-09-22
Enforce uniqueness of modelDef names, and hence model interface class names

### LINKED ISSUE #604: bug in assignment into compiled model after second model is compiled
It seems inconceivable that this would not have come up before, but I'm getting an error in the following (this is a simplified version of a Paleon user's code):

```
code1 <- nimbleCode({
  z ~ dnorm(0,1)
})
m1 <- nimbleModel(code1)
cm1 <- compileNimble(m1)
cm1$z <- 4

code2 =nimbleCode({
 y ~ dnorm(0,1)
})
m2 = nimbleModel(code2)
cm2 <- compileNimble(m2)

cm1$z <- 3
## Error in envRefSetField(x, what, refObjectClass(x), selfEnv, value) : 
##  ‘z’ is not a field in class “Ccode”
```

perrydv:
Fixes #604 

You should get a prize for this.  The problem occurs in what will be the 2nd or even 1st line of code executed.  We were not enforcing uniqueness of `modelDef` names.  Those names become part of the names of the generated reference classes to interface to compiled models.  The second reference class over-wrote the definition of the first because it had the same name.

One source of non-uniqueness was that if a `name` argument is not provided, we attempt to get it by the expression provided for the `code` argument, by `deparse (substitute(code))`.  But for historical reasons, `nimbleModel` simply calls `BUGSmodel`, which determines the name.  That means that the call to `BUGSmodel` *always* had simply `code` as the expression of its `code` argument, so deparsing that always yielded the name "code".

I remedied this in two ways:

- I flipped the roles of `nimbleModel` and `BUGSmodel`.  Now `nimbleModel` does the real action, and that means the determination of a name really sees the expression passed for `code`.  I also did some processing like using only the first 10 characters (because there could be `nimbleModel(code = nimbleCode({<LOTS>}))` and replacing spaces with underscores.

- I append a unique label "\_MID\_\<number\>" , created from one of our label generators.  MID stands for "model ID".

### DIFF
--- a/packages/nimble/R/BUGS_readBUGS.R
+++ b/packages/nimble/R/BUGS_readBUGS.R
@@ -13,43 +13,19 @@ BUGSmodel <- function(code,
                       check = getNimbleOption('checkModel'),
                       calculate = TRUE,
                       userEnv = parent.frame()) {
-    if(is.null(name)) name <- deparse(substitute(code))
-    if(length(constants) && sum(names(constants) == ""))
-      stop("BUGSmodel: 'constants' must be a named list")
-    if(length(dimensions) && sum(names(dimensions) == ""))
-      stop("BUGSmodel: 'dimensions' must be a named list")
-    if(length(data) && sum(names(data) == ""))
-        stop("BUGSmodel: 'data' must be a named list")
-    if(any(!sapply(data, function(x) {
-        is.numeric(x) || is.logical(x) })))
-        stop("BUGSmodel: elements of 'data' must be numeric and cannot be data frames")
-    ## constantLengths <- unlist(lapply(constants, length))
-    ## if(any(constantLengths > 1)) {
-    ##     iLong <- which(constantLengths > 1)
-    ##     message(paste0('Constant(s) ', paste0(names(constants)[iLong], sep=" ", collapse = " "), ' are non-scalar and will be handled as inits.'))
-    ##     inits <- c(inits, constants[iLong])
-    ##     constants[iLong] <- NULL
-    ## }
-    md <- modelDefClass$new(name = name)
-    if(nimbleOptions('verbose')) message("defining model...")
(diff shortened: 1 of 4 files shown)


perrydv:
@paciorek The test failure on this branch is from the line `class(cm) == "Ccode"` in `test-dynamicIndexing.R`.  This PR changes the naming scheme to ensure unique names, and it was an accident that compiled model interface classes were all named "Ccode".  I know you were just changing this testing section, so I don't want to stick my finger in it at the same time.  I think a valid replacement would be `inherits(cm, 'CmodelBaseClass')`.

danielturek:
@paciorek @perrydv Better would be to make use of the newly added utility functions: in this case `is.Cmodel()`, and also available are `is.Rmodel()` and `is.model()`.

paciorek:
Great, thanks. I'm fixing the Ccode bit in the cleanup_testing branch.

paciorek:
That said feel free to fix it in this branch if you want to have things pass tests.

perrydv:
Thanks @danielturek.  I made it use `is.Cmodel`, but I also changed how that works from:
```{r}
    return(inherits(obj, 'modelBaseClass') && !inherits(obj, 'RmodelBaseClass'))
```
to 
```{r}
    return(inherits(obj, 'CmodelBaseClass'))
```
Let me know if I'm missing something.

danielturek:
@perrydv Nope, just my oversight.

perrydv:
@paciorek This branch is ready to merge, but it may create a conflict with your `cleanup_testing` branch, since it modifies `test-dynamicIndexing`.  So let me know how you want to proceed (or merge it if you're ready).

paciorek:
Go ahead. If there ends up a conflict it should be simple and I'll deal with it then.

PR #611, opened 2017-09-22
Allow negative numbers as default arguments to work correctly

### LINKED ISSUE #562: Negative numbers not allowed as default values in nimbleFunctions
Default values are allowed for scalar arguments to a nimbleFunction. However, trying to provide a negative default value causes a nimbleFunction to produce an error on compilation.  An example:

```
simpleFunction <- nimbleFunction(
  run = function(outNum = integer(default = -1)){
    returnType(integer(0))
    return(outNum)
  }
)

csimpleFunction <- compileNimble(simpleFunction)
```

perrydv:
Fixes #562 

The problem was that when constructing the R wrapper to the compiled function, the insertion of default values failed for negative values.  This occurred because of use of `as.character` where `deparse` is always safer.  `as.character( quote(-1))` return a vector of "-" and "1" because "-" is handled as a unary negative operator, so the "-1" is parsed as a function call instead of a number.  This meant the code constructed was of the form `function(x = -) {...}`.   `deparse(quote(-1))` returns "-1" as desired and allows construction of `function(x = -1) {...}`.

### DIFF
--- a/packages/nimble/R/cppDefs_RCfunction.R
+++ b/packages/nimble/R/cppDefs_RCfunction.R
@@ -128,7 +128,9 @@ RCfunctionDef <- setRefClass('RCfunctionDef',
                                    argNamesCall = argNames
                                    for(i in seq_along(argNamesCall) ){
                                      if(argsCode[i] != '')
-                                       argNamesCall[i] = paste(argNames[i], " = ", as.character(argsCode[[i]]) )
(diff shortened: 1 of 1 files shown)


perrydv:
The test that failed on travis was fine on my machine.  To check if travis simply crashed, I restarted the tests.

PR #614, opened 2017-09-27
Add a warning message when a user assigns an `RW_block` sampler 

NLMichaud:
This PR causes a message to be issued if a user assigns an `RW_block` sampler to nodes in a model.  The message reads as follows:

```
Note: Assigning an RW_block sampler to nodes with very different scales can result in low MCMC efficiency.  If all nodes assigned to RW_block are not on a similar scale, we recommend providing an informed value for the "propCov" control list argument, or using the AFSS sampler instead.
```

The message is controlled by a new argument to the `addSampler()` method of the `MCMCconf` ref class, named `silent`.  `silent` defaults to `FALSE`, but is set to `TRUE` for `RW_block` samplers that are assigned via `addSampler` in the `initialize` method of `MCMCconf`.

### DIFF
--- a/packages/nimble/R/MCMC_configuration.R
+++ b/packages/nimble/R/MCMC_configuration.R
@@ -297,7 +297,7 @@ Invisibly returns a list of the current sampler configurations, which are sample
                 }
                 thisSamplerName <- if(nameProvided) name else gsub('^sampler_', '', type)   ## removes 'sampler_' from beginning of name, if present
                 if(thisSamplerName == 'RW_block' && !silent){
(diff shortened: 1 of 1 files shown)


danielturek:
@NLMichaud The changes looks great.  Nice way of going about it with the `silent` argument.  Thanks also for updating the sampler assigment rules system.

I pushed two extremely minor changes, to whitespace only.

LGTM.  Once it passes testing, I'd say merge it in.

NLMichaud:
@danielturek Thanks for the quick review!

Issue #457, opened 2017-06-30
Modulo of scalars possibly wrong for negative arguments


fritzo:
I'm seeing flaky test failures in `sapply(testsBasicMath, test_math)`, i.e., failures unless the random seed is set to lucky value.
```
### Testing modulo of scalars ###
function (arg1 = double(0), arg2 = double(0)) 
{
    out <- arg1%%arg2
    return(out)
    returnType(double(0))
} 
clang++ -Qunused-arguments -I/usr/share/R/include -DNDEBUG  -I -DEIGEN_MPL2_ONLY=1 -I"/home/fritz/R/x86_64-pc-linux-gnu-library/3.4/nimble/include" -Wno-misleading-indentation -Wno-ignored-attributes -Wno-deprecated-declarations     -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c P_202_rcFun_202.cpp -o P_202_rcFun_202.o
warning: unknown warning option '-Wno-misleading-indentation'; did you mean '-Wno-missing-declarations'? [-Wunknown-warning-option]
1 warning generated.
warning: unknown warning option '-Wno-misleading-indentation'; did you mean '-Wno-missing-declarations'? [-Wunknown-warning-option]
1 warning generated.
clang++ -Qunused-arguments -shared -L/usr/lib/R/lib -Wl,-Bsymbolic-functions -Wl,-z,relro -o P_202_rcFun_202_06_30_15_39_15.so P_202_rcFun_202.o /tmp/RtmpLkxMpS/nimble_generatedCode/nimble-tensorflow_06_30_15_32_52.so -L/usr/lib/R/lib -lR
 Show Traceback
 
 Rerun with Debug
 Error: Test failed: 'modulo of scalars'
* `out` not equal to `out_nfC`.
1/1 mismatches
[1] -0.106 - 0.114 == -0.22
Test of math (direct R calc vs. C nimbleFunction): modulo of scalars 
```
Note that [another test in `mathTestLists.R`](https://github.com/nimble-dev/nimble/blob/devel/packages/nimble/inst/tests/mathTestLists.R#L55) comments that "probably ... modulo of vecs doesn't work".

perrydv:
Oh yeah.  IIRC, R and C behave differently for modulo with negative arguments.  We could override one or the other but never made a decision to do so.

[...]

PR #616, opened 2017-10-01
Cleanup testing

paciorek:
Not for merging yet, but a placeholder that this should be merged in when I finish with it.

This fixes issue #457 insofar as testing is cleaned up though we still haven't made R and C modulo behavior consistent when negative numbers are involved.


paciorek:
@perrydv when you set up the goldfile system I'm wondering why the temporary file that is compared to the goldfile is not written to tempdir().  It seems to just write to the tests directory in the installed package, which in some cases a user would not have write permission for.

paciorek:
@perrydv @danielturek I'm noticing that in going from 0.6-6 to 0.6-7, the values of jNorm[1] and kLogNorm[1] in the C_samples when running test_mcmc on the 'various conjugacies' model is now giving values like 6e-310. Whereas in 0.6-6 the values in C_samples for those two unused nodes was exactly 0.  It's not a substantive problem of course, but in different runs of the tests the value can change, say from 6e-310 to 6.1e-310, which is causing problems with the goldfile checking. I can put in a workaround, but was wondering if we knew why missing elements of a variable would now no longer have exact zeros in the mvSamples object on the C side.

```
    code <- nimbleCode({
        x ~ dgamma(1, 1)       # should satisfy 'gamma' conjugacy class
        a  ~ dnorm(0, x)     # should satisfy 'norm' conjugacy class
        a2 ~ dnorm(0, tau = 3*x+0)
        b  ~ dpois(0+5*x)
        b2 ~ dpois(1*x*1)
        c ~ dgamma(1, 7*x*5)
        for(i in 2:3) {
            jTau[i] <- 1
            jNorm[i] ~ dnorm(c * (a+3) - i, var = jTau[i])
            kTauSd[i] <- 2
            kLogNorm[i] ~ dlnorm(0 - a - 6*i, kTauSd[i])
        }
    })
```

Example output (the 6.907096e-310 values used to be 0).
```
Browse[2]> head(C_samples)
                a          a2  b b2           c      jNorm[1]   jNorm[2]
[1,]  0.586748822 -0.23209301 14  0 0.010341199            NA -2.8548298
[2,]  0.851879592  0.36189684  8  2 0.010341199 6.907096e-310 -2.6511208
[3,]  0.103909928 -0.24787409  7  2 0.003846483 6.907096e-310 -0.9959005
```

danielturek:
@paciorek @perrydv Good question.  The only comment I can really add is that I'm pretty sure the cause of this isn't on the MCMC side of things, but rather on the C++ / instantiation side of things.

perrydv:
Sorry to be slow picking up on this.  Maybe it is moot for now.  

I'm not thinking immediately of what could have caused that change, but I could look into it.  But first: could it be a between-platform difference?  Or did you verify on *one machine* that there is different behavior from 0.6-6 to 0.6-7 (devel) in that "0" value?

paciorek:
yeah, this contrast was based on a single machine. In fact, I think repeated runs of the same code in 0.6-7 gave slightly different variations on 1e-310. I modified the testing to avoid having this cause a problem so let's just leave this as a curiosity for now.

perrydv:
See Pull Request #634.

PR #619, opened 2017-10-04
Add cross-validation

NLMichaud:
This PR adds the experimental function `runCrossValidate`, which can run k-fold cross-validation on a `nimbleModel`.  Additional information about the use of the function can be found in the Roxygen entry, along with an example.  More examples exist in the tests-crossVal.R testing file.


perrydv:
Following are my suggestions:

Type: "preictive" should be "predictive"

Revised entry:
```
@param returnSamples a logical argument indicating whether to return all 
 #' posterior samples from all MCMC runs.  This can result in a very large returned
 #'  object  (there will be \code{k} sets of
 #'  posterior samples returned).  Defaults to \code{FALSE}.
```

Replace "Mac and Linux" with "OS X (Mac) and Linux".

Revised phrasing suggestions in Details:
"...MCMC is run to estimate the posterior distribution and simultaneously impute posterior predictive values for the left-out data.  Then, the posterior predictive values are compared..." 

"...to produce a single out-of-sample loss estimate."

"If you don't wish to use \ NIMBLE's built in "MSE" or "predictive" loss functions..." (Also, is that a stray backslash?)

It would be great to use the `dyesModel` to construct the data node names used in the `dyesFoldFunction`, to avoid pasting together node names.  Example concept (not tested):

```
makeDyesFoldFunction <- function(model) {
 dataNodes <- model$getNodeNames(dataOnly = TRUE) 
 ## expandNodeNames could also be useful
 dyesFoldFunction <- function(i) dataNodes[i]
 dyesFoldFunction
}
```

We could include both versions for clarity.

NLMichaud:
Thanks a bunch for looking over this, and for catching my typos @perrydv! I really appreciate it. I've made the suggested fixes and alterations to the help text.  

One question about the new `makeDyesFoldFunction()` example: it seems like the function returned from `makeDyesFoldFunction()` will conduct leave-one-out cross validation, i.e. the equivalent of setting `foldFunction = 'random'` and `k = length(model$getNodeNames(dataOnly = TRUE))`  in the arguments to `runCrossValidate()`.  Could it be confusing for users to see this example function that replicates the default choice of fold function in `runCrossValidate()`?

perrydv:
Oh, I think I see your point.  (But I don't think it would be "random", would it?)

The main point of my quick example was to convey the idea of using model methods like "getNodeNames" as a way to avoid pasting together names.  That idea could be used for the row grouping of the actual example.  But we could stick with the pasting version since that conveys that there is nothing magical about node names: they are just character strings.  Maybe some users would find that more direct.

PR #620, opened 2017-10-04
Add warning message for discrete top-level params in MCEM.

NLMichaud:
This PR adds a warning message if a user sets up an MCEM algorithm on a model that has discrete top-level parameters.  The message reads:

```
Warning: MCEM cannot optimize over discrete top-level parameters. The following top-level parameters in your model are discrete: ... (parameter names here)
```


paciorek:
I thought we decided it would be an error not a warning?

NLMichaud:
Whoops, thanks @paciorek.  The warning has been changed to an error.

PR #622, opened 2017-10-12
New message for uninitialized variables in nimbleModel().

NLMichaud:
This PR replaces the former warning message output by `nimbleModel()` for non-initialized variables, which looked like:
```
note that missing values (NAs) or non-finite values were found in model variables: x. This is not an error, but some or all variables may need to be initialized for certain algorithms to operate properly.
```
with a new warning message that reads:
```
This model is not fully initialized.  To see which variables are not initialized, use model$initializeInfo().  This is not an error.  For more information on model initialization, see help(modelInitialization).
```

In addition, there is a new `initializeInfo()` method for `nimbleModel` objects that returns the names of all uninitialized  variables, in a message such as:
```
Missing values (NAs) or non-finite values were found in model variables: x. This is not an error, but some or all variables may need to be initialized for certain algorithms to operate properly. For more information on model initialization, see help(modelInitialization).
```

Finally, if a user calls `help(modelInitialization)`, they will get an informative R help entry with some guidelines on the effect that non-initialized variables can have in different circumstances.

### DIFF
--- a/packages/nimble/NEWS
+++ b/packages/nimble/NEWS
@@ -9,6 +9,8 @@ USER LEVEL CHANGES
 
 -- Addition of dcar_proper distributino, the proper Gaussian conditional autoregressive (CAR) distribution, and MCMC sampling support.
 
+-- nimbleModel no longer outputs the names of all uninitialized variables in a model.  Instead, users are directed to use the new $initializeInfo() method if uninitialized variables are detected. 
+
 BUG FIXES
 
 -- Fixed MCMC autoBlocking procedure, to work with new sampler default values system.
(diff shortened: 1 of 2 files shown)


perrydv:
I propose the following small changes:

- In the new warning message, make "This is not an error." be the second sentence.
- In `help(modelInitialization)`:

    - "effect performance" should be "affect performance"
    - Current work on initialization might make the first bullet point (about initializing dynamic indices) moot.
    - replace "convergence issues" with "slow convergence".

I suggest adding a `\link{\code{initializeModel}}` to the `@seealso` entry of the `nimbleModel` roxygen entry.

NLMichaud:
Thanks for looking at this @perrydv, I made the changes to the help text, and removed the bullet point referencing initialization for dynamic indices.

PR #626, opened 2017-10-19
test travis with branch that passes R CMD check

paciorek:
For purpose of testing on Travis only.


paciorek:
This is now all set for merging and will do so without review in order to forge ahead on steps for release.

Issue #625, opened 2017-10-14
reconsider first calculate(model) in my_initializeModel$run in MCMC


paciorek:
calculate(model) in this context fails when a dynamic index is not initialized.

Per @danielturek comment by email:
That first calculate call I believe is only necessary to propagate things downwards *from* RHS-only nodes (truly, the top level nodes) downwards through any "top level deterministic" nodes, and finally *into* the top-level stochastic nodes.  If that truly is the only use, then yes, the full calculate(model) call could be relaxed -- but I'll admit I could be forgetting some other reason.  But looking at the code, that seems to be the case.

Here's a possibility to consider. Happy to discuss here or perhaps even better in our next meeting.

1) call calculate only on topOnly=TRUE, determOnly=TRUE nodes. 
2) then initiate the current loop in my_initializeModel that loops over the stoch nodes and samples from their priors and then calcs and deps of the stoch node

alternatively we could find the RHSonly nodes and call calculate on their dependencies. I don't think we have any ability currently get only the RHSonly nodes.

danielturek:
Ideas seem reasonable, just one correction, we do have the ability to get RHSonly nodes, take a look at current `initializeModel`, that's a core part of what it does.

paciorek:
ah... It's not user-facing so I didn't realize we had that in the getMaps(). So looks like the (simple) change to the initialization would be to replace calculate(model) with:
```
calculate(model, model$getDependencies(model$getMaps('nodeNamesRHSonly'), determOnly = TRUE)
```

danielturek:
Sounds reasonable.  Let's discuss on Wednesday.

paciorek:
addressed by PR #627

PR #627, opened 2017-10-21
only call calc on deterministic deps of RHS-only nodes at MCMC initialization

paciorek:
This seeks to address issue #625.

I will not proceed until thumbs up from @perrydv and @danielturek .

### DIFF
--- a/packages/nimble/R/initializeModel.R
+++ b/packages/nimble/R/initializeModel.R
@@ -47,11 +47,10 @@ initializeModel <- nimbleFunction(
         ##allDetermNodes <- model$getNodeNames(determOnly = TRUE)
         ##determNodesNodeFxnVector <- nodeFunctionVector(model = model, nodeNames = allDetermNodes)
         determDepsOfRHSonly <- model$getDependencies(model$getMaps('nodeNamesRHSonly'), determOnly = TRUE)
-        topDetermNodesNodeFxnVector <- nodeFunctionVector(model = model, nodeNames = determDepsOfRHSonly)
     },
     
     run = function() {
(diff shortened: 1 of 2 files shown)


danielturek:
Two comments:

1.  Could the same changes be made in a more transparent manner, as:

Rather than creating a nodeFunctionVector, only create this (as currently done):
```
determDepsOfRHSonly <- model$getDependencies(model$getMaps('nodeNamesRHSonly'), determOnly = TRUE)
```

Then change the later `calcuate` line to:
```
model$calculate(determDepsOfRHSonly)
```

Sorry if I'm overlooking something, and that's not the same as what you have.


2. Still thinking about this change, overall.

danielturek:
@paciorek Also suggest this PR be renamed to refer to "model initialization", rather than "MCMC initialization", if that's possible.

As a historical note, the current `initializeModel` originated from code at the beginning of the MCMC algorithm, which was then pulled out as a separate `initializeModel` routine, in hopes of other algorithms using it, as well.

paciorek:
I used the nodeFunctionVector because that's what you had!  I wasn't really understanding why we had explicitly created a nodeFunctionVector before....

danielturek:
I pushed one commit to this PR, to simplify as discussed: no longer using a nodeFunctionVector.

I'll say LGTM, in the sense that I can't come up with cases where this doesn't work.  I'd say go with it, with the deeply-rooted belief that if we're overlooking some case where it doesn't work, we'll find out soon enough...

paciorek:
@paciorek - remember to add basic test of initialization of model with NA in dynamic indexes.

PR #634, opened 2017-10-30
Make NimArrBase::setLength obey fillZeroes even on first allocation.

perrydv:
This PR fixes the issue noted in Pull Request #616, that samples returned for an *unused* node seemed to change behavior across nimble versions.  I could not reproduce that behavior exactly, but I could see that the `NimArr`s used to store the MCMC samples (in a `VecNimArr`) were never being initialized upon memory allocation.  That is ok in normal usage because they will be populated by the MCMC.  And in the past we determined that unnecessary initialization can be a large computational cost in other places (repeatedly called `nimbleFunction`s).  But in this particular step the cost shouldn't be much, and initialization to 0 will create reasonable behavior for the samples of a node that is never used and therefore never even saved as part of the MCMC.

Should it initialize to NA instead?  I'm not sure how to do that but could probably figure out.

### DIFF
--- a/packages/nimble/R/cppDefs_modelValues.R
+++ b/packages/nimble/R/cppDefs_modelValues.R
@@ -67,7 +67,9 @@ cppModelValuesClass <- setRefClass('cppModelValuesClass',
                                            codeLines2 <- list()
                                            if(is.list(vars)) {
                                                for(v in names(Rnames2CppNames)) {
-                                                   ssc <- as.call(c(list(as.name('setSize')), as.list(vars[[v]])))
+                                                   ssc <- as.call(c(list(as.name('initialize')),
+                                                                    c(list(quote(NA_REAL), 1)),
+                                                                    as.list(vars[[v]])))
                                                    codeLines2[[v]] <- codeSubstitute(template2, list(SETSIZECALL = ssc, VAR = as.name(Rnames2CppNames[[v]])))
                                                    
                                                }
@@ -76,7 +78,9 @@ cppModelValuesClass <- setRefClass('cppModelValuesClass',
                                                for(v in names(Rnames2CppNames)) {
                                                    thisSize <- vars$getSymbolObject(v)$size
                                                    if(length(thisSize)==0) thisSize <- 1
-                                                   ssc <- as.call(c(list(as.name('setSize')), as.list(thisSize)))
+                                                   ssc <- as.call(c(list(as.name('initialize')),
+                                                                    c(list(quote(NA_REAL), 1)),
+                                                                    as.list(thisSize)))
                                                    codeLines2[[v]] <- codeSubstitute(template2, list(SETSIZECALL = ssc, VAR = as.name(Rnames2CppNames[[v]])))
                                                }
                                            }
--- a/packages/nimble/R/cppDefs_modelValues.R
+++ b/packages/nimble/R/cppDefs_modelValues.R
@@ -67,9 +67,7 @@ cppModelValuesClass <- setRefClass('cppModelValuesClass',
                                            codeLines2 <- list()
                                            if(is.list(vars)) {
                                                for(v in names(Rnames2CppNames)) {
-                                                   ssc <- as.call(c(list(as.name('initialize')),
-                                                                    c(list(quote(NA_REAL), 1)),
-                                                                    as.list(vars[[v]])))
+                                                   ssc <- as.call(c(list(as.name('setSize')), as.list(vars[[v]])))
                                                    codeLines2[[v]] <- codeSubstitute(template2, list(SETSIZECALL = ssc, VAR = as.name(Rnames2CppNames[[v]])))
                                                    
                                                }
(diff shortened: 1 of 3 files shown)


paciorek:
what is 'NA' on the C side? I thought those ended up being values like 1e-310 when brought back to R.

perrydv:
It is created in R's source code [here](https://github.com/wch/r-source/blob/2f116e3cebf6210b6d70f3e6fff88b2252b345cb/src/main/arithmetic.c#L108).

It is defined in R's API as `NA_REAL`, so we could use it.

paciorek:
I suppose NA is more accurate but not sure it matters that much. And I guess that what will end up in mvSamples will just be copies of whatever we put into the actual model node, so we either will end up with a column of NAs or a column of 0s.

perrydv:
I don't think it is ever actually copied from the model to the mvSamples, since the node is never used.  So I think it is really a question of initialization.

perrydv:
I have made modelValues be initialized with NA instead of 0.

paciorek:
I looked at the test-mcem failure (the other failure seems to be the same tf one that Nick's PR ran into).

This seems insane, but in the cmcmc_Latent$run that occurs just before the 4th call to optim() in the first test in test-mcem, all of the mvSamples end up being NAs. (The first printout of colMeans here is before the $run occurs and the second afterwards. Note that this is after cmcmc_Latent$run has already been run previously and has been fine. I've also verified on a branch from a week ago, that the NAs do not appear.

Given the code change was to initialize to NAs, it raises suspicion that that is the cause, but I have no idea why it would affect this particular algorithm or at the time in the algorithm it does.

Leaving this for now, but any thoughts welcome, @NLMichaud , @perrydv , @danielturek .

```
debug: cmcmc_Latent$run(mAdd, reset = FALSE)
Browse[2]> colMeans(as.matrix(cmcmc_Latent$mvSamples))
              alpha                beta           lambda[1]           lambda[2] 
         0.82805612          1.22757400          5.72581696          1.64335859 
          lambda[3]           lambda[4]           lambda[5]           lambda[6] 
         5.75441919         14.78572511          3.12449809         19.21159695 
          lambda[7]           lambda[8]           lambda[9]          lambda[10] 
         0.83501399          0.83094340          3.01200406         20.46377444 
lifted_d1_over_beta            theta[1]            theta[2]            theta[3] 
         0.81461484          0.06071916          0.10467252          0.09148520 
           theta[4]            theta[5]            theta[6]            theta[7] 
         0.11734702          0.59627826          0.61183430          0.79525142 
           theta[8]            theta[9]           theta[10]                x[1] 
         0.79137467          1.43428765          1.94893090          5.00000000 
               x[2]                x[3]                x[4]                x[5] 
         1.00000000          5.00000000         14.00000000          3.00000000 
               x[6]                x[7]                x[8]                x[9] 
        19.00000000          1.00000000          1.00000000          4.00000000 
              x[10] 
        22.00000000 
Browse[2]> cModel$theta
 [1] 0.06725415 0.02814382 0.04730181 0.13109494 0.09338485 0.58276061
 [7] 0.40888842 0.39499953 0.96254356 1.45093100
Browse[2]> cModel$beta
[1] 1.263885
Browse[2]> cModel$alpha
[1] 0.8252096
Browse[2]> n
debug: m <- m + mAdd
Browse[2]> colMeans(as.matrix(cmcmc_Latent$mvSamples))
              alpha                beta           lambda[1]           lambda[2] 
                 NA                  NA                  NA                  NA 
          lambda[3]           lambda[4]           lambda[5]           lambda[6] 
                 NA                  NA                  NA                  NA 
          lambda[7]           lambda[8]           lambda[9]          lambda[10] 
                 NA                  NA                  NA                  NA 
lifted_d1_over_beta            theta[1]            theta[2]            theta[3] 
                 NA                  NA                  NA                  NA 
           theta[4]            theta[5]            theta[6]            theta[7] 
                 NA                  NA                  NA                  NA 
           theta[8]            theta[9]           theta[10]                x[1] 
                 NA                  NA                  NA                  NA 
               x[2]                x[3]                x[4]                x[5] 
                 NA                  NA                  NA                  NA 
               x[6]                x[7]                x[8]                x[9] 
                 NA                  NA                  NA                  NA 
              x[10] 
                 NA 
```

paciorek:
Ok, some clarity here as I look at this a bit more:

There is a call to do:
```
cmcmc_Latent$run(m, reset = TRUE)
```
where m is 1000. This properly fills the mvSamples with 1000 MCMC samples.

Then after the next call to optim(), there is a call to:
```
cmcmc_Latent$run(mAdd, reset = FALSE)
```
where mAdd is 350. At this point 350 MCMC samples are correctly added into mvSamples, but the initial 1000 samples have NAs in place of the previous values.

A bit more debuggging shows that the resize call at line 114 of MCMC_build.R is now putting in NAs where formerly there were values.

@perrydv can I hand this over to you to figure out why the change to use NAs for initialization is overwriting the first 1000 elements as well as putting NAs in the 350 new elements?

paciorek:
One other note -- if I call resize(CmodelValues) from R, it doesn't seem to introduce NAs into the elements already occupied.

perrydv:
Sorry about that.  I accidentally clobbered behavior of other cases of resizing a modelValues.  The simplest solution for an immediate fix is to go back to initialization to 0 instead of NA, because that can be achieved with the behavior of resize without further changes.  So that's what I've done.

PR #635, opened 2017-10-30
CV updates

NLMichaud:
This PR has a few updates to the CV algo:

- the names of the elements of the `MCMCcontrol` list argument have been changed to match the arguments to `runMCMC` (specifically, they are now `niter` and `nburnin`).  

- The `nburnin` argument will default to 10% of the value of `niter`.

- The fold-specific MCMCs will now start at parameter values obtained from a posterior sample from an initial, whole-model MCMC run.

- A bug where a matrix would sometimes be turned into a vector if a scalar data element was held out was fixed.

### DIFF
--- a/packages/nimble/R/crossValidation.R
+++ b/packages/nimble/R/crossValidation.R
@@ -10,7 +10,7 @@ calcCrossValSD <- function(MCMCout, sampNum, nBootReps, saveData, lossFunction, 
     chunks <- c(apply(indexes, 1, function(x){
       seq(x[1], x[2], 1)}
     ))
-    blockcvValues[r] <- mean(apply(MCMCout[chunks,], 1, lossFunction, saveData))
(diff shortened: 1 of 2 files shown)


paciorek:
at risk of contradicting myself - is the initial whole-model run a new computation that we weren't doing before?  I guess the tradeoff is doing that extra computation vs. less burnin across multiple fold-specific chains...

NLMichaud:
@paciorek That's correct, the whole-model run is a newly added computation that previously wasn't included.  It would also be possible to run e.g. the MCMC for the first fold, and use the posterior parameter values from that run as initial values for future runs.  This would remove the whole-model MCMC run, so maybe that's the way to go, if we're pretty confident that the posteriors won't vary hugely from fold to fold.

paciorek:
OTOH, parallelization complicates things here in terms of doing any one run in advance not fully exploiting parallelization. Arggh...

PR #636, opened 2017-10-30
Allow a combined WAIC to be calculated from multiple MCMC chains 

NLMichaud:
This PR changes runMCMC() to return a single WAIC value, even if multiple chains have been run.  This is accomplished by a new, non-user facing function named `calculateWAICfomSamplesList`, which calculates the WAIC given a list of posterior samples, where each list element is a matrix of posterior samples from a single MCMC chain.

### DIFF
--- a/packages/nimble/R/MCMC_run.R
+++ b/packages/nimble/R/MCMC_run.R
@@ -15,7 +15,7 @@ samplesSummary <- function(samples) {
 #'
 #' Takes as input an MCMC algorithm (ideally a compiled one for speed)
 #' and runs the MCMC with one or more chains, any returns any combination
-#' of posterior samples, posterior summary statistics, and WAIC values.
+#' of posterior samples, posterior summary statistics, and a WAIC value.
 #'
 #' @param mcmc A NIMBLE MCMC algorithm.  See details.
 #'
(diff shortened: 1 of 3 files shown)


danielturek:
@NLMichaud Thanks for jumping on this, and doing an excellent job with the implementation!

I've just pushed stylistic changes (only) to the changes in `MCMC_run.R`.

@paciorek @perrydv @NLMichaud I'll raise one thing I'm not entirely happy about with this change.  Now, when someone uses `nimbleMCMC()` for multiple chains and calculating a WAIC value, it triggers **two** invocations of `compileNimble()`.  The first occurs in `nimbleMCMC()`, to compile the model and the MCMC.  The second occurs in `runMCMC()`, to compile the new function for calculating multi-chain WAIC values.

Related, calls to `runMCMC()` of the form `runMCMC(compiled_mcmc_object, nchains=3, WAIC=TRUE)`, which didn't previously trigger any compilation, now trigger a necessary call to `compileNimble()`.

I'm not sure I see any way around this... but I wanted to raise the issue.  The one the really bothers me is the **two** invocations of `compileNimble()`, when using `nimbleMCMC()`.

Thoughts?  @NLMichaud any ideas for avoiding this?  It's not obvious to me, with the current implementation.

Otherwise, I'm happy with this PR as-is.

NLMichaud:
Thanks for the corrections and the feedback @danielturek!  I agree that the double `compileNimble()` invocation is annoying.  I've attempted to remedy this by changing the method of calculation for the combined WAIC, so that the combined set of posterior samples (from all chains) are copied into the `mvSamples` of the `mcmc` object.  Then, calling `mcmc$calculateWAIC()` will calculate the combined WAIC as normal, removing both the need for a second call to `compileNimble()` and the need for a duplicate, almost identical WAIC calculation function.  

@perrydv To copy from an R matrix of posterior samples into `mcmc$mvSamples`, I used the `matrix2mv()` function in `types_util.R`, as you recommended.  I found it necessary to change a line of code in the `Cmatrix2mvOneVar` function, so that the `.Call()` uses `nimbleUserNamespace$sessionSpecificDll$matrix2VecNimArr` instead of `mv$dll$matrix2VecNimArr`.  Could you check if this is a correct modification to make?

Thanks!

packages/nimble/R/MCMC_run.R
```diff
@@ -128,7 +124,19 @@ runMCMC <- function(mcmc,
...
+      if(nchains > 1){
+        samplesPerChain <- dim(samplesList[[1]])[1]
+        posteriorSamplesMatrix <- matrix(0, nrow = samplesPerChain*nchains, ncol = dim(samplesList[[1]])[2])
+        for(i in seq_along(samplesList)){
+          posteriorSamplesMatrix[((i-1)*samplesPerChain + 1):(i*samplesPerChain),] <- samplesList[[i]][,]
+        }
+        colnames(posteriorSamplesMatrix) <- colnames(samplesList[[1]])
+        matrix2mv(posteriorSamplesMatrix, mcmc$mvSamples)  ## transfer all posterior samples into mcmc$mvSamples
+        mcmc$calculateWAIC()
+      }
```
danielturek: @NLMichaud Is there an `else` missing here?  I'm just glancing at it quickly, I could be wrong.

NLMichaud: Oops nice catch!  The call to `mcmc$calculateWAIC()` within the `if(nchains > 1)` statement was unnecessary.

danielturek:
@NLMichaud The change looks great, to simplify things.  Very good idea.  No extra calls to `compileNimble()` any longer.

I pushed a few stylistic changes, again.

I'm happy with it, as it.

Issue #631, opened 2017-10-25
use inits for dimensions?


paciorek:
If dimensions not available from any other information, should we take it from 'inits'?

perrydv:
At first glance, taking dimensions from `inits` makes sense.

paciorek:
plan is to do this in assignDimensions() in similar fashion to what we do with constants.

paciorek:
This is resolved via PR #643

PR #643, opened 2017-11-25
fill in dimensions with dims from inits

paciorek:
For testing only; do not merge.

This adds dims from inits in similar fashion to use of dims from constants when dimensions input is missing information.

This addresses issue #631


paciorek:
Using dims from inits is affecting test-size.R (not surprisingly), so I need to look into that.

paciorek:
@perrydv @danielturek: thoughts on the following?

If we set values in dimensionsList based on inits, we can run into cases where the inits have inconsistent dimensionality compared to the BUGS code and for which in the past NIMBLE was able to forge ahead, but now we would emit an error. For example in the following case, one can initialize a 1x1 matrix with a scalar and things work fine, but with the changes I am making to grab dimensions from inits, this errors out.

```
code <- nimbleCode({
    for(i in 1:1)
        for(j in 1:1)
           y[i,j] ~ dnorm(mu[i,j], sd = sig)})
m  = nimbleModel(code, debug = T, dimensions = list(mu = 0))

Error in addMissingIndexingRecurse(code[[i]], dimensionsList) : 
  inconsistent dimensionality provided for node 'mu'
```

danielturek:
@paciorek Tough call.  We're always fighting against this ambiguity in demensions in R.  scalar or length-1 vector?  In your example, scalar or `1 x 1` matrix?  I don't know how much work you've put into this system for inferring dimensions from `inits`, but would it be within reason to have it recognize (and track) these inconclusive cases?  That a scalar could be an initial value for a length-1 vector, or for a matrix?  And that a vector could be an initial value for an `n x 1` matrix, or for that matter, for a higher order dimension array?  It might be a bit of work, but could be an answer to this.

paciorek:
I've fixed this up in a form that I'm satisfied with - basically the use of inits to find dimensions now only does so if the inits come in as non-scalars (I had overlooked this in the original coding up of this functionality).  So we can initialize 1-length vectors or matrices with scalars, but if we initialize scalars with 1-length matrices, we trigger an error, which I think makes sense. 

Once I get testing to pass, I will merge this in.

paciorek:
This is set to go but given recent nimble-users issue (see example below), we could also allow dims to be provided via 'data' implicit dimensions, which will handle a case where data specifies a RHS-only variable that is stochastically-indexed (e.g., the code below with 'habmat' provided as 'data'.

Also we should probably trap as error at model definition time when there is stochastic indexing of constants, e.g., the code below with 'habmat' provided as constants.

```
birdcode2<-nimbleCode({
lam0~dunif(0,5)
sigma~dunif(0, 100)
psi~dunif(0,1)

for(i in 1:M){
 s[i,1]~dunif(1,Xu)
 s[i,2]~dunif(1,Yu)
 ind1[i]<-trunc(s[i,1]+0.5)
 ind2[i]<-trunc(s[i,2]+0.5)
 pOK[i] <- habmat[ind1[i], ind2[i]] # habitat check --> pOK[i]=1 if s falls within suitable habitat 
 OK[i] ~ dbern(pOK[i]) # OK[i] = 1 if s falls within suitable habitat, 0 if not, the ones trick

for(j in 1:J){
d[i,j]<- pow(pow(s[i,1]-X[j,1],2) + pow(s[i,2]-X[j,2],2),0.5)
y[i,j] ~ dpois(p[i,j])
p[i,j]<- z[i]*lam0*exp(-(d[i,j]*d[i,j])/(2*sigma*sigma))
}
}
N<-sum(z[1:M])
})

m <- nimbleModel(birdcode2,
                 constants = list(habmat = matrix(rbinom(100, size = 1, p = 0.2), nrow = 10),
                                  Xu = 10,
                                  Yu = 10,
                                  M = 5,
                                  J = 15,
                                  X = matrix(rnorm(30), nrow = 15)),
                 data = list(
                             OK = rep(1, 5)),
                 calculate = FALSE)
```

PR #646, opened 2017-12-19
Cross validation fixes

NLMichaud:
This PR fixes two issues related to NIMBLE's experimental k-fold cross-validation, and adds a new feature:

1)  For some models, if the default "random" foldFunction is used, some values of k will cause the algorithm to error out immediately.  This was due to the method used to randomly chose k roughly equally sized folds, and has been fixed.

2)  The algorithm now returns the averaged loss over all k folds, instead of the sum of the loss over all k folds.  Average loss is what is commonly reported in cross validation procedures.

3)  The user can now set the "nBootReps" argument to NA, in which case no estimates of the Monte Carlo error will be calculated.  This can save significant computation time if the user is not interested in the MC error.


danielturek:
@NLMichaud From looking at the blog post, my only suggestion might be turning off the MCMC progress bar by default in the `calcCrossVal` function

PR #656, opened 2018-01-17
Possible fix for rc function arguments being replaced.

### LINKED ISSUE #653: Bug in user-defined functions (with all constant arguments and that return non-scalars) in BUGS code
Reproducible example of failure, and working cases, below:

```
library(nimble)

## Failure

myNF <- nimbleFunction(
    run = function() {
        returnType(double(1))
        x <- c(1,2,3,4,5)
        return(x)
    }
)

myNF()
## [1] 1 2 3 4 5      -- CORRECT

code <- nimbleCode({
    a[1:5] <- myNF()
})

Rmodel <- nimbleModel(code)

Rmodel$calculate()
Rmodel$a
## [1] NA NA NA NA NA     -- incorrect



## This works:

vectorTimesTwo <- nimbleFunction(
    run = function(x = double(1)) {
        returnType(double(1))
        return(2*x)
    }
)

vectorTimesTwo(1:5)
## [1]  2  4  6  8 10

code <- nimbleCode({
    for(i in 1:3) {
        mu[i] ~ dnorm(0, 1)
    }
    mu_times_two[1:3] <- vectorTimesTwo(mu[1:3])
})

Rmodel <- nimbleModel(code, inits = list(mu = 1:3))
Rmodel$mu
## [1] 1 2 3     --  correct

Rmodel$mu_times_two
## [1] 2 4 6     -- correct


## This also works:

myNF <- nimbleFunction(
    run = function(a = double()) {
        returnType(double(1))
        x <- c(1,2,3,4,5)
        return(x)
    }
)


code <- nimbleCode({
    mu ~ dnorm(0, 1)
    a[1:5] <- myNF(mu)
})

Rmodel <- nimbleModel(code, inits = list(mu=0))

Rmodel$calculate()
Rmodel$a

```

NLMichaud:
This pr fixes #653 by preventing the replacement of rc function names that are used in nimbleModels, and that take no arguments that are other nodes. Previously, such function names were being replaced, resulting in incorrect calculate() and simulate() nodeFunctions for nodes that used these rc functions.


NLMichaud:
@perrydv @danielturek @paciorek Would any of you be up for taking a look at the change I made to line 1016 of BUGS_BUGSdecl.R in this PR?  

It seems like what was happening in Daniel's example (#653) is that `myNF()` was being marked as `allContentsReplaceable = TRUE` in the `genReplacementsAndCodeRecurse` function, which was causing `myNF()` to be replaced with `myNF_oP_cP` , which in turn led to an incorrect `simulate()` function being generated.  The check I added prevents replacement for rc functions (such as `myNF()`).  Does the logic of this fix seem correct?

perrydv:
Thanks @NLMichaud .  This all happens at an earlier processing step than my initial guess.  Thanks for tracking it down.  Your fix makes sense.  I looked more at what is going on.  See comment on the Issue #653 : I think the scope is really any user-defined function that returns a non-scalar and happens to have all constant (or replaceable) arguments when used, which will occur if there are no arguments.  It looks like there is a slightly earlier point (10-20 lines above than where you modified) where there is an intervention to prevent replacement of a predefined list of functions (`functionsThatShouldNeverBeReplacedInBUGScode`).  I thought that might be a slightly more natural place to check for the current case, because it is will be handled in the same way.  It is a small nuance to do it at this earlier step but seems slightly simpler.  I used the same logic you put together, just in the different place.  To move this along, I am going to file a separate PR to see if doing it there passes testing as well.

NLMichaud:
Thanks a bunch for checking this out @perrydv.  I agree that the more natural place to check for this case is before the code gets marked as `allContentsReplacable`, so I'll close this PR and we should use the one you proposed.

PR #660, opened 2018-01-25
Make nimbleRcall work uncompiled by inserting new body after creating RCfunction

### LINKED ISSUE #658: make nimbleRcall functions work in uncompiled mode, or give a useful error message.
```library(nimble)

Rdigamma <- nimbleRcall(function(x = double()){}, Rfun = 'digamma',
                        returnType = double())

code <- nimbleCode({
    b ~ dexp(1)
    a <- Rdigamma(b)
})

model <- nimbleModel(code, calculate = FALSE)
```

The `nimbleModel` call would throw an error without the `calculate=FALSE`, because the nimbleRcall won't work uncompiled.

It does compile:
```
cmodel <- compileNimble(model)

cmodel$simulate()
cmodel$b
cmodel$a
digamma(cmodel$b)
```

perrydv:
Fixed #658.

Previously a nimbleRcall would not actually execute in R (but it would compile, which was the main goal).  That wasn't good.

I fixed this by a bit of a trick.  The nimbleRcall involves construction of somewhat arcane RCfunction code for the compiler to handle.  That is the code that is not executable in R.  But after the RCfunction is created, including the nfMethodRC in its closure environment, it is ok to modify the body of the R function that was used as input.  So at the end of nimbleRcall, I change the body of the constructed function to a simple call to the desired R call.  Hence it executes in R.  This seems to work fine.  I updated test-externalCalls.R as well.

### DIFF
--- a/packages/nimble/R/externalCalls.R
+++ b/packages/nimble/R/externalCalls.R
@@ -224,4 +224,13 @@ nimbleRcall <- function(prototype, returnType, Rfun, envir = .GlobalEnv) {
     body(fun) <- as.call(allLines)
     ans <- quote(RCfunction(fun, check = FALSE))
     ans <- eval(ans)
+    ## By calling RCfunction, there is an nfMethodRC with the necessary body
+    ## for subsequent compiler processing.
+    ## Now we can replace the body of the R function with something that
+    ## will actually execute in R.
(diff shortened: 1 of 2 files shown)


paciorek:
@perrydv is the plan to have this in 0.6-9? If so, can we merge it in?

PR #663, opened 2018-01-27
make arguments like x[1:10] avoid copies

perrydv:
This PR is for testing, not immediate merging.

For a long time we've had some cases of "stupid" copies in generated C++ code for arguments that are simple blocks of a non-scalar.  I decided to try to fix that.  It's been a bit of corner in how the system works, which has hindered earlier efforts.  The fix is not very elegant, but I think it works.

In particular, non-scalar arguments are always passed this way in node functions.  Aside from this as a general issue, it could put a damper on the benefits of re-writing models to take advantage of user-defined functions and distributions.  

Performance improvement from removing an unnecessary copy depends on how big the objects are and how much of the total computation involves the unnecessary copy.  It also depends on whether the objects are in contiguous memory or not, since copies may be more efficient if they are.  For example, my informal benchmarks indicate that for `dmnorm`, the log probability calculation swamps the copy cost anyway.

For a case of copying both a non-contiguous vector `(x[2 ,1:N])` and a non-contiguous matrix `matrix(chol_cov[2, 1:N, 1:N])`, the speedup of the copying alone is about 6x for `N = 50` on my mac.  That's good, although even in the slow version the copy cost will often be small on absolute scale (.0045 seconds for 10000 runs, with unnecessary copies).

If this passes tests, there is one further change I want to add.  Then it could make sense to merge because there may be cases where it helps appreciably, even if it many cases it helps but only negligibly.

This is not a priority for the imminent release.


perrydv:
The two failed tests were test-copy and test-externalCalls.  Both worked on my system so I have restarted them to see if they were travis flukes.

PR #665, opened 2018-01-27
added ability to save MCMC history

### LINKED ISSUE #436: provide acceptance rate history for RW samplers?
In preparing a M-H example for the JSM workshop, I realized we don't store the acceptance rate history. This would be useful for pedagogical purposes as well as for user diagnosis of sampler behavior.

@danielturek, @perrydv what do you think about adding a 1-d array into the RW and RW_block samplers that keeps the acceptance rate from each adaptation period for adaptive samplers, while for non-adaptive we simply store the number accepted over the entire sampling period?

Other ideas welcome.

On a related note, nimbleOptions(buildInterfacesForCompiledNestedNimbleFunctions = TRUE) is pretty clunky for a purpose like this. We might revisit how we handle this.

paciorek:
This is hidden behind nimbleOptions(saveMCMChistory). When enabled (and with `nimbleOptions(buildInterfacesForCompiledNestedNimbleFunctions = TRUE)`), users can use 
code like `cmcmc$samplerFunctions[[3]]$getScaleHistory()` to get the vector of scales, and similarly for vector of acceptance rates.

Will merge when passes testing.

This fixes #436 .

### DIFF
--- a/packages/nimble/R/MCMC_samplers.R
+++ b/packages/nimble/R/MCMC_samplers.R
@@ -142,7 +142,11 @@ sampler_RW <- nimbleFunction(
         timesRan      <- 0
         timesAccepted <- 0
         timesAdapted  <- 0
-        ##scaleHistory  <- c(0, 0)   ## scaleHistory
+        scaleHistory  <- c(0, 0)   ## scaleHistory
+        acceptanceHistory  <- c(0, 0)   ## scaleHistory
+        if(nimbleOptions('saveMCMChistory')) {
+            saveMCMChistory <- TRUE
+        } else saveMCMChistory <- FALSE
         optimalAR     <- 0.44
         gamma1        <- 0
         ## checks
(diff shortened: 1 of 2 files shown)


danielturek:
Generally looks good.

I'm fine with this, except why not also add in the (already written) code for the RW_block sampler, to save and access the proposal covariance history.  It's already written, and that's at least as important in the block sampler case.

paciorek:
Yeah, I was worried about the size of what would be saved, which of course really varies depending on the dimension of the parameter vector...

danielturek:
Fair enough about the saved output being large, when `propCov` is a matrix.  But there's nothing efficient about the code for saving the sampler scale history, anyway.  e.g., it resizes the storage vectors on **every** MCMC iteration.

So, personally, I'd still vote to include the `RW_block` code for also saving the proposal covariance information (since the `RW_block` `scale` is more or less useless, without it), with the understanding that turning on this option to save the sampler parameters is inherently inefficient, and you shouldn't do it if speed / efficiency is your goal.

paciorek:
I'm worried about memory use, not computational efficiency, recognizing that the resizing will slow things down. So I'm still inclined not to save the covariance though I realize the scale for the block is not useful.

paciorek:
Maybe a compromise is to only save the covariance if 'd' is "small".

danielturek:
Sounds reasonable.  Like <= 10 ?

paciorek:
Yes, I updated to save propCovHistory if d<=10

PR #667, opened 2018-01-28
Make model[['b']] work when b is a variable, not a node

### LINKED ISSUE #666: Silent failure when compiled NF incorrectly treats non-scalar model variables as scalar when accessed using `[[` and a constant name.
The title says it all. 

There's a subtle and devious error / inconsistancy, and the worst part is it's a silent failure.  Compilation and execution continue without any warnings or errors, but the behaviour is wrong.

The first example shows **correct** syntax for NF accessing model variables in the `run` function, model variables of different dimensions (0, 1, 2).  Thar variables are accessed and printed correctly here.

```
library(nimble)

code <- nimbleCode({
    a ~ dnorm(0, 1)
    for(i in 1:5) {
        b[i] ~ dnorm(0, 1)
        for(j in 1:3)
            c[i,j] ~ dnorm(0, 1)
    }
})
constants <- list()
data <- list()
inits <- list(a = 0, b=rep(1,5), c = array(2, c(5,3)))

Rmodel$calculate()

## [1] -51.79771

nfDef <- nimbleFunction(
    setup = function(model) {
        aNode <- 'a'                ## CORRECT
        bNode <- 'b'
        cNode <- 'c'
    },
    run = function() {
        a <- model[[aNode]]      ## CORRECT
        b <- model[[bNode]]
        c <- model[[cNode]]
        print('a = ', a)
        print('b = ', b)
        print('c = ', c)
    }
)

Rnf <- nfDef(Rmodel)

Cnf <- compileNimble(Rnf, project = Rmodel)

Rnf$run()

## a = 0
## b = 1 1 1 1 1## c =      [,1] [,2] [,3]
## [1,]    2    2    2
## [2,]    2    2    2
## [3,]    2    2    2
## [4,]    2    2    2
## [5,]    2    2    2

Cnf$run()

## a = 0    ## CORRECT OUTPUT
## b = 1    ## FOR COMPILED EXECUTION
## 1
## 1
## 1
## 1
## c = 2 2 2
## 2 2 2
## 2 2 2
## 2 2 2
## 2 2 2
## NULL
```
Now we'll try again, but get the syntax wrong, using `model[["nodeName"]]` in the `run` function.

This executes fine, but when the non-scalar model variables are accessed in the `run` function, they instead return scalars.  
```
code <- nimbleCode({
    a ~ dnorm(0, 1)
    for(i in 1:5) {
        b[i] ~ dnorm(0, 1)
        for(j in 1:3)
            c[i,j] ~ dnorm(0, 1)
    }
})
constants <- list()
data <- list()
inits <- list(a = 0, b=rep(1,5), c = array(2, c(5,3)))

Rmodel$calculate()

nfDef <- nimbleFunction(
    setup = function(model) { },
    run = function() {
        a <- model[['a']]     ## INCORRECT
        b <- model[['b']] 
        c <- model[['c']] 
        print('a = ', a)
        print('b = ', b)
        print('c = ', c)
    }
)

Rnf <- nfDef(Rmodel)

Cnf <- compileNimble(Rnf, project = Rmodel)

Rnf$run()

## a = 0
## b = 1 1 1 1 1
## c =      [,1] [,2] [,3]
## [1,]    2    2    2
## [2,]    2    2    2
## [3,]    2    2    2
## [4,]    2    2    2
## [5,]    2    2    2

Cnf$run()

## a = 0
## b = 1    ## INCORRECT
## c = 2    ## INCORRECT
## NULL
```
Again, this happens silently.

perrydv:
Fixes #666, whose title has been revised to be more precise.

This issue raised the following case:
`model[["b"]]` in a nimbleFunction.

The features are:

1. "b" is hard-coded
2. The variable `b` in `model` is non-scalar

This silently behaved wrong by returning only the first element of `model$b`.  This was more a bug in error-trapping (silent incorrect behavior), testing (this was not covered in testing), and/or documentation than in functionality of intended usage.  If writing a nimbleFunction that will directly access a fixed (will not vary across instances) node name, the intended way to do so was `model$b`.  In model-generic programming, one typically wants `model[[bNode]]` where `bNode` is a character variable with node or vector name, and that works fine. Section 14.4.2 of the manual covers this, but not extremely clearly.

The case where `model$b` is a scalar was accessed correctly by `model[["b"]]`.

In order to trap the case that `b` in `model` is non-scalar, we must look in `model`.  When possible at this "keyword processing" stage, we try whenever possible to determine what is going on by syntax alone to avoid costly looking around the model.   Until this point, we did not look in model if the index of `[[` for a model was a string. However, given that we are going to need to check in the model, then it becomes as easy to make `model[["b"]]` work as it would be to error-trap what is going on just to tell the user not to do it.  So I made it work.

There is a new test file `test-nimbleFunctionModelAccess.R` that illustrates the different ways to access nodes and variables.  It is not elegant (a failure will take work to pin down) but it covers a lot of cases.

I edited section 14.4.2 ("Getting and setting variable and node values") of the manual.

A subtlety (explained in 14.4.2) is that if `b` in `model` is a scalar, then `model[["b"]]` is a scalar but `model$b` is a length-one vector.  There is a rationale to that, and I'm hoping we can live with it.  It has always been the case for `model[[bNode]]` if `bNode == "b"`.

In summary, `model[["b"]]` works in this PR.

This involves steps of processing used quite a lot, so let's see what testing reveals.

### DIFF
--- a/UserManual/chapter_WritingNimbleFunctions.Rnw
+++ b/UserManual/chapter_WritingNimbleFunctions.Rnw
@@ -233,12 +233,13 @@ can only be indexed within a call to \cd{calculate}, \cd{calculateDiff},
 \subsubsection{Using indexing with nodes}
   
 
-Here is an example that illustrates getting and setting of nodes and subsets of nodes.  Note the following:
+Here is an example that illustrates getting and setting of nodes,
+subsets of nodes, or variables.  Note the following:
 
 \begin{itemize}
-  \item In \cd{model[[node]]}, \cd{node} can only be a single node name, not a vector of multiple nodes nor an element of such a vector (\cd{model[[ nodes[i] ]]} does not work).  The node itself may be a vector, matrix or array node.
-  \item In fact, \cd{node} can be a node-name-like character string, even if it is not actually a node in the model.  See example 4 in the code below.
-  \item One can also use \cd{model\$nodeName}, with the caveat that \cd{nodeName} can't be a variable (i.e., it needs to be the actual name of a variable or node) and so would only make sense for a nimbleFunction written for models known to have a specific node.
+  \item In \cd{model[[v]]}, \cd{v} can only be a single node or variable name, not a vector of multiple nodes nor an element of such a vector (\cd{model[[ nodes[i] ]]} does not work).  The node itself may be a vector, matrix or array node.
+  \item In fact, \cd{v} can be a node-name-like character string, even if it is not actually a node in the model.  See example 4 in the code below.
+  \item One can also use \cd{model\$varName}, with the caveat that \cd{varName} must be a variable name.  This usage would only make sense for a nimbleFunction written for models known to have a specific variable.  (Note that if \cd{a} is a scalar node in \cd{model}, then \cd{model[['a']]} will be a scalar but \cd{model\$a} will be a vector of length 1.)
  \item one should use the \cd{<<-} global assignment operator to assign into model nodes. 
 \end{itemize}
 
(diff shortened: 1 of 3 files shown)


danielturek:
@perrydv Thanks for getting right on it.

The difference between `model[["b"]]` and `model$b` is a surprise to me.

Issue #640, opened 2017-11-22
Bug with 3-dimension parameters and getParam() ?


danielturek:
New error that comes up, as best I can tell looks like resulting from having 3-dimensional parameters to custom distribution functions.

Upon model building, this results in an error like:

```
> Rmodel <- nimbleModel(code, constants, data, inits, check = FALSE)
defining model...
Adding prior,condition as data for building model.
building model...
setting data and initial values...
running calculate on model (any error reports that follow may simply reflect missing values in model variables) ... 
checking model sizes and dimensions...Error in envRefInferField(x, what, getClass(class(x)), selfEnv) : 
  'getParam_3D_double' is not a valid field or method name for reference class "y_L64_UID_52"
Error in envRefInferField(x, what, getClass(class(x)), selfEnv) : 
  'getParam_3D_double' is not a valid field or method name for reference class "y_L64_UID_52"
```

Reproducible example:
goto github repository:
danielturek/HMM-MCMC
(any of the "goose" or "orchid" scripts will demonstrate this error).

For example, download:
HMM-MCMC/data/gooseData.RData,
and execute the script:
HMM-MCMC/code/goose_FilterRR.R

Sorry this example is far from "minimal".....

paciorek:
We only provide up to 2D for getParam. This should be straightforward to extend. 

@danielturek What is the use case / what timeline would be helpful on this?

danielturek:
No rush whatsoever.  I think it was only that the legacy HMM-MCMC code no longer works, since it uses custom distributions that have 3D+ parameters.  Some folks were trying to use the code (posted along with the paper), and it gave errors / warnings.  But I'd forgotten about this, it was more a problem for them rather than me, so it's not inconveniencing me.  But, it does sound like something that we should cover, if it's not too difficult.

paciorek:
ok, that's helpful to know - if we have examples out there associated with the paper that now fail, that's bad. I'll see if I have time over weekend to fix up.

paciorek:
Looks like the issue here will be that getParam goes through eigenization, so we get this error even if the calculations with the 3-d param don't actually do any non-allowed math.

```
>     cm <- compileNimble(m)
compiling... this may take a minute. Use 'showCompilerOutput = TRUE' to see C++ compiler details.
Error : Cannot do math with arrays that have more than 2 dimensions.
 This occurred for: PARAMANSWER_
 This was part of the call:  PARAMANSWER_ <- map("model_theta",3,0,list,list)
[...]

PR #668, opened 2018-01-29
allow 3D params but avoid creation of getParam_3D

paciorek:
This addresses issue #640.  

Example from @danielturek  in issue #640 looks fine when run under code in this PR.


paciorek:
This passes tests except the Travis Mac test -- it's failing nimbleList for some reason, but nimbleList seems fine on my own Mac. So I think things are fine.

PR #670, opened 2018-02-01
fix configure.win

perrydv:
I think this fixes the broken windows installation.  @NLMichaud can you test it too?


NLMichaud:
Oops sorry for not trying compilation before, definite oversight on my part.  @perrydv I was able to install successfully, and both MCMC and filtering tests passed.

paciorek:
This was simultaneously resolved elsewhere.

PR #676, opened 2018-03-02
Deprecate identityMatrix in favor of diag

perrydv:
For a long time we have supported R-like use of diag() to create an identity matrix.  I noticed there were three cases of the old identityMatrix still in the code base.  That is more cumbersome because it uses an entirely separate nimbleFunction.  This PR puts identityMatrix in deprecated status by adding a note to its roxygen entry and changing its uses to diag() instead.


perrydv:
For some reason travis tests did not run.  I will re-start later.

danielturek:
Nice catches @perrydv

perrydv:
test-nimbleList failed on travis.  It passed on my machine, so I am restarting travis testing to see if it was a glitch there.

perrydv:
The test passed this time, with no changes to the code.  It is attributed to a travis blip.  I think this PR is ready to merge.

PR #679, opened 2018-03-15
Faster checkForSelfParents

perrydv:
A step if building the `modelDef` is `checkForSelfParents`, which traps cases where a deterministic node depends on itself.  (Cases where a stochastic node depends on itself seem to trigger an error at an earlier step.)  This worked by iterating through each edge in the graph and checking if it points in a circle.  A user case of slow build times for a large model revealed that about 1/3 of `nimbleModel` time was spent on this inefficient iteration.  In this PR, that iteration is replaced with a vectorized check of whether any edges point in a circle (same “from” and “to” nodes).  This is much faster.


paciorek:
LGTM. I'm merging as this seems straightforward.

PR #680, opened 2018-03-15
speed up makeParamInfo and related functions/methods

perrydv:
With some stochastic indexing examples we see cases where `makeParamInfo` takes a huge amount of time while compiling an MCMC.  This PR contains some hasty fixes that reduce the time by perhaps  around 75% in the motivating example (careful before/after timing was not done because it is clearly must faster, and case-specific).  This could be made faster still, but this PR is limited to the quick fixes that were feasible in a short time.

### DIFF
--- a/packages/nimble/R/BUGS_model.R
+++ b/packages/nimble/R/BUGS_model.R
@@ -93,7 +93,10 @@ Details: The return value is logical vector with an element for each node indica
 
                                   ## returns the declaration ID corresponding to nodes
                                   getDeclID = function(nodes) {
-                                      graphIDs <- modelDef$nodeName2GraphIDs(nodes, unique = FALSE)
+                                      graphIDs <- if(is.character(nodes))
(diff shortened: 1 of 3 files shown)


perrydv:
The changes I made yesterday passed testing.

I made additional changes that speed up handling of makeParamInfo for getParam much more.  The information from getParamInfo is really only needed for size processing of getParam, and really only one instance is needed.   I made a delayed evaluation system where actual evaluation is only done on instances until the necessary type information is obtained.  Typically this will be on the first instance.  Any inconsistency of types across instances will result (I think) in a run-time error, which is the same as it has been.

perrydv:
This did not work.  The problem is that one element of the makeParamInfo result *is* needed in all instances, while other information is not needed.  I will try to fix it later.

perrydv:
I reverted this morning's attempt to achieve greater efficiency.  It is not going to work.

We could make this faster but it would take a bit more work.

paciorek:
LGTM. Given a bit of history in this PR thread, I'll let @perrydv merge this.

PR #681, opened 2018-03-17
WAIC fixes

NLMichaud:
This PR adds a check to ensure that, if a user would like to calculate a WAIC value, this value will be valid.  It also fixes a bug where downstream nodes were not simulated during WAIC calculation for certain combinations of models and monitors.  The following features are added:

1. A new argument to `buildMCMC()` named `enableWAIC`, which defaults to `FALSE`.  If a user sets `enableWAIC = TRUE`, pre-processing is executed to calculate the necessary sets of nodes for WAIC calculation, and a check is run to ensure that the set of nodes monitored by the MCMC will lead to a valid WAIC calculation.  If a user sets `enableWAIC = FALSE`, no pre-processing or checking will occur.  

2. There is also a nimble option, likewise named `enableWAIC`, that defaults to `FALSE`.  If `nimbleOptions(enableWAIC)` is set to `TRUE`, then WAIC calculations will be enabled for all calls to `buildMCMC`. 

3. A bug where downstream dependencies were not being simulated for certain combinations of models and monitored nodes has been fixed.

4. The function `checkWAICmonitors` now checks to ensure that the set of nodes monitored by the mcmc algorithm lead to a valid WAIC calculation.  If the calculation is not valid, the following error message is printed:
```
"In order for a valid WAIC calculation, all parameters of data nodes in the model must be monitored, or be downstream from monitored nodes.   See help(buildMCMC) for more information on valid sets of monitored nodes for WAIC calculations.
Currently, the following data nodes have un-monitored upstream parameters: ...
```
The first 10 such data nodes will be printed in the above message to keep the error relatively short.

If the calculation is valid, the following informational message is printed:
```
"Monitored nodes are valid for WAIC.  
```

A few questions for @danielturek, although @perrydv and @paciorek please let me know if you have any thoughts too :

1. I've added the `enableWAIC` argument to `buildMCMC`'s setup function after the current two arguments (`conf` and `...`).  This seemed much preferable to adding the `enableWAIC` argument before `...` (which would break existing code), but since `enableWAIC` comes after `...`, I believe that means that a user must set the argument by name instead of using positional argument matching (since the `...` can take an arbitrary number of arguments).  I think an alternative would be to package `enableWAIC` into `...` if you feel that's preferable.  

2. I changed the example model in the MCMC help documentation slightly (added an additional data node).  The original model had no data nodes, so in fact, the WAIC could not be calculated for it.  Let me know if you'd like me to undo this change or use a different example model. 

Thanks!

### DIFF
--- a/UserManual/chapter_MCMC.Rnw
+++ b/UserManual/chapter_MCMC.Rnw
@@ -581,7 +581,7 @@ Obtaining samples as matrices is most common, but see Section \ref{sec:modelValu
 \section{Calculating WAIC}
 \label{sec:WAIC}
 
-Once an MCMC algorithm has been run, as dscribed in \ref{sec:executing-the-mcmc-algorithm}, the WAIC \citep{Watanabe_2010} can be calculated from the posterior samples produced by the MCMC algorithm.  Note that in order to calculate the WAIC value after running an MCMC algorithm, the argument \cd{enableWAIC = TRUE} must have been supplied to the call to \cd{buildMCMC}, or the \cd{enableWAIC} nimble option must have been set to \cd{TRUE}.
+Once an MCMC algorithm has been run, as described in \ref{sec:executing-the-mcmc-algorithm}, the WAIC \citep{Watanabe_2010} can be calculated from the posterior samples produced by the MCMC algorithm.  Note that in order to calculate the WAIC value after running an MCMC algorithm, the argument \cd{enableWAIC = TRUE} must have been supplied to the call to \cd{buildMCMC}, or the \cd{enableWAIC} NIMBLE option must have been set to \cd{TRUE}.
 
 The WAIC is calculated by calling the member method \cd{mcmc\$calculateWAIC} (see \cd{help(buildMCMC)} in R for more details).  The \cd{calculateWAIC} method has one required argument, \cd{nburnin}, the number of initial samples to discard prior to WAIC calculation.  \cd{nburnin} defaults to 0.
 
--- a/UserManual/chapter_MCMC.Rnw
+++ b/UserManual/chapter_MCMC.Rnw
@@ -18,7 +18,7 @@ require(igraph, warn.conflicts = FALSE, quietly = TRUE)  # same question
 
 NIMBLE provides a variety of paths to creating and executing an MCMC algorithm, which differ greatly in their simplicity of use, and also in the options available and customizability.
 
-The most direct approach to invoking the MCMC engine is using the \cd{nimbleMCMC} function (Section \ref{sec:nimbleMCMC}).  This one-line call creates and executes an MCMC, and provides a wide range of options for controlling the MCMC: specifying monitors, burn-in, and thinning, running multiple MCMC chains with different initial values, and returning posterior samples, summary statistics, and/or WAIC values.  However, this approach is restricted to using NIMBLE's default MCMC algorithm; further customization of, for example, the specific samplers employed, is not possible.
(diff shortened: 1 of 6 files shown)


danielturek:
@NLMichaud Thanks for checking these few things with me.

1.  I would have been tempted to make the ordering of arguments as `buildMCMC(conf, enableWAIC, ...)`, that is, making the new `enableWAIC` argument precede the `...` argument, and expecting on people who use the `...` argument to have provided **named** arguments for `...`.  If they were relying on passing **unnamed** arguments via `...`, and those arguments doing something meaningful only through positional matching when they get passed along to other functions, then I feel like they're not using the `...` quite right.  But I could be wrong here.  Thoughts, @perrydv or @paciorek ?

2. Could I suggest that the default value for the `enableWAIC` argument instead be `nimbleOptions("enableWAIC")`?  Then, in the setup code, only use the value of the `enableWAIC` argument.  That way they could opt into all WAIC calculations using the nimble option, but still turn it off for specific MCMCs by specifying `enableWAIC = FALSE`.  Does that make sense?  What do you think?

3. What, you think a model with no data isn't a good example?  I'm just kidding.  That example has been like that since the first incarnation of the MCMC, it certainly could use an update to include some data and be slightly more realistic.

NLMichaud:
Awesome, thanks for the quick feedback @danielturek.

1.  That's a really good point, the `...` should only be used to provide named arguments, which wouldn't be broken by the inclusion of `enableWAIC` beforehand.  

2.  Also a good idea, and clearer (both in code and function) than my current solution.

3.  :) Thanks! 

I just pushed these changes to the PR.  Here's another little detail I happened upon:

1.  The `runMCMC()` function requires some updates in light of the `enableWAIC` change.  Specifically, it's now possible for a user to set `WAIC = TRUE` as an argument to `runMCMC()`, but also pass an MCMC object to `runMCMC()` that was created with `enableWAIC = FALSE`.  I've accounted for this by checking that `enableWAIC = TRUE` in the `mcmc` argument to `runMCMC()` if the `WAIC` argument to `runMCMC()` was also set to `TRUE`.  If `enableWAIC = FALSE` but `WAIC = TRUE`, the `runMCMC()` function errors out.  I've also updated the help documentation to reflect this.  Likewise, `nimbleMCMC()` has been updated to work with the new `enableWAIC` system.

danielturek:
@NLMichaud Looks great.

I assume you did some reasonable testing of the `checkWAICmonitors` function?

NLMichaud:
Great, thanks @danielturek !

Thanks for checking about testing too.  I believe that I've tested out the cases that we could encounter, but I'm going back and adding a few additional model / monitor combinations to the testing file just to be safe.

NLMichaud:
Hi @danielturek - additional testing (thanks again for the encouragement!) revealed a little quirk in the algorithm that I hadn't considered before.  As it was written, the `checkWAICmonitors()` function required a complete set of _stochastic_ nodes upstream of each data node to be monitored.  This meant that if a user monitored only a set of deterministic nodes that fully determined the parameters for all data nodes, the `checkWAICmonitors()` function would not recognize that as a valid set of monitors, even though using the sampled deterministic nodes in this case would lead to a correct calculation.  This was happening because `model$getDependencies()` continues getting node names downstream of deterministic nodes until a stochastic node is reached, so even if a monitored deterministic node was directly in between a stochastic parameter and a data node in a model, `model$getDependencies()` would still show the data node as depending on the un-monitored stochastic parameter.

I just pushed a PR that fixed this by checking to see which of the monitored nodes were deterministic, calling them e.g. `monitoredDetermNodes`, and then including the argument `omit = monitoredDetermNodes` in the call to `model$getDependencies(S, stochOnly = TRUE, self = FALSE, includeData = TRUE, omit = monitoredDetermNodes)` in the initial algorithm.  This removes all nodes downstream from the monitored deterministic nodes from the get dependencies call, so that monitoring deterministic nodes now registers as valid sets of monitors.

Let me know if this sounds like a good fix, or even a problem that needed to be fixed :)  I believe that   Thanks!

danielturek:
@NLMichaud  I completely follow you're line of reasoning, and agree this could in some situations be a source of incorrect behaviour.  Nice catch.  Yes, I think your suggested fix should handle it fine, using the `omit` argument.  I remember writing the first implementation of `omit` (since been re-written), and I'm not sure how much usage it's gotten, so I'm actually happy to see it come up here.

Also, I forgot to mention before, but thanks for also updating the `nimbleMCMC` function appropriately for these `enableWAIC` changes.  Everything LGTM.

PR #682, opened 2018-03-19
push parse step of parseEvalNumericManyList to C++

perrydv:
This PR moves a costly step in lots of model processing into C++.  On one example model, configureMCMC with useConjugacy = FALSE saves 40% from this change.


perrydv:
This passed testing.  I merged something else into devel and decided to merge devel into this PR branch to initiate new testing.

perrydv:
@paciorek Here is an additional speed improvement that could be included in 0.6-10 if not too late.

Issue #684, opened 2018-03-23
Error-trapping of multivariate expressions defeats some valid models


perrydv:
Reproducible example of a Markov chain where the state of the chain is a stochastic index.  The expression `i-1` in a non-scalar distribution argument triggers error-trapping where it is not necessary.
```{r}
library(nimble)

mc <- nimbleCode({
   x[1] <- 1
   for(i in 2:10) {
       x[2] ~ dcat(transition_probs[x[i-1], 1:2])
   }
}
)

transition_probs <- matrix(c(.1, .9, .4, .6), nrow = 2)

m <- nimbleModel(
    mc,
    data = list(transition_probs = transition_probs)
)
```

paciorek:
closed by PR #685

PR #685, opened 2018-03-23
make checkMultiVarExpr optional

perrydv:
This PR provides a short-term work-around for issue #684.  It simply provides a new nimbleOption, `disallow_multivariate_argument_expressions`.  If this option is set to `FALSE` (it's default is `TRUE`, yielding current behavior), then the step of checking for expressions in multivariate arguments is simply skipped. This will allow users a path to build a model such as the example given in the Issue.

This seems sufficiently minor and does not change regular behavior that it could be squeezed in 0.6-10, but it is also fine if not.


paciorek:
Sure - it passes testing, so if you're confident about it, I will add it in. I'm just now merging the setData df PR.

PR #689, opened 2018-04-03
Add a dtypedef

NLMichaud:
This PR adds the line:
```
typedef Matrix<CppAD::AD<double>, Dynamic, Dynamic> MatrixXd;
```
to all generated `calculate_ADproxyModel` C++ methods, and updates the AD model testing to only check for values and first order derivatives.


NLMichaud:
@perrydv I agree entirely, finding a way to avoid the compiler warnings would be great.  I'll merge this for now.

PR #696, opened 2018-04-16
Fix setInits in complicated cases

### LINKED ISSUE #695: `model$setInits` gives incorrect behavior in complicated cases
`model$setInits` can perform incorrectly when the following conditions are met:

* some nodes in a variable are data and some are not.
* some elements of a variable are not defined as nodes.

Reproducible example:

```{r}
library(nimble)

mc1 <- nimbleCode({
    for(i in 1:2) {
        for(j in f[i]:n) {
            x[i, j] ~ dnorm(0,1)
        }
    }
})
inits = list(x = matrix(1:6, nrow = 2))
data = list(x = matrix(c(100, NA, 100, NA, NA, NA), nrow = 2))
m <- nimbleModel(mc1,
                 constants = list(f = c(1, 2),
                                  n = 3),
                 data = data,
                 inits = inits)
## This model defines x[1, 1], x[1, 2], x[1, 3], x[2, 2], and x[2, 3]. 
## x[2,1] is not a node in the model.
## data are provided for x[1, 1] and x[1, 2]

m$x ## This is wrong
##       [,1] [,2] [,3]
##[1,]   100   3    5
##[2,]   NA    4    NA
matrix(c(100, 2, 100, 4, 5, 6), nrow = 2) ## It should be this
##       [,1] [,2] [,3]
##[1,]  100  100    5
##[2,]    2    4    6
```

It doesn't matter if `inits$x` has NAs where `data$x` has values.

A fix is on the way.

perrydv:
Fixes #695 

`model$setInits` uses `model$isData()` to look up what nodes in a variable are data.  This goes wrong if not all elements in a variable were defined as nodes.  The problem is that `model$isData` returns a logical vector of the wrong length, which is then used according to the recycling rule to index elements of the inits variable.

The proposed solution uses logical variables in `model$isDataEnv` directly.  This should work because setInits always operates for whole variables (skipping any `NA` elements).

### DIFF
--- a/packages/nimble/R/BUGS_model.R
+++ b/packages/nimble/R/BUGS_model.R
@@ -818,7 +818,7 @@ inits: A named list.  The names of list elements must correspond to model variab
                                           if(any(dataVals)) {
                                               .self[[names(inits)[i]]][!dataVals] <- inits[[i]][!dataVals]
                                               if(any(!is.na(inits[[i]][dataVals])))
-                                                  warning("Ignoring elements in inits for data nodes: ", names(inits)[[i]], ".", call. = FALSE)
+                                                  warning("Ignoring some or all values in inits for data nodes: ", names(inits)[[i]], ".", call. = FALSE)
                                           } else  .self[[names(inits)[i]]] <- inits[[i]]
                                       }
                                   },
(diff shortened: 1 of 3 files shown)


perrydv:
A test failure occurred because I updated an error message and the test that checks if that error-trap works was expecting the old version of the message.  I have updated both together.  Tests should re-run.

PR #698, opened 2018-04-16
Silence auto block

### LINKED ISSUE #697: autoBlock is noisy
Since a cautionary message about adding RW_block samplers was added, autoBlock is now very noisy, issuing that warning many times as it reconfigures blocks.  A PR will be submitted to silence these warnings during use of autoBlock.

perrydv:
Fix #697 

This has not been tested.  They are simple code changes so I am going to let travis do the testing.


perrydv:
This test failure is related to the other branch I have under PR.  I must have branched from that one by accident.  No matter, both are very small changes.  I'll fix it.

perrydv:
No, that's not what happened.  But it's another similar case.  I changed a warning message, and the test for whether that warning works as expected looks for a match in the message.  I'll update.

perrydv:
Oh, it was from the other branch.

PR #704, opened 2018-04-21
make AF_slice enforce correct topological order in calculate()

perrydv:
A previous attempt to avoid problems with dynamic indices led to an error in respecting topological ordering of nodes.  This is a short-term fix for devel while we discuss a more permanent fix.


perrydv:
Per e-mail discussion, I am rushing through a merge to devel as a quick fix.  Then we will discuss a more general solution soon.

Issue #703, opened 2018-04-18
Make MCMC `thin` into a runtime argument


danielturek:
Same for `thin2`.

Also handle `thin` being provided at the MCMC configuration stage.

PR #706, opened 2018-04-24
added MCMC functionality: sampler execution order

### LINKED ISSUE #701: Allow same sampler to be used multiple times in one configuration
Sometimes it is useful to have one sampler used more frequently than others.  That seems almost to be achieved by `MCMCconf$setSamplers` if we if one provides the same index multiple times.  However IIUC correctly this would instead result in multiple independent instances of the same type of sampler for the same target, rather than a single instance that would be re-used.  The benefit of re-using the same instance repeatedly would be faster adaptation, as opposed to having multiple instances adapt independently.  I think this came up early on but don't remember why it came out as it is now.

danielturek:
Allows modification of the ordering of sampler execution, including repeated execution of the same sampler function.  This order can be specified as part of the MCMC configuration, or also changed at MCMC runtime.

Also updated documentation, User Manual, and NEWS.

Fixes #701

### DIFF
--- a/UserManual/chapter_MCMC.Rnw
+++ b/UserManual/chapter_MCMC.Rnw
@@ -106,7 +106,7 @@ This entry point for using \cd{nimbleMCMC} is the \cd{code}, \cd{constants}, \cd
 Based on its arguments, \cd{nimbleMCMC} optionally returns any combination of
 
 \begin{itemize}
-\item posterior samples,
+\item Posterior samples,
 \item Posterior summary statistics, and
 \item WAIC value.
 \end{itemize}
@@ -375,6 +375,33 @@ samplerConfList[[4]]$setTarget("y", model)   ## model argument required
 mcmcConf$setSamplers(samplerConfList)
 @
 
+\subsubsection{Customizing the sampler execution order}
+
+The ordering of sampler execution can be controlled as well.  This allows for sampler functions to execute multiple times within a single MCMC iteration, or the execution of different sampler functions to be interleaved with one another.
+
+The sampler execution order is set using the function \cd{setSamplerExecutionOrder}, and the current ordering of execution is retrieved using \cd{getSamplerExecutionOrder}.  For example, assuming the MCMC configuration object \cd{mcmcConf} contains 5 samplers:
+
+<<setSamplerExecutionOrder, eval=FALSE>>=
+## first sampler to execute twice, in succession:
+mcmcConf$setSamplerExecutionOrder(c(1, 1, 2, 3, 4, 5))
+
+## first sampler to execute multiple times, interleaved:
+mcmcConf$setSamplerExecutionOrder(c(1, 2, 1, 3, 1, 4, 1, 5))
+
+## fourth sampler to execute 10 times, only
+mcmcConf$setSamplerExecutionOrder(rep(4, 10))
+
+## omitting the argument to setSamplerExecutionOrder()
+## resets the ordering to each sampler executing once, sequentially
+mcmcConf$setSamplerExecutionOrder()
+
+## retrieve the current ordering of sampler execution
+ordering <- mcmcConf$getSamplerExecutionOrder()
+
+## print the sampler functions in the order of execution
+mcmcConf$printSamplers(executionOrder = TRUE)
+@
+
 \subsubsection{Monitors and thinning intervals: \cd{printMonitors}, \cd{getMonitors}, \cd{addMonitors}, \cd{setThin}, and \cd{resetMonitors}}
 
 An MCMC configuration object contains two independent sets of variables to monitor, each with their own thinning interval: \cd{thin} corresponding to \cd{monitors}, and \cd{thin2} corresponding to \cd{monitors2}.  Monitors operate at the \textit{variable} level.  Only entire model variables may be monitored.  Specifying a monitor on a \textit{node}, e.g., \cd{x[1]}, will result in the entire variable \cd{x} being monitored.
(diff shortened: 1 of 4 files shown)


danielturek:
Updated this PR, with new options in `mcmc$run` function to accept thinning intervals (`thin` and `thin2`) as runtime arguments.

Default is to use the thinning intervals from the MCMC configuration, but if either argument is provided at runtime, then that value will be used instead.

Also updated `runMCMC` to accept `thin` argument at runtime.

Also updated Roxygen documentation and User Manual.

Fixed #703

danielturek:
@paciorek Nice eye on the Roxygen for `runMCMC`.  I'm just pushing a change to this, to correct it.

Yes, that's right, in order to modify the sampler execution order (newly added functionality), you have to go down to the level of using `mcmc$run()`.  In my mind, this level of customization (similar to choosing the sampling algorithms that are assigned) will be reserved for those actually creating and manually executing the MCMC algorithm (rather than those using `nimbleMCMC` or `runMCMC`, which by design only provide more "common" MCMC options, such as multiple chains, initial values, thinning, burnin, etc).

Yes, same answer goes for why `runMCMC` only allows dynamically setting the first thinning interval `thin`.  Also, since `runMCMC` accepts a `burnin` argument, it would get confusing as to what the burn-in applies to.  So, as far as `runMCMC` is concerned, we're dealing with a single set of monitors, a single burnin, and a single thinning interval.

perrydv:
@danielturek thanks for these changes.  I am late looking at them.

About having samplers used multiple times within an MCMC kernel (see comment by @paciorek on the Issue), I have found it can be an effective sampling strategy.  Basic example would be with a RW MH block sampler: The possibility of accepting a change to multiple dimensions can be improved if several proposals are made, perhaps spread out among other samplers.  To put it another way, it is a shame when block sampling dimensions because they don't mix well to then have them not mix well for a different reason: because they aren't updated enough.  I can provide examples, but I'm not sure if it's worth a debate vs. just providing the flexibility so it is available, which seems to be what we are doing.

@danielturek Do I follow correctly that such a custom ordering could now be provided *either* in the MCMC configuration (which would then be the default) *or* from `mcmc$run` (which would then over-ride whatever was in the MCMC configuration?)?  If so, that sounds good to me.

My one quibble would be to allow `thin2` to be modified from `runMCMC`.  My sense from users is that `runMCMC` is a handy entry point and that it is a pain to keep track of just what can be done in each of the ways to use an MCMC.   So, to the extent possible, I think it would be preferable to provide easy elements of flexibility via `runMCMC`.  The burnin could be defined in relation to `thin`, not `thin2`.  Having said that, reordering/reusing samplers does seem sufficiently advanced to limit it to `mcmc$run`. 

Let me know if I didn't fully follow the issues under discussion.

danielturek:
@perrydv In response to your suggestion about allowing `runMCMC()` to accept `...` argument, then passing `...` along to the `mcmc$run` method, as a catch-all for any additional runtime arguments.

Thinking about it, I'm skeptical that our processing and handling of nimbleFunction run-time arguments and their type declarations would correctly handle the `...` argument.  Does that sound right to you?  Or do you think that by the magic of R argument matching, it might work?

danielturek:
Despite the one travis failure, this is ready to merge as far as I'm concerned.

danielturek:
One more update to this PR.

Upon looking at the MCMC and how it handles `enableWAIC`, I feel that `enableWAIC` should be a property of the MCMC **configuration**, rather than a property of the MCMC **algorithm**.

I've created a commit that makes this change.  Practically, the only consequence of this change is that `enableWAIC = TRUE` can now be set using:

- `configureMCMC(..., enableWAIC = TRUE)`   ## new
- `configurationObject$setEnableWAIC(TRUE)`   ## new
- `buildMCMC(..., enableWAIC = TRUE)`    ## old, still supported

Thus giving more options for users.

But from a design standpoint, I think this change is also important since all properties of the MCMC are defined by the configuration, so it didn't make sense to have **only this one** being a property of the algorithm, alone.

I genuinely welcome the thoughts, opinions, and critical code review from @NLMichaud .

danielturek:
For @NLMichaud , these changes alone regarding `enableWAIC` can be viewed [here](https://github.com/nimble-dev/nimble/pull/706/commits/951bec574a9f67c4dbfc6ef8e2e44a30a0b1b220)

danielturek:
@NLMichaud good point, although I agree it probably would never come up.

I just pushed an update to fix this.

Now, in theory, a user could have a configuration object with `enableWAIC = FALSE`, but they could then enable WAIC for particular MCMCs created from that configuration using `buildMCMC(conf, enableWAIC = TRUE)`, without altering the underlying configuration object.

Looks ok?

NLMichaud:
Looks good, thanks!

danielturek:
@perrydv @paciorek Ok if I merge this, despite the OSX travis failure?

How do we stand on that?

perrydv:
Thanks for all this @danielturek 

@paciorek: The comment you questioned was simply in response to your comment on Issue #701, which this PR addresses.  Not important.

@danielturek: You didn't address the nature of the Travis failure, but from inspection I gather it is one of the glitches in setting up R packages on Travis.  Although it is probably trivial in this case, I don't know how far down a road we want to go of ignoring testing since testing itself is broken.  @paciorek Do you know if this will resolve easily?  @danielturek Do you want to relaunch tests to see if it works itself out?

paciorek:
Strangely, devel is now passing Travis (I made a very minor edit a couple days ago). Yet this branch is not even though .travis.yml and requirements.txt are the same and Travis was run on this branch since Travis succeeded on devel.

At some point I should dig more deeply into how we might set up a more robust Travis environment in which we control how R and its packages are installed.

But ... to keep kicking the can down the road ... I've initiated another Travis attempt for this branch.

paciorek:
Ok, that didn't work. I'm now monkeying some (somewhat blindly, somewhat informed) with the Travis setup...

paciorek:
Forcing Travis to use R 3.4 seems to have gotten us past the Mac OS R 3.5 Travis issue with rlang.

@danielturek back to you to check on the full testing results and merge when appropriate.

PR #708, opened 2018-05-02
Pmcmc user filter

NLMichaud:
This PR allows user-defined filters to be used in PMCMC algorithms.  User-defined filters can be given as the `pfType`  control list argument to either the `RW_PF` or `RW_PF_block` sampler, and must satisfy the following in order to be used generically within our PMCMC samplers:

- The function provided to `pfType` must be the result of a call to `nimbleFunction()`
- The `nimbleFunction` provided to `pfType` must have setup code that accepts the following (and only the following) arguments:

1. `model`, the nimble model object that the MCMC algorithm is defined on.
2.  `latents`, a character vector specifying the latent model nodes over which the particle filter will stochastically integrate over to estimate the log-likelihood function.
3.  `control`, an R `list` object.  Note that the `control` list can be used to pass in any additional information or arguments that the custom filter may require. 

- The `nimbleFunction` provided to `pfType` must have a `run` function that accepts a single integer arugment (the number of particles to use), and returns a scalar double (the log-likelihood estimate).

- The `nimbleFunction` provided to `pfType` must define, in setup code, a `modelValues` object named `mvEWSamples` that is used to contain equally weighted samples of the latent states (that is, the `latents` argument to the setup function.)  Each time the `run()` method of the `nimbleFunction` is called  with number of particles `m`, the `mvEWSamples` `modelValues` object should be resized to be of size `m` via a call to `resize(mvEWSamples, m)`.


NLMichaud:
@perrydv @paciorek @danielturek I think this is ready to be submitted, but if any of you would like to check out the updated manual info I've added to the Particle MCMC section of Chapter 8, that would also be great.  Please let me know if the information is confusing or does not provide enough detail. Thanks!

perrydv:
I made some revisions to the User Manual entry.

PR #711, opened 2018-05-07
If returnType() is provided, check that there is a return.

### LINKED ISSUE #669: better error trapping when `return` statement is missing.
Here's a case where the user (me) forgot to use return() and the result is an R crash at run-time. 

```
a <- nimbleFunction(
    run = function() {
        returnType(double(1))
        out <- rnorm(10, 0, 1)
    })

ca <- compileNimble(a)
ca()  # crashes
```

perrydv:
Fixes #669.

It appears that in C++ a missing return statement yields undefined behavior, and this is giving us a crash.

I added a check when the returnType() statement is extracted.  If there is a returnType statement, and if it is not in a virtualNimbleFunction, then I added a simple check for the presence of `return` in the code.  Note that incorrect return types are (or should be) trapped at another stage.  This simply makes sure there is a return statement somewhere in the code.  It would take more work to ensure that every path through the code ends in a return statement.

### DIFF
--- a/packages/nimble/R/RCfunction_core.R
+++ b/packages/nimble/R/RCfunction_core.R
@@ -80,7 +80,7 @@ nfMethodRC <- setRefClass(
                 nf_checkDSLcode(code, methodNames, setupVarNames)
             generateArgs()
             generateTemplate() ## used for argument matching
-            removeAndSetReturnType()
+            removeAndSetReturnType(check = check)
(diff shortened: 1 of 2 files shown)


perrydv:
@paciorek do you want to look at this fix you Issue #669, which you created?

paciorek:
LGTM. Merging.

PR #719, opened 2018-05-10
Fully initialize model between MCMCs in MCMCsuite.  

### LINKED ISSUE #718: MCMCsuite does not fully reset initial model states between MCMCs
`MCMCsuite`, in its method `run_nimble`, attempts to reset the model between each MCMC using:

```{r}
Cmodel$setInits(inits);     calculate(Cmodel)
```

If the `inits` contains elements for every parameter and latent state in the model, this should fully reset the model.  However, if `inits` is not complete, then some of the parameters and/or latent states will have been set by simulation and sampling in the first of the MCMCs to run.  Then those states will remain unmodified in the model upon initiating the next of the MCMCs.

This should often be harmless, but it raises two problems.  

1. It means there can be a slight dependence on the order of MCMCs in the MCMCsuite results.  While usually trivial, this is not intended.

2. It can lead to invalid model states at the start of the second MCMC if the states in `inits` are in some way incongruous with the states left unchanged from the first MCMC.  The way this issue was discovered is that such a situation led the second MCMC to be trapped in the infinite `while` loop of Issue #715 (`AF_slice` sampler).

A pull request to fix this issue will be filed shortly.

perrydv:
Fixes #718 

Also update some warning messages.

The code changes are small and should be easy to follow.

### DIFF
--- a/packages/nimble/R/MCMC_build.R
+++ b/packages/nimble/R/MCMC_build.R
@@ -184,6 +184,7 @@ buildMCMC <- nimbleFunction(
         progressBarIncrement <- niter/(progressBarLength+3)
         progressBarNext <- progressBarIncrement
         progressBarNextFloor <- floor(progressBarNext)
+        if(iter < 1) return()
         for(iter in 1:niter) {
             checkInterrupt()
             if(time) {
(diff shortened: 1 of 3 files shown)


danielturek:
Fine with me.  Why the need to return from the MCMC$run method, when `niter < 1` ?  Does that occur somewhere?

danielturek:
@perrydv Just pushed some very minor stylistic changes to MCMC_suite (the new model initialization stuff).

Ready to merge, as far as I'm concerned.

perrydv:
@danielturek about the `niter < 1` case: I found it useful for debugging to be able to see the model state after initialization but before any sampling.  This requires calling `mcmc$run(niter = 0)`.  

I will wait to merge until I have time when I can handle any unexpected glitches, probably next week.

danielturek:
@perrydv That makes sense.

Issue #716, opened 2018-05-09
adjust calculate and simulate with invalid dynamic indexes


paciorek:
Have calc and sim produce warnings not errors and return NaN(s) when there is an invalid index. That way MCMC samplers will not error out if an invalid index is used.

paciorek:
Current state of code in branch robustify_dyn_idx_nodefun:

1) NaN produced for calculate and simulate if have invalid numeric-valued dynamic index in both compiled/uncompiled

2) if have NA as the invalid index, compiled code produces NaN, while uncompiled errors out because of the usual R trouble with `if(NA)`. 

Adding a check for NA in the nodeFunctions would add additional complexity to the conditionals in the nodeFunction generation, so I am somewhat inclined to leave this inconsistency.

paciorek:
One additional item to consider. In an MCMC we can now produce many many message about invalid dynamic indexes. We may want to consider a mechanism for nimCat() and nimPrint() to not print repeated identical messages (like R's options for warning()).

paciorek:
I'm now realizing that while setting logProb values to NaN is straightforward, when we have multivariate nodes that have parameters that are dynamically indexed, I don't see an easy way of setting node values to NaN when simulate is called (since we'd have to somehow get info into the nodeFunction generating code that tells us the dims of the node so we can create a vector or matrix of NaNs).

I think our main use case here is making sure MCMCs don't error out, so I'm inclined to have calls to simulate error out (as is currently done) while calls to calculate return NaN. But interested if @perrydv or @danielturek have other thoughts.

paciorek:
Ok, given that a deterministic node calculate actually calls simulate and we need NaN injected so that downstream logProbs become NaN, I don't think there is a way around more complicated code generation.

I think the thing to do is pass in the dims of the node to the simulate nodeFunction generator functions and then code generate using nimArray in cases where it's not a scalar node.

paciorek:
Fixed by PR #722

PR #722, opened 2018-05-14
have invalid dynamic index return NaN and warn rather than erroring

paciorek:
For review and testing only; do not merge.

This addresses issue #716 .


paciorek:
Have now modified simulate so that NaN arrays can be produced.

paciorek:
@perrydv or @danielturek if you'd like to review, please do. I believe this is ready to merge.

danielturek:
Not close enough to this issue to be able to offer helpful feedback.

PR #725, opened 2018-05-18
experimental AD functionality 

NLMichaud:
This PR introduces experimental AD functionality to nimble.  Functionality is controlled by the `experimentalEnableDerivs` `nimbleOption`.  Examples of using AD, along with warnings about its current limitations, can be found in `chapter_chapter_ADinNimbleFunctions.Rnw`.


paciorek:
@NLMichaud @perrydv can we remove this PR given ongoing work since May?

perrydv:
Yes I don't think having this sit around as a PR is useful right now.  We did learn that there is some obstacle to compilation on linux.

Issue #726, opened 2018-05-23
Incorrect (excessive) checking for expressions in multivariate distribution arguments.


perrydv:
The following should work but is trapped as an error:

```
library(nimble)

mc <- nimbleCode({
   x[1] <- 1
   for(i in 2:10) {
       x[2] ~ dcat(transition_probs[x[i-1], 1:2])
   }
}
)

transition_probs <- matrix(c(.1, .9, .4, .6), nrow = 2)

m <- nimbleModel(
    mc,
    data = list(transition_probs = transition_probs)
)
```

Result:

"defining model...
Error in checkMultivarExpr() : 
  Error with parameter 'prob' of distribution 'dcat': multivariate parameters cannot be expressions; please define the expression as a separate deterministic variable and use that variable as the parameter."

There is a recently built-in workaround:
```
nimbleOptions(disallow_multivariate_argument_expressions = FALSE)
```
but that is clunky.

Possible solutions:
- Make the checking smarter.
- See if it would work to leave non-scalar expression arguments as is, including not trying to lift them.

perrydv:
This was addressed by PR #727

PR #727, opened 2018-05-28
experimentally do not lift multivariate expression arguments

perrydv:
Experimental:

This PR takes a step toward not trying to lift multivariate expression arguments.  We don't allow such things anyway, but we may need to in order to address Issue #726.

This PR is for travis testing purposes of one idea.


perrydv:
Test failures are occurring from `test-size` because expected failures are no longer failures.

paciorek:
@perrydv was our final decision at last meeting to have multivar exprs trigger warning not error? If so, I can update the failing test-size cases.

perrydv:
@paciorek I think we decided on the more conservative choice to have multivariate expressions trigger error as before.

The change is that checking for multivariate expressions will not recurse inside `[`.

I think it would be good if the code can easily be switched to allow multivariate expressions even if they are not lifted (and requiring that they are not lifted).  To do so, I need to figure out the best access to argument types.  I realized this needs to include user-defined distributions.   I am going to flag you on a couple of comments in the code to ask you to help with this since you wrote the user-defined distribution system.

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -668,29 +668,42 @@ modelDefClass$methods(expandDistributions = function() {
...
         if(BUGSdecl$type != 'stoch') next
-        dist <- deparse(BUGSdecl$valueExpr[[1]])
+        ## dist <- deparse(BUGSdecl$valueExpr[[1]])
+        dist <- BUGSdecl$distributionName
+        ## The following line is a one-time insertion to break testing in any case
+        ## where the condition fails.
+        ## If the condition is always met, we can use BUGSdecl$distributionName in place of deparse(BUGSdecl$valueExpr[[1]])
+        ## if(dist != BUGSdecl$distributionName)
+        ##     stop(paste0("dist (", dist,") != BUGSdecl$distributionName (",BUGSdecl$distributionName,")"))
         types <- nimble:::distributionsInputList[[dist]]$types
```
perrydv: @paciorek Here is one place we need to look up types and I am not sure this is the right way to do it.  I think we want more processed types (I can see that they exist somewhere) and also to be sure that types from user-defined distributions will be accessed.

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -979,11 +992,14 @@ modelDefClass$methods(liftExpressionArgs = function() {
         
         if(BUGSdecl$type == 'stoch') {
             params <- as.list(valueExpr[-1])   ## extract the original distribution parameters
+
+            types <- nimble:::distributionsInputList[[BUGSdecl$distributionName]]$types
```
perrydv: @paciorek The issue of how to correctly look up types occurs here too.

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -1018,20 +1033,25 @@ modelDefClass$methods(liftExpressionArgs = function() {
...
+        if(callText == 'chol')         return(TRUE)    ## do lift calls to chol(...)
+        if(callText == 'inverse')      return(TRUE)    ## do lift calls to inverse(...)
+        if(callText == 'CAR_calcNumIslands') return(TRUE)    ## do lift calls to CAR_calcNumIslands(...)
+        if(callText == 'CAR_calcC')    return(TRUE)    ## do lift calls to CAR_calcC(...)
+        if(callText == 'CAR_calcM'  )  return(TRUE)    ## do lift calls to CAR_calcM(...)
+        if(callText == 'CAR_calcEVs2') return(TRUE)    ## do lift calls to CAR_calcEVs2(...)
+        if(callText == 'CAR_calcEVs3') return(TRUE)    ## do lift calls to CAR_calcEVs3(...)
+        if(length(paramExpr) == 1)     return(FALSE)   ## don't lift function calls with no arguments
+        if(callText == '[')            return(FALSE)   ## don't lift simply indexed expressions:  x[...]
+        if(any(grepl(callText, types))) return(FALSE)  ## beyond above cases, don't lift non-scalar arguments
```
perrydv: @paciorek What we want here is to check if the relevant type is non-scalar and if so return FALSE.  As written, this returns FALSE if any type is non-scalar.  As written, the types object has raw entries like "mean = double(1)" (but no entries for scalars) but I think the processed information exists so it would be good to use it.

paciorek:
We have a getDimension function.  Works for user-defined distributions as well as NIMBLE-defined ones.

E.g., here ddirchmulti is a user-defined dist:
```
> getDimension('ddirchmulti',includeParams = TRUE)
value alpha  size 
    1     1     0 
```

@perrydv Let me know if you need more info or want me to actually code something.

perrydv:
Thanks @paciorek.  That's what I needed.  I can try to make use of `getDimension` where needed shortly.

perrydv:
pushing a new version for testing.

PR #728, opened 2018-05-28
make setSize check for local vs. non-local assignment

### LINKED ISSUE #724: possible scoping issue with R execution of nimbleFunction with resized matrices
This came up in Claudia's work. Here's a simplified example. Perhaps I am doing something wrong?

Compiled execution is fine, but uncompiled it doesn't seem to realize that 'samples' is now 5x5 and so can't assign into the newly allocated space (unless one does `samples[2,2] <- 3`.

Haven't delved into this; just reporting it.

```
mynf <- nimbleFunction(
    setup = function(){
        samples <- matrix(0, nrow=1, ncol=1)
    },
    run = function() {
        setSize(samples, 5, 5)
        samples[2,2] <<- 3  
        returnType(double(2))
        return(samples)
    })
rnf = mynf()
rnf$run()
#Error in `[<-`(`*tmp*`, 2, 2, value = 3) : subscript out of bounds
cnf=compileNimble(rnf)
cnf$run()  # runs fine
```

perrydv:
The makes the R version of setSize determine whether it should make local or non-local assignment in the calling frame.

Fixes #724


perrydv:
This passed tests on travis but failed on appveyor, so I suspect an appveyor glitch.  I will restart it.

PR #730, opened 2018-05-30
Only enter name as neededRCfun when allowed as an argument to optim.

### LINKED ISSUE #729: Infinite recursion when local variable has same name as function
```
library(nimble)

f <- nimbleFunction(
    run = function(x = double(0)) {
        f = x + 1
        return(f)
        returnType(double(0))
    }
)
cf <- compileNimble(f)
```
R says:
Error: evaluation nested too deeply: infinite recursion / options(expressions=)?

perrydv:
Fixes #729 

The problem was that in size processing, any name encountered that is findable as an rcFunction (nimbleFunction) was being entered in neededRCfuns, even if it was not being used as a function.  This led to an infinite recursion.  In this PR, recording a name (when not used as a function) in neededRCfuns is only done in a circumstance where a function could be an argument, which occurs for optim.


paciorek:
@perrydv ok to merge this?

PR #732, opened 2018-06-01
minor change to checking dimensions in varInfo3

paciorek:
This adds a check for whether any maxs value is smaller than mins value so we don't rely on catching the error by seeing a 0 as either max or min.


danielturek:
@paciorek Good idea.  The informative messages are worth a lot.

PR #733, opened 2018-06-01
added error trap for invalid param name in getParam

paciorek:
This does a better job of trapping and giving a good error message for an invalid parameter name by putting a check in makeParamInfo that catches errors in both uncompiled use of getParam and in compilation of nf's that use getParam.

Note that at one point this trapping was done via:
`paramIDvec <- sapply(distNames, getParamID, param)`

in makeParamInfo but at some point one of us changed makeParamInfo to directly access info in a list with distributionInfo (not sure why that was done...)


perrydv:
@paciorek  `makeParamInfo` was an (in)efficiency bottleneck.  Those `sapply` statements were called many times and were slow.  I'd want to be sure that any new error-trapping here does not slow down processing substantially.

paciorek:
Ok, fair enough.  The new code simply does `any(is.na(paramIDvec)`, which is comparable to the current check which does `length(unique(typeVec)) != 1 || length(unique(nDimVec)) != 1`. paramIDvec will always be trivially short, so I'm supposing this comes up only in models with many nodes? Can you say more about what sort of test case you were using in which the speed disparity came up?

paciorek:
I did a basic speed test with a model with 10000 dnorm observations and 10000 corresponding dnorm random means. Didn't see any speed impact on model building or in creating/compiling a nimbleFunction that loops over 10000 nodes and calls getParam. Also `any` and `is.na` are direct C calls in R, so that is suggests there should not be much impact. @perrydv let me know if you don't think this is convincing / had some other use case in mind.

perrydv:
I think the cases that bogged down `makeParamInfo` in the past came from `model$getParam(nodes[i], 'mean')`, where `nodes` might be very long.  It ends up looking up the distribution for every element of `nodes` (in a single call to `makeParamInfo`), and in typical cases there is a lot of redundant processing (if every node actually follows the same distribution) in order to ensure correct general behavior.  I agree the `any(is.na(...))` shouldn't be a bottleneck.  LGTM @paciorek

PR #737, opened 2018-06-04
avoid skipped tests with testthat 2.0.0

paciorek:
As of 2.0.0 of testthat any test_that in which an expect_xxx is not called results in a report of an empty test. Also nested test_that(test_that(expect_xxx)) results in empty tests.

This PR cleans that up in all cases except test-size.

For test-size, we have various test_that calls where we do `eval(code)` and in some cases there is no output and in others there is a warning.  There are 63 of these with 17 giving warnings, so for now I am not going down the rabbithole of distinguishing these cases as there is no expect_xxx that will handle either a warning or no output.


paciorek:
Also worth noting that I did a manual scan of the various output produced by tests under testthat 2.0.0 and earlier testthat, and it looks like this testing is consistent with earlier testthat.

Issue #721, opened 2018-05-10
trap incorrect non-scalar variable creation more clearly


perrydv:
If a nimbleFunction contains code like

`x[1] <- 5`

before x has been created as a vector, the error message from compileNimble is not helpful.

paciorek:
Fixed by PR #739

PR #739, opened 2018-06-04
trap error when non-existent array in a nf is indexed

paciorek:
Fixes issue #721


perrydv:
If it passes tests, LGTM

Issue #715, opened 2018-05-09
avoid infinite loops in AF_slice (and check for this in slice sampler)


paciorek:
AFSS can get stuck inside looping.

@NLMichaud can you remind us when this can occur?

@danielturek could the same thing happen with slice and slice_ESS?

Solution is presumably to break out of the loop(s) after some max number of iterations.

danielturek:
@paciorek @NLMichaud I don't believe this can happen in the slice sampler, but I'll wait to hear Nick's description of the problem-case to be sure about that.

No, there are no such loops in the elliptical slice sampler.

perrydv:
Based on code inspection, I think that yes the same infinite behavior of the while loop could occur [here](https://github.com/nimble-dev/nimble/blob/47be0c76987aabf4290590ec587f0264e5e4ef42/packages/nimble/R/MCMC_samplers.R#L474) in `sampler_slice`.

When the situation is healthy, the conditions for exiting the loop are achieved.  But in an unhealthy situation like a bimodal conditional distribution, the `L` and `R` bounds can contract into each other forever.  Two ideas for checking would be (i) track the number of iterations and set a maximum or (ii) check abs(R - L) and if it is 0 (or < epsilon), there is a problem.

paciorek:
@perrydv I don't see how the infinite contraction could occur unless one has point mass in the density, the current value is at that point mass, and the density is not flagged as a discrete distribution.

paciorek:
I've looked at the slice, AF_slice and ess samplers and all of them have a while loop where they could get stuck if for some reason the logProb is always NA or NaN or if the L and the R bounds contract infinitely (see above for a question about cases in which this could occur). 

Here's my tentative plan for this (for each of these samplers):
 - add a 'maxContraction' control argument for the maximum number of times the interval contracts
 - add a 'warnMaxContraction' control argument that controls if a message is printed when the maximum number of contractions is reached
 - if the maxContraction is reached, break out of the loop and copy from mvSaved to model. Note that for AF_slice if any of the dimensions reach maxContraction, I will reject the samples for all dimensions --  This is for simplicity and in recognition that if this happens something is likely pretty wrong, so causing the chain to not move may clue the user in to something going on.

Definitely welcome any suggestions/thoughts.

danielturek:
Following the specifics of these troublesome cases only at a very high level, my reactions are:

[...]

PR #741, opened 2018-06-05
gracefully exit out of slice / AF_slice/ ESS while loop

paciorek:
This exits out of the while loop when reach maxContractions or L and R converge to each other. 

Also adds a basic test in for the slice_sampler case.

@danielturek please take a look, and @perrydv if you like.

### DIFF
--- a/packages/nimble/R/MCMC_samplers.R
+++ b/packages/nimble/R/MCMC_samplers.R
@@ -437,6 +437,11 @@ sampler_slice <- nimbleFunction(
         adaptInterval <- if(!is.null(control$adaptInterval)) control$adaptInterval else 200
         width         <- if(!is.null(control$sliceWidth))    control$sliceWidth    else 1
         maxSteps      <- if(!is.null(control$sliceMaxSteps)) control$sliceMaxSteps else 100
+        maxContractions        <- if(!is.null(control$maxContractions))
+                                      control$maxContractions else 100
+        maxContractionsWarning <- if(!is.null(control$maxContractionsWarning))
+                                      control$maxContractionsWarning else TRUE
+        eps <- 1e-15
         ## node list generation
         targetAsScalar <- model$expandNodeNames(target, returnScalarComponents = TRUE)
         calcNodes <- model$getDependencies(target)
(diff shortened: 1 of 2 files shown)


paciorek:
Whoops, meant to note this fixes #715

packages/nimble/R/MCMC_samplers.R
```diff
@@ -471,15 +476,26 @@ sampler_slice <- nimbleFunction(
...
+            ## The checks for R-L small and max number of contractions are for cases where model is in
+            ## invalid state and lp calculations are NA/NaN or where R and L contract to each other
+            ## division by R+L+eps ensures we check relative difference and that contracting to zero is ok
             if(x1 < x0) { L <- x1
                       } else      { R <- x1 }
             x1 <- L + runif(1, 0, 1) * (R - L)           # sample uniformly from (L,R) until sample is inside of slice (with shrinkage)
             lp <- setAndCalculateTarget(x1)
+            numContractions <- numContractions + 1
+        }
+        if((R-L)/(abs(R)+abs(L)+eps) <= eps | numContractions == maxContractions) {
```
perrydv: @paciorek Should the numerator be wrapped in `abs`?

packages/nimble/R/MCMC_samplers.R
```diff
@@ -437,6 +437,11 @@ sampler_slice <- nimbleFunction(
         adaptInterval <- if(!is.null(control$adaptInterval)) control$adaptInterval else 200
         width         <- if(!is.null(control$sliceWidth))    control$sliceWidth    else 1
         maxSteps      <- if(!is.null(control$sliceMaxSteps)) control$sliceMaxSteps else 100
+        maxContractions        <- if(!is.null(control$maxContractions))
+                                      control$maxContractions else 100
```
perrydv: @paciorek Should we choose a larger default?  I'm thinking we want to be conservative, with the goal to avoid a infinite loop in invalid or pathological cases but also to avoid false positives.

packages/nimble/R/MCMC_samplers.R
```diff
@@ -548,13 +570,23 @@ sampler_ess <- nimbleFunction(
...
+            ## for cases where model is in invalid state and lp calculations are NA/NaN or where
+            ## theta interval contracts to zero
             if(theta < 0)   theta_min <- theta   else   theta_max <- theta
             theta <- runif(1, theta_min, theta_max)
             model[[target]] <<- f*cos(theta) + nu*sin(theta) + target_mean
             lp <- calculate(model, calcNodes)
+            numContractions <- numContractions + 1
         }
-        nimCopy(from = model, to = mvSaved, row = 1, nodes = calcNodes, logProb = TRUE)
+        if(theta_max - theta_min <= eps | numContractions == maxContractions) {
```
perrydv: @paciorek Why did you use a relative difference check in `sampler_slice` but an absolute difference here?

packages/nimble/R/MCMC_samplers.R
```diff
@@ -639,16 +676,28 @@ sampler_AF_slice <- nimbleFunction(
...
+                ## interval contracts to zero
                 if(prop < 0) { Lbound <- prop }
                 else         { Rbound <- prop }
                 nContracts[i] <<- nContracts[i] + 1
                 prop <- Lbound + runif(1, 0, 1) * (Rbound - Lbound)
                 x1 <- x0 + prop * eigenVec
                 lp <- setAndCalculateTarget(x1)
+                numContractions <- numContractions + 1
             }
+            if(Rbound - Lbound <= eps | numContractions == maxContractions)
```
perrydv: @paciorek Same question on relative vs. absolute check.

paciorek:
The convergence seems to generally happen in less than 100, so probably doesn't matter much - I'm fine with making it bigger, say 1000, but I think the R-L part of the check will be triggered first. 

For the R-L part of the check, both AF_slice and ESS are contracting to 0, so dividing by the magnitude of R and L (which are called Rbound,Lbound and theta_min,theta_max in those samplers) means we are dividing by numbers going to zero. Given the eps in the denominator, if we use the relative check, it would essentially amount to having (R-L)/eps < eps if we wanted it as a relative check. I suppose that is logically consistent, but it does mean that we don't break out until R-L is less than 1e-30, which seems excessive.

PR #749, opened 2018-06-13
Fix bootstrap filter

NLMichaud:
This PR fixes a bug introduced to the bootstrap filter by changing an instance of `normWts` to `wts` in the `bootFStep` function of filtering_bootstrap.R.


paciorek:
Nick, this PR seems to have commits from before that are already merged. It would probably be fine, but I'm going to make the same fix to normWgts in the bootstrap filter in a separate branch in which I also wrap the entire test_filtering body in test_that so errors are fully reported by test_that.

paciorek:
The issue is addressed in PR #750 . Closing without merging.

PR #771, opened 2018-07-29
RW_wishart sampler: non-conjugate wishart & inverse-wishart nodes

danielturek:


danielturek:
@perrydv @paciorek Thoughts on what to call this sampler?  At the moment it's called `RW_wishart` following the existing `RW_multinomial` and `RW_dirichlet` samplers for those nodes.

But unlike those others, this sampler isn't exclusively only for Wishart (`dwish`) nodes.  It also operates on inverse-Wishart (`dinvwish`) nodes, and more generally any symmetric positive-definite matrix node.

I'm leaning towards keeping it as-is, named `RW_wishart`.  But I wanted to raise the topic for discussion.

paciorek:
I'm fine with RW_Wishart.

danielturek:
I'm going with `RW_wishart` then, which I also like.

I just updated NEWS.  Once testing completes on this branch, I'll merge it in.

paciorek:
A few fairly minor comments.

  - Should we allow the user to provide an initial proposal covariance matrix. I realize this is on a non-standard scale so often user wouldn't know this. But they might have run the model previously and have an idea from the adaptation of what the propCov should be.
  - You say it could be used on non-Wishart/invWishart distributions but setup code check prevents this.
  - I think we need to avoid having any error messages as vague as 'something went wrong'. At least we need to say what function call this is occurring in, and how about something indicating incorrect dimensionality.
  - I think we should take this out: `propValue_chol <<- propValue_chol * 0`. There is nothing in the code that sets values in the lower triangle and all other values are explicitly overwritten. Unless we have some sort of verbose calculation nimbleOption. 
  - I'd say just do this as one line: `model[[target]] <<- t(propValue_chol) %*% propValue_chol`, but perhaps the C++ compiler optimizes this away anyway?
  - As far as the Hastings adjustment calculation... that brought up some math I've forgotten or maybe never knew. I worked through it enough to think that it's right, though it's possible I missed something.

danielturek:
@paciorek Fantastic, thanks for the detailed code review.  Let me address your comments:

- I added the ability to provide a proposal covariance through the `control` list.
- I removed the check for operating only on `dwish` or `dinvwish` distributed nodes.
- I removed the (precautionary) `stop('something went wrong')` messages, which should never happen.
- I removed the unnecessary line `propValue_chol <<- propValue_chol * 0`.
- I removed the extra line, and just went with `model[[target]] <<- t(propValue_chol) %*% propValue_chol`.
- Thanks for checking over the math.  I worked through it, and also the results agree with conjugate sampling on a few examples, so I have confidence in it.

Once testing passes, I'll assume this is ready to merge.

danielturek:
Noting that testing has passed, but this PR page doesn't seem to indicate such.

PR #772, opened 2018-08-04
set up RW and RW_block to calculate prior log prob first.  Added tries control param to RW_block.

perrydv:
Do not merge.

This PR is to trigger testing.

This change may speed up performance in some cases by avoiding long model calculations if a proposal's prior probability is 0.

In addition I added a "tries" control parameter to RW_block, to easily set it to do `tries` iterations.

The change in RW_block to calculate priors first assumes that one element of the target is not a dependency of another element.  This would not work for something like sequential states of a state-space model.  We can generalize if needed.


danielturek:
@perrydv Once this is settled and finalized, I'd like to edit for aesthetics.  (see MCMC style guide)

perrydv:
@danielturek Can you remind me where the MCMC style guide is.

@paciorek and @danielturek  I think the logic on this branch is basically ok, but it is impossible to reproduce exactly the MCMC chains as on devel during testing.  On devel, some cases of a proposal to an unsupported region of the prior create logProb == -Inf, while other cases create logProb == NaN, depending on what calculations are in the dependencies.  On the branch here, all of those cases essentially become -Inf, because dependencies are not calculated after that.  However, the decide() function does not do runif(1, 0, 1) if it gets NaN as input, but it does do runif(1, 0, 1) if it gets -Inf as input.  That makes it impossible, for testing purposes, to ensure the exact same sequence of random number draws as previously throughout the run of the MCMC, so we end up with different chains.  So I think the path forward could be:

1. Do some special case testing to be sure that this really explains all testing discrepancies.

2. Revise testing expectations as part of this branch and PR.

danielturek:
@perrydv Yes, the [MCMC style guide](https://github.com/nimble-dev/nimble/wiki/MCMC-Style-Guide) is linked from the main Wiki page, under the section Developer Manuals and Tutorials.

Noted about the reasons it's impossible (or very difficult) to maintain consistent testing.  In light of everything, since the updated code is simply better and more efficient, I would personally say update the expected test results, to match this new behaviour.

perrydv:
@paciorek I got pinged on the status of this branch and PR.  I remember that it was ready to go except that there was going to be some labor involved in updating test results because it calls for updating expected test results since the exact sequence of an MCMC will change.  The effort required for that led to stalling.

paciorek:
@perrydv Ok, I can put this on my to-do to walk through and do some additional manual checking. Let's wait for me to do that before we merge.

perrydv:
@paciorek The version of the C++ `decide` in this branch was temporarily modified to try to emulate past behavior.  Evidently that is not entirely the case, so we need to track down cases like the "dugongs" example to see if the discrepancies are any cause for concern.

packages/nimble/R/MCMC_samplers.R
```diff
@@ -252,9 +267,18 @@ sampler_RW_block <- nimbleFunction(
         adaptInterval  <- if(!is.null(control$adaptInterval))  control$adaptInterval  else 200
         scale          <- if(!is.null(control$scale))          control$scale          else 1
         propCov        <- if(!is.null(control$propCov))        control$propCov        else 'identity'
+        tries          <- if(!is.null(control$propCov))        control$tries          else 1
```
danielturek: Error on this line
```suggestion
        tries          <- if(!is.null(control$tries))          control$tries          else 1
```

packages/nimble/R/MCMC_samplers.R
```diff
@@ -169,10 +170,24 @@ sampler_RW <- nimbleFunction(
             }
         }
         model[[target]] <<- propValue
-        logMHR <- calculateDiff(model, calcNodes) + propLogScale
-        jump <- decide(logMHR)
-        if(jump) nimCopy(from = model, to = mvSaved, row = 1, nodes = calcNodes, logProb = TRUE)
-        else     nimCopy(from = mvSaved, to = model, row = 1, nodes = calcNodes, logProb = TRUE)
+        logMHR <- calculateDiff(model, target)
+        if(logMHR == -Inf) {
+            nimCopy(from = mvSaved, to = model, row = 1, nodes = target, logProb = TRUE)
```
danielturek: problem here.  doing to need deterministic dependents of `target`, too.

danielturek:
I left one comment in the code, about a nimCopy call only using target, which I think also needs deterministic dependents of target, too.

In the changes for RW_block, it looks like we're counting on the topological sorting in calcNodes.  I know that's the case, but also wonder (suspect?) if there could be some corner cases where it doesn't work .  But I read through pretty carefully, and agree with the logic.

packages/nimble/R/MCMC_samplers.R
perrydv: We discussed this and determined that it looks ok.

paciorek:
@danielturek In the 'various conjugacies setup' test in test-mcmc, how did you decide on the exactSample values? I need to update this because calcPriorFirst has changed the actual values (presumably because of the RW sampler for 'c', which presumably proposes a negative value early in the iterations).

danielturek:
@paciorek By running it and seeing what samples were generated

paciorek:
@danielturek I'm not seeing the point of having the 'exactSample' values then? I guess it is just to make sure that the MCMCs are same as we go through different versions of NIMBLE.

danielturek:
@paciorek That's correct.  Just in case things change, in any part of the (quite lengthy) pipeline...

paciorek:
@perry, in `decide()` in calcPriorFirst, I think the "junk" line is a legacy of trying to achieve comparable RNG to devel? I'm planning to remove it...

``
function (logMetropolisRatio) 
{
    if (is.na(logMetropolisRatio)) {
        junk <- runif(1, 0, 1)
        return(FALSE)
    }
    if (logMetropolisRatio > 0) 
        return(TRUE)
    if (runif(1, 0, 1) < exp(logMetropolisRatio)) 
        return(TRUE)
    return(FALSE)
}
```

paciorek:
@perrydv Just noticed that your code comment says it should be removed, so I'll just take care of it.

paciorek:
Weird - no conflict from command line on my machine , but conflict showing here. Resolved via browser-based editor - it was just some spacing differences. 

Checking differences in tests compared to new gold files done in fix_gold_check branch and just merged into devel. Hopefully will see that the only failed testing is from known differences in exact MCMC output in oxford, seeds, and a few other MCMC tests.

PR #779, opened 2018-08-14
addSampler method now takes ... arguments as control list elements

danielturek:
For MCMC configuration objects, the `addSampler` method now has a `...` argument.

Named arguments passed via `...` will be used as control list elements.

That is, when you used to do this:
```
conf$addSampler("x", "RW", control = list(logScale = TRUE, adaptInterval = 100))
```
you can now do this:
```
conf$addSampler("x", "RW", logScale = TRUE, adaptInterval = 100)
```

Creating a PR to trigger testing.


danielturek:
@perrydv @paciorek Passed tests.  I think this is sufficiently simple to be safe to merge in; but would either of you like to take a look, first?

danielturek:
@perrydv @paciorek Would either of you like to look at this before I merge it?

PR #780, opened 2018-08-14
change a min() to max() in test-trunc

perrydv:
@paciorek I think this two-character change in test-trunc fixes a typo.  First there is a check that the min of an MCMC sample is greater than the truncation lower bound.  Then there is what looks like it should check that the max of the sample is less than the truncation upper bound.  But it was using min instead of max.  Please confirm.


paciorek:
LGTM

PR #782, opened 2018-08-23
HMC sampler

danielturek:
Creating a PR to merge the HMC and langevin samplers into branch `ADcleanup`.  I don't think any of the AD work is ready to merge into devel yet, right?

The HMC sampler is at a good point, I believe:

- It implements HMC-NUTS sampling, which automatically chooses the HMC step size and number of leapfrog steps.
- It automatically transforms parameters for sampling on an unconstrained space.  Specifically, parameters restricted to (0, Inf) are log-transformed, and parameters bounded to an interval [a, b] use a shifted-scaled-logit transform (that is, shift and scale to the interval [0, 1], then logit-transform).  Also, note, unconstrained parameters on (-Inf, Inf) don't requite any transformation.
- There is not currently support for parameters restricted to (a, Inf), or to (-Inf, 0), or (-Inf, a).  Those cases could be added, but I don't think they ever come up.
- The HMC sampler has special handling for multivariate nodes.  It will handle `dmnorm` nodes once those derivatives are available, and it will also be able to handle `dwish` nodes, but that will take a little bit more work to get working.
- I did a fair bit of testing, and a lot of interesting numerical issues (e.g. NaN values) can arise during the course of sampling.  So I added a number of error traps throughout, to catch and handle these gracefully during the course of sampling (without erroring out).

I've done a lot of testing, and the results generally agree with other sampling methods.  The only exception, which I find quite interesting, is on Perry's election88 example.  Perry, I'll encourage you to try it out.

- MCMC1: default MCMC on election88
- MCMC2: remove the default samplers and add an HMC sampler on the top-level params (b.0, b.female, b.black, b.female.black, b.v.prev, sigma.age, sigma.edu, sigma.age.edu, sigma.state, sigma.region), then leave the default RW or conjugate samplers on everything else.  The posteriors are almost the same, except for two parameters (b.0, b.v.prev) that have noticeably different posterior distributions.  I don't think this is an error in the HMC sampler, I actually believe it's sampling correctly, but giving generally a different shaped posterior for these two parameters, only.  @perrydv I'll suggest you take a look at this and see what you think.

Anyway, I'm opening this PR, and putting this out there for code review and constructive feedback.


paciorek:
I'll comment in a separate post on high-level thoughts about our HMC compared to Stan's. I also plan to do a low-level code review.

This comment regards computational efficiency. 

I don't see any obvious places to wring speedups out of the current HMC code. With elections88, it appears that essentially all of the cost is the call to nimDerivs. And @danielturek has already cached values of the gradient where possible.

For getting a bit of speedup on larger models, it might be worth doing:

- in `jacobian`, avoid use of `grad` and just work directly with `derivsOutput$jacobian[1, 1:d]`
- don't loop over `1:d` in `jacobian` or `logH` or the two transformation functions. Instead define `anyLog`, `anyLogit` and `logPositions` and `logitPositions`. Then only if any log values are present, just loop over `logPositions`. Similarly for any logit values.

paciorek:
A few high-level thoughts on what we may want to do given Stan's implementation:

One thing we probably want to tackle even for the first release of HMC is to deal with warmup/burning: Stan estimates the step size during warmup but then discards those samples. Our current code has a phase in which we estimate step size as specified in the NUTS algo, but at the moment, users might simply use samples from that phase. Let's discuss how to either warn users or discard warmup samples.

A couple things we probably want soon-ish, but perhaps not while our HMC is EXPERIMENTAL:

- I think we want to estimate and report divergences, as Stan does, as well as provide tuneable parameters for reducing divergences, as they do
- Set a max tree depth
- Provide users diagnostic info such as number of leapfrog steps and tree depth by iteration.

At some point, we probably want to provide the ability to have a diagonal mass matrix that is estimated during warmup. Stan defaults to a diagonal mass matrix while currently we use an identity mass matrix. 

Also, I have some possibly less promising news regarding timing than I reported in a meeting back in May.  On elections88, the timing for 500 sampling iterations of Stan with default diagonal mass matrix is actually about the same as for 500 sampling iterations of NIMBLE HMC with a unit mass matrix.
However for apples-apples comparison, Stan with a (non-default) unit mass matrix is about 3x faster than NIMBLE. For election88, I think that the Stan diagonal mass matrix chain is 3x slower than Stan unit mass matrix chain because the tree depths are larger and the number of leapfrog steps much greater. So for fair comparison we'd need to collate info on the tree depths and number of leapfrog steps with our unit mass matrix approach. And we may need to set the max tree depth to 10 as Stan does.

All that being said, note that for elections88 for the sigma parameters, mixing for Stan with either mass matrix and for NIMBLE HMC are about the same - ESS of around 50 based on 500 iterations. And it may well be that NIMBLE speed will increase if we set a max tree depth.

Side note: I see hints in the online forums that one can get the estimated mass matrix from a Stan run (perhaps only with cmdstan?), but so far have not figured out how to do that or seen specific documentation on how the mass matrix is estimated.  The most info is in the Stan reference manual Sec. 14.2.

paciorek:
Also, so we have a record of it here, per email in summer 2018: 
a Bob Carpenter blog post says Stan uses the leapfrog (Störmer-Verlet) integrator for NUTS.

paciorek:
Ok, I looked through the HMC sampler code in detail. Generally LGTM. Comments:

- your comment of `alternate: -2*log(1+exp(-x))-x` should have `+x` not `-x` as the last term, I think.
- I think the code assumes the bounds for the logit transform don't change, but this isn't necessarily true for all models, so we will probably need some sort of check that bounds are not stochastic.
- I changed the condition in while loop of Algo 4 to be done on log scale (see latest commit)
- can we name it 'transformModelValues' not 'transformedModelValues' so it uses a verb?
- at the end of initializeEpsilon do we need to make sure all calcNodes (specifically deterministic deps of targetNodes) are reset to their original value?
- I could be wrong here but I think from a mathematical perspective, the 'jacobian' function should be named 'gradient' as it's a gradient not a Jacobian.
- How about we have gamma, t0, kappa, and delta be variables rather than hard-coded, for clarity?

I haven't wrapped my head around the caching of the jacL and jacR values...

paciorek:
Key points from 2019-09-18 dev meeting:

High priorities for HMC are: 
  - set max_treedepth and allow user to modify it (and ideally report how often it is reached). One important reason is so we can assess speed on an apples-apples basis.
  - figure out how Stan decides number of warm-up steps (@paciorek to look into) and mimic that
  - set up HMC (possibly requiring modifying MCMC system) to dynamically set warm-up based on user's burnin and niterations, and warn if burnin is less than whatever warmup ends up being used
  - report number of divergences

Perry suggested we could think about modifying our AD functionality so the calling function allocates space for derivatives rather than having that constantly allocated/deallocated in the derivs code.

We agreed that Chris' idea of modifying the looping over the log/logit transforms is probably not important and would muddy the code.

paciorek:
@danielturek  here's the discussion of warm-up in the Stan manual (https://mc-stan.org/docs/2_20/reference-manual/hmc-algorithm-parameters.html). It's more complicated than is discussed in 'help(stan)', where it states simply that warm-up defaults to half the number of iterations. 

I think for now we would be ok with allowing more flexible warm-up length and avoid the question of using a non-unit mass matrix and how to estimate that matrix.

Another note is the default acceptance rate goal of delta is 0.8 in Stan (adapt_delta). This also argues for having our 0.65 change to a parameter a user can control.

paciorek:
More notes -- on a basic hierarchical model (I can share as needed - it has a linear regression in the mean and a linear regression in the log variance), NIMBLE HMC takes much longer than default Stan (100-fold). I think this is because NIMBLE uses unit mass matrix and Stan a diagonal. When I run Stan with unit mass matrix, the speed advantage is reduced to 10x faster than NIMBLE and it reports one should increase max_treedepth.  If I then set max_treedepth to 50 (which is very large I think and presumably comparable to NIMBLE's lack of a max depth), then Stan is 2x faster than NIMBLE. So it appears that on a pure computational basis we are close to competitive but the details of the algorithm make a huge difference. 

So one the one hand, we may be fine with our speed, but on the other, the user experience may be that NIMBLE seems a lot slower. Let's see what happens when we add max_treedepth with default of 10, and then discuss the diagonal mass matrix question.

danielturek:
Changes made to HMC sampler:

- Added `maxTreeDepth` control argument to control maximum descent search tree depth (default = 10).
- Made threshold for a divergent path `deltaMax` into a control list argument, which user can set.
- Added message to report divergent paths during HMC sampling.
- Preallocated vectors for basically everything possible: `q`, `qL`, `qR`, `qNew`, `qDiff`, `p`, `pL`, `pR`, `p2`, `p3`, `grad`, and `gradFirst`.
- Changed names from "jacobian" to "gradient" in HMC sampler, to better represent these are gradients.  This includes both function and variable names.
- Added a check for "constant" bounds of distributions, when doing logit transformation in HMC sampler.  Throws an error if distribution has stochastic bounds.
- Changed function name `transformedModelValues` to `transformValues`, to parallel the `inverseTransformValues` method.
- Restored all `calcNode` values at the end of `initializeEpsilon` method, in particular the deterministic dependents of the `target` nodes was missing.
- Made HMC step-size adaptation parameters `delta`, `t0`, `kappa`, and `gamma` into control list arguments, which user can set.  Also added Roxygen documentation for these.
- Added more testing for `HMC` sampler, including exact samples for various `maxTreeDepth` values, and errors for non-constant distribution bounds.
- Improved clarity of warning messages in `HMC` sampler.
- Made `initialEpsilon` a control list argument of `HMC` sampler, so users can set the initial step-size value.

danielturek:
Still on my HMC to-do list:

1. Report total number of divergent paths, at the end of sampling.
1. Report how many times `maxTreeDepth` was reached, at the end of sampling.
1. Dynamically set warmup number of iterations, based on MCMC iterations and burnin.

The way I imagine it working, these last changes will require multiple-inheritance of NFs, something I'm in touch with @perrydv about.

paciorek:
Let's discuss:

1) whether to have default `delta` be 0.8
2) what causes the `NaN` values in leapfrog and in initializing step-size and if there is a different way we want to report this to user
3) unit versus diagonal mass matrix. Stan's sampling using the diagonal matrix is much faster on one example I was working with, presumably because smaller trees can be used with diagonal matrix.

Also a note for posterity that all three items on Daniel's list from Oct 1 appear to be done.

danielturek:
@paciorek 

Yes, all those items on my previous list were completed on a train ride or two.

I'm fine with any default value for `delta`.  I got the current default of 0.65 from the Hoffman & Gelman paper I've been using.  Does Stan use the default value of 0.8 ?

In terms of the "diagonal mass matrix", would that mean that each `p[i]` is no longer sampled i.i.d. from a N(0, 1)?  I'm sorry that I don't fully understand this, but could you explain just what is meant, or point me to documentation on this, the "diagonal mass matrix" ?

Issue #790, opened 2018-09-30
error in dcat range checking in case with indexing


paciorek:
Example:

```
code <- nimbleCode({
    for(i in 3:4) {
        z[i-1] ~ dcat(alpha[z[i-2], 1:2])
    }
}
)

m <- nimbleModel(code, inits = list(alpha = matrix(c(0.9,.1,0.1,.9), 2),
                                    z = c(1,2,1)), calculate = FALSE)
m$nodeFunctions[[1]]$calculate
```

```
defining model...
building model...
setting data and initial values...
running calculate on model (any error reports that follow may simply reflect missing values in model variables) ... Error in envRefInferField(x, what, getClass(class(x)), selfEnv) : 
  ‘i’ is not a valid field or method name for reference class “code_MID_12_modelClass_UID_102_UID_103”
```
```
   if (model$z[model$i - 2] >= 1 & model$z[model$i - 2] <= 2L) 
        model$logProb_z[getNodeFunctionIndexedInfo(INDEXEDNODEINFO_, 
            2L)] <<- dcat(model$z[getNodeFunctionIndexedInfo(INDEXEDNODEINFO_, 
            2L)], prob = model$alpha[model$z[getNodeFunctionIndexedInfo(INDEXEDNODEINFO_, 
            3L)], 1:2], log = 1)
```

Note use of `model$i`.

paciorek:
Fixed by PR #791.

PR #791, opened 2018-09-30
use symbolicParentNodesReplaced in findDynamicIndexParticipants

paciorek:
This should fix issue #790.

Still need to think through whether there are any cases in which we would want to use `symbolicParentNodes` rather than `symbolicParentNodesReplaced` or whether this was just an oversight originally. I believe the only use of the `dynamicIndexInfo` created in `findDynamicIndexParticipants` is for range checking in dynamic index expressions, so should be ok.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -1833,8 +1833,8 @@ modelDefClass$methods(findDynamicIndexParticipants = function() {
         for(iDI in seq_along(declInfo)) {
             if(declInfo[[iDI]]$type == "unknownIndex") next
             declInfo[[iDI]]$dynamicIndexInfo <<- list()
-            for(iSPN in seq_along(declInfo[[iDI]]$symbolicParentNodes)) {
-                symbolicParent <- declInfo[[iDI]]$symbolicParentNodes[[iSPN]]
+            for(iSPN in seq_along(declInfo[[iDI]]$symbolicParentNodesReplaced)) {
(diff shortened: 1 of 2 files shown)


paciorek:
Need to create a test, probably using these cases:
```
# NOT OK  because of bug
code <- nimbleCode({
    for(i in 3:4) {
        z[i-1] ~ dcat(alpha[z[i-2], 1:2])
    }
}
)

m <- nimbleModel(code, inits = list(alpha = matrix(c(0.9,.1,0.1,.9), 2),
                                    z = c(1,2,1)), calculate = TRUE)

# OK
code <- nimbleCode({
    for(i in 1:2) {
        z[i+1] ~ dcat(alpha[z[i], 1:2])
    }
}
)

m <- nimbleModel(code, inits = list(alpha = matrix(c(0.9,.1,0.1,.9), 2),
                                    z = c(1,2,1)), calculate = FALSE)

# OK, but "by luck" because it relies on subtracting 1 in the dynamic index check (getNodeFunctionIndexedInfo(INDEXEDNODEINFO_, 1L) - 1) rather than using  getNodeFunctionIndexedInfo(INDEXEDNODEINFO_, 2L)

code <- nimbleCode({
    for(i in 2:3) {
        z[i] ~ dcat(alpha[z[i-1], 1:2])
    }
}
)


m <- nimbleModel(code, inits = list(alpha = matrix(c(0.9,.1,0.1,.9), 2),
                                    z = c(1,2,1)), calculate = FALSE)
```

Issue #793, opened 2018-10-04
turn on parallel Eigen


paciorek:
We should do this for next release.

paciorek:
In the parallel_eigen branch, I've inserted `-fopenmp` in `nimble/inst/make/*.in`.

I don't think we need to insert `-fopenmp` in `src/Makevars*in` or `inst/include/CppCode/Makevars` as I don't think we have any Eigen-based linear algebra in our hard-coded C++, but @perrydv let me know if you have other thoughts.

paciorek:
One other comment is that Eigen's parallelization seems limited. As of Eigen 3.3.5:

```
Currently, the following algorithms can make use of multi-threading:

    general dense matrix - matrix products
    PartialPivLU
    row-major-sparse * dense vector/matrix products
    ConjugateGradient with Lower|Upper as the UpLo template parameter.
    BiCGSTAB with a row-major sparse matrix format.
    LeastSquaresConjugateGradient
```

So turning on parallelization will probably not help end users by and large.

paciorek:
Testing indicates an issue with `-fopenmp` on macOS. 

More info:
https://stackoverflow.com/questions/42280628/requiring-openmp-availability-for-use-in-an-rcpp-package/42281495#42281495

Also not clear that `-fopenmp` would by default be supported on Windows.

Writing R extensions section 1.2.11 indicates one should use SHLIB_OPENMP_CXXFLAGS in Makevars* files.

I've now implemented this, but on one of my Macs, -fopenmp is being injected despite seeming to not be supported. Perhaps this means R was configured incorrectly (though I think I did a basic R install on that machine). But we should be cautious here for fear of breaking users' use of nimble if they don't have openMP support.

Side note: users can also do this at run time if their compiler supports openMP:
```
Sys.setenv("PKG_CXXFLAGS"="-fopenmp")
Sys.setenv("PKG_LIBS"="-fopenmp")
```

Cautionary note from Writing R extensions: Note that the macro SHLIB_OPENMP_CXXFLAGS applies to the default C++ compiler and not necessarily to the C++11/14/17 compiler: users of the latter should do their own configure checks (an example is available in CRAN package ARTP2).

paciorek:
Ok, so this is failing on two separate Macs. -fopenmp is being injected into the clang++ call and that is erroring out with an "unsupported option '-fopenmp'" error. 

[...]

PR #802, opened 2018-11-12
add -fopenmp to parallelize eigen

paciorek:
addresses issue #793


paciorek:
Failed tests seem spurious. Build issue for linux test case 6 and for mac, test-compareMCMCs fails to build shared library but runs locally. (But note that running on a Mac seems to produce 'unknown warning option' output even though we should/have been suppressing that. Also getting weird errors when run test-compareMCMCs.R on 0.6-12 on a Mac. Arggh.

paciorek:
Need to reconsider the approach so closing this.

PR #814, opened 2018-11-25
fix CRP conjugacy detection when have truncation

paciorek:
If have fewer clusters than obs, the presence of 'nodes' like `muTilde[11:20]` was causing a bug in checkCRPconjugacy. Added code to omit `muTilde[11:20]` in cases like this.


paciorek:
removing as I want to merge this into BNP as it was branched from BNP.

PR #815, opened 2018-11-25
fix CRP conj detection for truncated case and non-exch tilde node priors

paciorek:


paciorek:
closing so I can reopen to force Travis build

PR #818, opened 2018-12-01
adding detection of normal-invgamma conjugacy

paciorek:
For testing only. Do not merge.

This is for BNP only as we do not have a conjugate sampler for standard models.


paciorek:
whoops, merging in reverse direction of what I want.

PR #819, opened 2018-12-01
added conjugacy checking for dnorm_invgamma

paciorek:
For testing only.

This adds conj detection for normal-invgamma but for use only in BNP models.


paciorek:
This did not trigger travis. Will try again.

Issue #795, opened 2018-10-17
Incorrect identification of end nodes


aaronspringford:
I'm new to nimble, so this may be an issue with syntax or something else, but I'm having difficulty with a model in which the likelihood depends on a stochastic vector. In this case, the first element of the stochastic vector is correctly identified as *not* being an end node, but the remainder of the stochastic vector is incorrectly identified as end nodes. Here is a (dumb) simplified example:

```r
library(nimble)

model <- nimbleCode({
  p ~ dbeta(1,1)
  q1 ~ dbeta(1,1)
  q2 ~ dbeta(1,1)
  
  for( i in 1:10 ){
    a[i] ~ dbern(0.3)
    pp[1:3] <- nimC( a[i]*(1-p-q1) + (1-a[i])*(1-q2), a[i]*q1 + (1-a[i])*q2, a[i]*p )
    y[i] ~ dcat( pp[1:3] )
  }
})

model_data <- list( y = sample( 1:3, 10, replace = TRUE, prob = c(0.2, 0.3, 0.5) ) )
mod <- nimbleModel( code = model, data = model_data )
mod$isEndNode("a")

 a[1]  a[2]  a[3]  a[4]  a[5]  a[6]  a[7]  a[8]  a[9] a[10] 
FALSE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE  TRUE 
```
I'm running Windows R-3.4.2 with nimble v. 0.6-12
Thanks.

aaronspringford:
I think what I need to do here is have a matrix for `pp` i.e. `pp[1:3,1:10]`

paciorek:
Hi Aaron,

A fundamental issue here is that your code tries to redefine pp[1:3] multiple times (once for each iteration of the loop), but BUGS is a declarative language, so pp[1:3] should only appear on the left hand side once.

We should probably produce an error or warning when you try to build this model because it is not BUGS syntax that we want to allow, but what is happening is that pp[1:3] is being defined for i=1 and then our processing does not redefine pp[1:3], so a[2], ..., a[10] have no dependents. 

I don't know exactly what you are trying to do, but I could imagine it would be natural to have some thing like:
```
pp[1:3,i] <- nimC( a[i]*(1-p-q1) + (1-a[i])*(1-q2), a[i]*q1 + (1-a[i])*q2, a[i]*p )
y[i] ~ dcat(pp[1:3, i])
```
[...]

PR #822, opened 2018-12-01
added warning about multiply-defined nodes in genExpandedNode...

paciorek:
This addresses issue #795, namely a user not having complete indexing for a LHS expression within a loop, e.g.,
```
for(i in 1:3) {
   mu ~ dnorm(0,1)
   for(j in 1:3)
     mu0[i,1] ~ dnorm(0,1)
}
```

@perrydv and @danielturek please take a quick look as this check will be done on all models. I think it's basically what we want as I don't think any cases of redefining a LHS are standard use of NIMBLE, but open to thoughts.


danielturek:
Restarted travis build.  Hopefully the failure was a fluke

danielturek:
Changes look good.

PR #824, opened 2018-12-11
allows user to request error messages after compileNimble fails

paciorek:
This addresses NCT issue 100. 

@danielturek @perrydv if one of you has a chance, please run the following on a Windows machine and report the results: 
```
a = nimbleFunction(
    run = function(x=double(1)) {
        for(i in 1:3)
                x[i-1] = x[i]*'a'
        returnType(double(1))
        return(x)
    })
ca = compileNimble(a, showCompilerOutput=TRUE)
ca = compileNimble(a, showCompilerOutput=FALSE)
printErrors()
printErrors(excludeWarnings = FALSE)
```

### DIFF
--- a/packages/nimble/R/all_utils.R
+++ b/packages/nimble/R/all_utils.R
@@ -181,3 +181,20 @@ createNamedObjectsFromList <- function(lst, writeToFile = NULL, envir = parent.f
     }
 }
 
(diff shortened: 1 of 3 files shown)


paciorek:
@perrydv tried on Windows and things looked good.

PR #833, opened 2018-12-19
more flexible cross-validation

perrydv:
The purpose of this pull request is to trigger testing on a branch with an experimental feature to allow more flexibility in `runCrossValidate`.

There is a new argument to `runCrossValidate` called `lossFunctionIsSimple`, which defaults to `TRUE` to maintain current behavior. If it is set to `FALSE`, then the `lossFunction` argument should be a function that takes as arguments the compiled MCMC, the compiled model, and the `leaveOutNodes`, which is a vector of node names for the held-out data for one CV fold (similar to what is returned from the `foldFunction`).  The `lossFunction` should return a single number that is the cross-validation score for the CV fold.  The MCMC sample can be extracted by `as.matrix(CMCMC$mvSamples)`, where `CMCMC` is the compiled MCMC object.

Help pages have not been updated because the design of this idea may evolve.

### DIFF
--- a/packages/nimble/R/crossValidation.R
+++ b/packages/nimble/R/crossValidation.R
@@ -27,7 +27,8 @@ calcCrossVal <- function(i,
                          nburnin,
                          returnSamples,
                          nBootReps, 
-                         silent){
+                         silent,
+                         lossFunctionIsSimple = TRUE){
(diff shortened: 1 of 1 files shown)


paciorek:
Closing due to staleness.

Issue #835, opened 2018-12-20
Another typo in NimArr.h


danielturek:
Line 1022 of NimArr.h

Should be:
size1 * size2 * size3 * size4 ????

danielturek:
@perrydv

danielturek:
@perrydv @paciorek Raising these for attention again, hopefully before the release.  (2/2)

danielturek:
Fixed by #836

Issue #834, opened 2018-12-20
Typo in NimArr.h


danielturek:
On line 888 of NimArr.h

3 should be a 4 ??

Need @perrydv to look at, confirm, or disagree, and/or fix.

danielturek:
@perrydv @paciorek Raising these for attention again, hopefully before the release.  (1/2)

danielturek:
Fixed by #836

PR #836, opened 2018-12-20
added case for 5-dimensional nimArray to NimArr.h

danielturek:
Opening PR so I can carefully look over the code changes

### DIFF
--- a/packages/nimble/inst/include/nimble/NimArr.h
+++ b/packages/nimble/inst/include/nimble/NimArr.h
@@ -1389,7 +1389,7 @@ class NimArr<5, T> : public NimArrBase<T> {
     NimArrBase<T>::NAlength = size1 * size2 * size3 * size4 * size5;
   }
 
-  NimArr<4, T>(int is1, int is2, int is3, int is4, int is5) : NimArrBase<T>() {
+  NimArr<5, T>(int is1, int is2, int is3, int is4, int is5) : NimArrBase<T>() {
     setSize(is1, is2, is3, is4, is5);
   }
 
--- a/packages/nimble/inst/include/nimble/NimArr.h
+++ b/packages/nimble/inst/include/nimble/NimArr.h
@@ -1047,6 +1047,7 @@ class NimArr<4, T> : public NimArrBase<T> {
     NimArrBase<T>::NAdims[1] = size2 = is2;
     NimArrBase<T>::NAdims[2] = size3 = is3;
     NimArrBase<T>::NAdims[3] = size4 = is4;
+
     NimArrBase<T>::NAstrides[0] = NimArrBase<T>::stride1 = str1;
     NimArrBase<T>::NAstrides[1] = stride2 = str2;
     NimArrBase<T>::NAstrides[2] = stride3 = str3;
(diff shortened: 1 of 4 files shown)


danielturek:
This PR appears to successfully enable the use of 5-dimensional arrays in model code, through addition of the new templated class `NimArr<5, T>` to `NimArr.h`.

It has passed my testing.  But a few comments:
1. I suspect I put the test in the **wrong test file**.  *Please advise which file this new test actually belongs in.*
2. If you feel the test I added is not rigorous, in particular there's some aspect(s) of the 5-d nimArray code that's not exercised at all, then please let me know how the test should be extended.

As for this PR, I will wait for approval and comments from @perrydv before merging it.

danielturek:
Found one error.  Fixed.

danielturek:
@perrydv Testing passed, so I'm leaving this for you to look at.

perrydv:
Hi @danielturek, @paciorek.  Sorry this escaped my attention.

The code looks right on first pass.  I suggest more extensive testing.  This could include:

- Making a map that will be implemented as a lower-dimensional NimArr.  This would arise from e.g. `my5d[2:4, 3:6, 2, 4, 1:3]`.  This needs to copy successfully, assign and be assigned to successfully, and be passed as an argument successfully.  Also the case with `drop = FALSE` should be checked.  Having the first index be one of the scalars is important because in that case the first stride is not 1.
- Making a map that can be eigenized, like `my5d[2, 3, 4, 2:4, 3:5] %*% b[1:3]`.  Cases of assigning from, assigning to, and passing as an argument would also be important. 
- Making some nested indexing cases, such as `my5d[3, 2:4, 3:6, 2:5, 3:7][2:3, 2:4, 1, 1] <- someMatrix[1:2, 1:3]`.  Cases of assigning from, assigning to, and passing as an argument would also be important. 
- Having expressions in indices, such as `my5d[2:4, 3:6, 2, 4, indexVector + 1]`.
- Use in a model.

danielturek:
@perrydv Which test file do the tests for this belong in?

perrydv:
@danielturek test-coreR.R.  See `blockTests` and `indexChainTests`.

danielturek:
@perrydv I've managed to be quite unsuccessful in using the testing system in `test-coreR.R`.  I tried adding one and only one test to the `blockTests`, and cannot get it to work.

Any help, please, understanding if I'm using the testing system wrong, or perhaps the implementation of 5D nimArrays is incorrect?

From the test I added, the final one now appearing in `blockTests`, I get the following error:

```
> blockTestsResults <- test_coreRfeature_batch(blockTests, 'blockTests') ##lapply(blockTests, test_coreRfeature)
compiling... this may take a minute. Use 'showCompilerOutput = TRUE' to see C++ compilation details.
Error : Cannot do math with arrays that have more than 2 dimensions.
 This occurred for: out
 This was part of the call:  out <- map("temp",3,call,list,list)
Error: There is some problem at the Eigen processing step for this code:
 {
    {
        temp <- arg1
        out <- temp[2:4, 3:6, 2, 4, 1:3]
    }
    return(out)
}
```

danielturek:
Note, I'm expecting the testing to fail.....

perrydv:
OK, sorry this slipped by me for a while.

I've pushed:

1. Corrections to both the bugs you found, Issue #834 and #835.  Only for #834 could I actually trigger an error from the bug.
2. Correction to one typo in a `setSize` method for the 5d case in NimArr.h
3. A set of tests is `test-coreR.R`.

A reminder about what works and not for dimensions > 2:

If x is 3-dimensional, then `y <- x` works, but `y <- x[1:2, 1:2, 1:2]` does not work.  That is why your test failed.  `y <- x[2, 1:2, 1:2]` works because the result is <3-dimensional.

Somewhat awkwardly, the following does work:
```r
copy3d <- nimbleFunction(
    run = function(x = double(3)) {
        ans <- x
        return(ans)
        returnType(double(3))
    })
foo <- nimbleFunction(
    run = function(x = double(3)) {
        ans <- copy3d(x[2:5, 2:4, 3:6])
        return(ans)
        returnType(double(3))
    })
```

In other words, `ans <- x` works even when `x` is passed by slicing an object, but `ans <- <slice of x>` does not work directly.  Of course this could be fixed, but that is the current status.

danielturek:
@perrydv Two of the travis test batches appear to have failed, but I'm guessing(?) they're just flukes.  In any case I'm restarting those two batches right now.

danielturek:
@perrydv @paciorek Testing passed.  Changes, including the two bug fixes I found, and the one you found in my code, all look good.  Same goes for the extensive new tests you added.

Merging this in.

PR #838, opened 2019-01-03
wrap BNP cluster samplers so only used if cluster has observations

paciorek:
Do not merge: for @danielturek to review @paciorek approach  to not wrapping MCMC samplers for BNP cluster parameters when a given cluster has no constituent observations.

This work builds off @perrydv reversible jump sampler, extending it to wrap whatever sampler is assigned to the cluster parameter by configureMCMC.


paciorek:
arggh - other changes beyond what I did for BNP MCMC. Need to merge into BNP branch probably. Will retry...

PR #840, opened 2019-01-03
add ddexp/dlaplace distribution

paciorek:
Per user request fall 2018.


paciorek:
@danielturek @perrydv feel free to review this if you have an interest. I use {location,scale} as our canonical parameterization. In the C functions I use R's dexp distribution, which requires that I sometimes go back and forth between the log and non-logged scales for pdexp and qdexp - a bit awkward and slightly non-ideal for numerical accuracy but I preferred to use the existing distribution rather than coding up the CDF/quantile calculations, though that would be easy enough.

Many of the files changed related to adding `inherits=FALSE` in mid-December - this is already in devel and is in ddexp branch so unclear why the merge diff is showing this...

danielturek:
@paciorek I've looked over, and noted the addition.  Thanks for flagging for attention.

That is curious about the `diff` results, and `inherits = FALSE`.  I took a quick look, and it seems wrong to me, as well....

PR #842, opened 2019-01-12
address rchk warnings about PROTECT patterns

perrydv:
rchk now issues warnings about PROTECT code patterns.  This PR attempts to resolve all warnings.

In our code, the majority of the warnings appear to be superfluous, but it is worth changing the usage pattern/style to satisfy rchk and conform to the recommended practices.

There were also cases where there might be a real issue, but no effort was made to determine if that was the case since it was easy to simply fix the usages.

### DIFF
--- a/packages/nimble/R/cppDefs_core.R
+++ b/packages/nimble/R/cppDefs_core.R
@@ -172,14 +172,14 @@ cppClassDef <- setRefClass('cppClassDef',
                                                         Sderived = cppSEXP(name = 'Sderived_EXTPTR'))
                                    CBobjectDefs <- c(CBobjectDefs, baseClassPtrObjectDefs)
                                    newCodeLine <- cppLiteral(c(paste0('newObj = new ', name,';'),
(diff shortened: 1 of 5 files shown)


paciorek:
I tried to run rchk via this mechanism in a Docker container:
```
https://github.com/nfultz/notes/issues/9
```

```
cd /opt/R-svn
bin/R -e "install.packages('randomizr',repos=c('https:/r.declaredesign.org/', 'https://cloud.r-project.org'))" --slave
. $RCHK/scripts/cmpconfig.inc
$RCHK/scripts/check_package.sh nimble
```

(To use the fix-PROTECT-warnings branch, I need to use `R CMD INSTALL nimble_0.6.13.tar.gz` after copying the tarball into the container.)

Unfortunately I am getting:
```
docker@91cce8bc7b67:/opt/R-svn$ $RCHK/scripts/check_package.sh nimble
/opt/rchk/scripts/check_package.sh: line 77:  5095 Segmentation fault      (core dumped) $RCHK/src/$T ./src/main/R.bin.bc $F > $FOUT 2>&1
```

Not sure what is going on. Will plan to merge this in and submit to CRAN, noting to them that we think we've fixed the issues that rchk flagged.

PR #845, opened 2019-01-25
make sampler_categorical and sampler_binary use a prior/dependencies split in calculations

perrydv:
This could speed up cases where there are categories that have 0 prior probability.

### DIFF
--- a/packages/nimble/R/MCMC_samplers.R
+++ b/packages/nimble/R/MCMC_samplers.R
@@ -97,30 +97,33 @@ sampler_categorical <- nimbleFunction(
         ## numeric value generation
         k <- length(model$getParam(target, 'prob'))
         probs <- numeric(k)
+        logProbs <- numeric(k)
         ## checks
         if(length(targetAsScalar) > 1)  stop('cannot use categorical sampler on more than one target node')
         if(model$getDistribution(target) != 'dcat') stop('can only use categorical sampler on node with dcat distribution')
     },
     run = function() {
         currentValue <- model[[target]]
-        probs[currentValue] <<- exp(getLogProb(model, calcNodes))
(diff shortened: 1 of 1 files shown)


paciorek:
LGTM. 

One question though - do we want to check for all -Infs as the log probs and take a uniform sample. The change to the code still leaves in place that rcat(1, rep(0,k)) produces a 1 every time. That said I'm not sure what the correct thing to do in this improper situation is. Maybe always having a 1 is good in that a user in some chains might see all 1s. But if the underflow only occurs sometimes, one gets an extra chunk of mass piling on 1. 

I also looked for other cases where underflow might occur and don't see any. We do have a similar exponentiation of log probs in sampler_CRP, but we are already subtracting off the max logProb.

@perrydv in your mind ok to merge and me to proceed with release; presumably once we make a decision about all -Infs case.

paciorek:
Hit return too quickly. Given we are subtracting max log prob, I'm not sure when we would ever get all -Infs. Unless somehow all the categories are impossible.

perrydv:
I thought about that case too, but wasn't sure what to do.  I agree it represents an invalid model state that should never be obtained in a healthy scenario.  I guess could we make the new value be a NaN or NA in order to quickly show there is a problem, but we haven't done that kind of thing in the past.  We also haven't built in error-trapped bail-outs from MCMCs based on invalid values, maybe because there is a chance it occurs during first iterations and will recover.  @paciorek I'm ok with however you want to handle it.

paciorek:
I like the NaN/NA idea in principle but agree that if we haven't done before then doesn't make sense to do just here. Offhand what are the other similar cases of invalidity that are occurring to you.

perrydv:
I don't have a particular one in mind.  Unless the model is badly initialized or another sampler has done something wrong, at the least the current state should have a valid logProb value.

PR #846, opened 2019-01-27
add conjugacy for regression models with `inprod` and `%*%` for univariate priors

paciorek:
For testing only.

This finally adds the conjugacy detection for such common cases:

```
y[i] ~ dnorm(inprod(x[i,1:p],beta[1:p], 1)
beta[i] ~ dnorm(0,1)
```

```
y[i] ~ dnorm((X[i,1:p]%*%beta[1:p])[1,1], 1)
beta[i] ~ dnorm(0,1)
```

@danielturek no hurry at all, but at your convenience please review my extension of the "linear" conjugacy to handle these cases. In particular see how I handled the `structureExpr` and note that the `offset` and `coeff` I produce are NOT valid code, but since we don't use those to actually generate code, I think that is ok.  See the new tests in test-mcmc.R to see what cases  I am handling and what the result of ``cc_checkLinearity` is for these.

This addresss NCT issue 56.


danielturek:
@paciorek I just restarted the (failed) travis test batch 4, in case the failure was a fluke.

danielturek:
@paciorek Looks like this is legitimately failing a test that has to do with dynamic indexing.

danielturek:
@paciorek Overall, LGTM.  A few comments of varying importance:

Shall we remove the line:
```
if(exists('paciorek')) browser()
```

Are we concerned about also catching (identical) cases that are written as:
```
y[i] ~ dnorm(b0 + sum(beta[1:p] * X[i, 1:p]), 1)
```

Lines 1152 and 1153, something concerns me about using the nested list indexing with `expr`, namely `expr[[2]][[1]]` and similarly `expr[[3]][[1]]`, without first checking that second level of list-indexing exists, and also the potential for the common R warning about "the condition has length > 1 and only the first element will be used", if it's the case that `expr[[2]][[1]] is itself an expression of some sort.

That said, I tried to come up with an example of this causing a problem and I couldn't so, I can't say for sure it's a problem.  I guess someone will tell us if it causes a problem ever, and we can fix it then.

Also, I see and I'm ok with the changes to how the results of `cc_checkLinearity` look in some cases, in particular, no longer being valid expressions.

paciorek:
Ok, in lines 1152-3, I added a check for `length(expr[[2]])` and same for `length(expr([[3]])`, though it seems the only way to not have the current code be ok would be if we were multiplying inputs that are not indexed, which I don't think is possible.

Good catch on the `sum()`. I've added code to handle that case.

danielturek:
@paciorek I restarted failed test batch 5.

danielturek:
@paciorek I fixed a typo in the recent changes, changing `len(...)` into `length(...)`.

danielturek:
@paciorek I cleaned up the new tests you added, the issue was that `constants` was being redefined randomly, so the samples didn't agree.  I also cleaned up extra definitions of `inits`.

danielturek:
@paciorek Tests passed, I've reviewed this, LGTM.

Thanks for this improvement to the conjugacy detection.

PR #849, opened 2019-01-29
change initializeModel approach to be faster

perrydv:
We fixed a bug in initializeModel in way that slowed it down substantially.  This PR attempts to speed it up.

### DIFF
--- a/packages/nimble/R/initializeModel.R
+++ b/packages/nimble/R/initializeModel.R
@@ -27,11 +27,10 @@
 #'       ....
 #'    }
 #' )
-#' @export
+##' @export
 initializeModel <- nimbleFunction(
     name = 'initializeModel',
     setup = function(model, silent = FALSE) {
-        
         initFunctionList <- nimbleFunctionList(nodeInit_virtual)
         startInd <- 1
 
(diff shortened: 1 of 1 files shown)


danielturek:
@perrydv Nice approach to this

perrydv:
Thanks @danielturek.
@paciorek @danielturek it looks like tests passed.  (I had to restart one where it looked like travis stopped before getting far).  We could push farther to speed up the steps of `isData` and `getNodeNames()`, which might be faster if we go more directly to the source of information.  Or we could go with this as is.  What do you think?

paciorek:
Not sure. Wary of rabbit holes but if fairly easy to do, might be worthwhile. We have a bit of time before the Feb. 8 deadline for next release, but we should try to keep things moving along expeditiously.

perrydv:
Ok, I am trying some ways to improve it by using some internal objects more directly.  If this fails, we can revert to what worked.

danielturek:
@perrydv Tbh unless the speedups are noticeable from the use of internal objects, the previous clarity of implementation using `isData`, `getNodeNames`, etc, definitely carries some weight for me.  But I will quickly concede that if the speedups are noticeable, then the most recent changes are fine.

Could I suggest the following?  Keeping as commented code what the new changes are actually doing? For example:

```
#####
## the same as:
## RHSonlyNodes <- model$getNodeNames(RHSonly = TRUE)
allNodes <- model$modelDef$maps$graphID_2_nodeName
graphIDs <- seq_along(allNodes)
allTypes <- model$modelDef$maps$types
bool <- allTypes == 'RHSonly' | allTypes == 'stoch' | allTypes == 'determ'
allNodes <- allNodes[bool]
allTypes <- allTypes[bool]
graphIDs <- graphIDs[bool]
bool <- allTypes == 'RHSonly'
RHSonlyNodes <- allNodes[bool]
#####
```

PR #853, opened 2019-02-07
change makeParsedVarList to avoid protect stack overflow

### LINKED ISSUE #852: protection stack overflow in a large model in 0.7.0
A user reported an inexplicable error from a model that worked in 0.6-12.

The error was "There an unknown problem looking for variables..." [sic].  This comes from `parseEvalNumericManyHandleError`, which is the error handler for `parseEvalNumericMany` and `parseEvalNumericManyList`.

It turns out that the C function `makeParsedVarList` has `PROTECT`s in a for loop and then a single `UNPROTECT` at the end.  For an input vector of length > 150000, this apparently triggered the "protection stack overflow".   I think handling of vector of such length probably arises from the new `initializeModel` (a good thing because it is more vectorized!) but I haven't confirmed that.

I will make a Pull Request changing `makeParsedVarList` to `UNPROTECT` within the for loop, so that the protect stack never grows so large.

perrydv:
Fixes #852.

See Issue #852 for description.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -2905,7 +2905,7 @@ parseEvalNumericManyFindErrors <- function(x, env) {
 
 parseEvalNumericManyHandleError <- function(cond, x, env) {
     problems <- parseEvalNumericManyFindErrors(x, env)
-    if(length(problems)==0) message(paste0('There an unknown problem looking for variables ', paste0(x, collapse=','), ' in the model.\n'))
+    if(length(problems)==0) message(paste0('There is an unknown problem looking for variables ', paste0(x, collapse=','), ' in the model.\n'))
     else {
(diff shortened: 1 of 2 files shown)


perrydv:
@paciorek Could you look at the test failures on this PR?  They seem to have to do with the new set of correct test results and I'm wondering if that could be the issue.

paciorek:
Ok will do.

Also, we may not want to merge this into devel, depending on how we want to handle getting this out to users. So let's discuss that before any merging...

paciorek:
Test failures were spurious.

I think we should keep this fix distinct from devel (but of course we can/should merge into devel) so that users can pick up these fixes without getting whatever else we are currently working on in devel.

paciorek:
I am not yet deleting this branch (until 0.7.1 is live) so that any users still seeing our convo on nimble-users can access it.

PR #855, opened 2019-02-09
avoid sampling parameters of empty clusters in BNP

paciorek:
branch `nosample_empty_clusters2` replaces `nosample_empty_clusters` and this PR replaces PR #839 

@danielturek please take a look at the modifications and let me know of any other thoughts.


danielturek:
@paciorek Looks like tests are failing, related to `helperFunctions` or similar.

I'm guessing it will be a quick and easy fix (typo or similar), but I'll leave it for you.

paciorek:
One note for future reference - when this is implemented (and in combination with v 0.7.0 changes to sampler_CRP), for non-conjugate samplers we now propose new values of the cluster parameter, and if they are rejected the proposed values are maintained in the MCMC output. Now they are parameter values for clusters with no observations so they are in any real sense not meaningful, but it is different than having the cluster parameters stay constant until the cluster actually has observations.

paciorek:
Actually regarding the note above for future reference, I implemented code where we put back the old values if the new cluster is not occupied.

PR #860, opened 2019-02-20
updated initializeModel() to always re-calculate deterministic nodes

danielturek:
Per discussion.

Removed full `model$calculate()` call at the end of `initializeModel()`.

Replaced with so that the initialization routine for deterministic nodes **always** recalculates the value of the node (in topological order), to ensure the value is not only a "valid" number, but also correctly represents the upstream state of the model.

Also added a test, to check for this correct behaviour.


paciorek:
You have two identical 'if' statements in the initialize deterministic node function ;)

danielturek:
@paciorek It never hurts the check things twice, in immediate succession, in case the result changes somehow

paciorek:
True, but then probably best to check three times.

danielturek:
Tests passed, merging this in.

Issue #859, opened 2019-02-18
bug in BNP sampler_CRP introduced in 0.7.0


paciorek:
In setting up sampler_CRP to also sample the cluster parameters of new clusters, I forgot to include the cluster parameters as part of the nodes that need to be updated in mvSaved. Result is that for the non-conjugate sampler that mvSaved has and incorrect value and in correct logProb when subsequent MCMC sampling outside of sampler_CRP is done. This affects 0.7.0 but not earlier releases as in <=0.6-12 we did not update the cluster parameters for new clusters for the nonconjugate case within sampler_CRP. 

This could also affect the conjugate sampler in corner cases where a user doesn't use the default conjugate sampler for the cluster parameters (by not using the default samplers or by adding things like block samplers). This would affect the conjugate samplers in 0.7.0 and in 0.6-12, but I doubt this is something users would be doing with the cluster parameters.

We should fix this ASAP and make clear in the 0.7.1 release that BNP users should upgrade and redo any MCMCs done with 0.7.0.

paciorek:
Fixed by hotfix_bnp_update.

PR #861, opened 2019-02-23
Fixes issue #859

paciorek:
Do not merge. For testing only.


paciorek:
Also fixes bug in findClusterNodes that wasn't detecting cases like
```
dnorm(mu[xi[i]], sd = sigma[xi[i]])
```
when the 2nd use of `xi[i]` was not lifted.  Only clear case where this would have caused a problem was when ignoring 2nd use would have made us detect conjugacy based only on the first use of `xi[i]`.

paciorek:
Not deleting branch until 0.7.1 is live as we have pointed to this branch on nimble-users.

PR #866, opened 2019-03-07
added besselK to nimbleOrRfunctionNames

### LINKED ISSUE #863: use of besselK in BUGS model code doesn't work
The `besselK` function doesn't appear to work in BUGS models, unless the arguments are static.

Comments on the matter from Perry:

It looks like since it is not a nimbleFunction (if I follow, we rely on R's besselK for uncompiled execution, and we convert to bessel_k for compiled execution), it needs to be added to nimbleOrRfunctionNames in BUGS_BUGSdecl.R

Reproducible example:

```
library(nimble)

code <- nimbleCode({
    a <- besselK(1, 1)
    b <- 1
    c <- besselK(b, 1)
})

Rmodel <- nimbleModel(code)
## defining model...
## Error in getSymbolicParentNodesRecurse(code, constNames, indexNames,
## nimbleFunctionNames,  : 
##  R function 'besselK' has arguments that cannot be evaluated; either the function must be a
## nimbleFunction or values for the following inputs must be specified as constants in the model: b.
```

danielturek:
This fixes (allows) the use of the `besselK` function inside BUGS model code.

Fixes #863.

Note, the `besselK` function still suffers from the inconsistency of `dim=2` **array arguments** giving rise to a `dim=1` **vector output**.


danielturek:
I think this one is pretty safe.  Testing passed.  Merging in.

Issue #858, opened 2019-02-18
rpath has hard-coded path


paciorek:
We are now failing a CRAN install, because of a new check they are in the process of putting in place.
Here is info on the new check: https://developer.r-project.org/Blog/public/2019/02/14/staged-install/

Apparently they want packages to be installable such that they can be moved after installation. Since we hard code in the installation directory via -rpath (only on Linux not MacOS nor Windows), this is causing an issue.  The link above mentions we could try to opt out. However first we should:

1) discuss (remind) why we link libnimble.so into nimble.so and whether we need this
2) see if we can have rpath use a relative path

Here is the failure on Brian Ripley's machine:

```
* installing *source* package ‘nimble’ ...
** package ‘nimble’ successfully unpacked and MD5 sums checked
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether clang++ -std=gnu++11 accepts -g... yes
working on eigen
checking how to run the C++ preprocessor... clang++ -std=gnu++11 -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking Eigen/Dense usability... no
checking Eigen/Dense presence... no
checking for Eigen/Dense... no
Using the version of Eigen provided with NIMBLE
NIMBLE_DIR = /data/gannet/ripley/R/packages/tests-clang/nimble.Rcheck/00LOCK-nimble/00new/nimble
ENABLE_LIB=true
configure: creating ./config.status
config.status: creating inst/make/Makevars
config.status: creating inst/make/Makevars_lib
config.status: creating R/config.R
config.status: creating src/Makevars
removing debugging flags
make[1]: Entering directory '/data/gannet/ripley/R/packages/tests-clang/nimble/inst/CppCode'
/data/gannet/ripley/R/R-clang/bin/Rscript --vanilla ./customizeMakeconf.R
make[1]: Leaving directory '/data/gannet/ripley/R/packages/tests-clang/nimble/inst/CppCode'
[...]

PR #867, opened 2019-03-09
avoid use of rpath

paciorek:
This fixes CRAN issue #858


paciorek:
Ok, so even after taking our rpath, we are still failing staged installation.
From the info from CRAN, it appears the other main cause of failure is hard-coded paths from use of `system.file` being run at install time. 

In `cppDefs_utils.R` we have:
```
NimbleCodeDir = system.file("CppCode", package = "nimble")
```
but `NimbleCodeDir` is never used elsewhere. @perrydv I'm going to comment this out, but let me know if that raises any flags for you.

paciorek:
Ok, one more wrinkle. We are using -rpath in our linking at run time when calling compileNimble. E.g.
```
g++ -std=gnu++11 -shared -L/usr/lib/R/lib -Wl,-Bsymbolic-functions -Wl,-z,relro -o P_21_code_MID_1_03_09_12_56_03.so P_21_code_MID_1.o P_21_code_MID_1_nfCode.o -L/tmp/nim_devel/nimble/CppCode -lnimble -Wl,-rpath=/tmp/nim_devel/nimble/CppCode -llapack -lblas -L/usr/lib/R/lib -lR
compilation finished.
```
This seems necessary as otherwise the .so's we create can't find libnimble.so at run time (e.g., when `dyn.load` is called. 

However, we are currently hard-coding a path in via RPATH into `make/Makevars_lib` via configure substitution on `make/Makevars_lib.in`. 

I think we need to do this at run-time via createMakevars().

In createMakevars(), it looks like we use make/Makevars_lib if we can. Not sure exactly when we can and can't and not sure if there would be repercussions from either not using it or modifying it to put RPATH in during createMakevars().

@perrydv if you do have a moment over the next couple hours, might be helpful to walk through with you, but I will work my way through it regardless.

perrydv:
Hi @paciorek I'm around if you want to get on zoom or something.

It's not clear to me that a run-time rpath issue would trigger CRAN checking issues if those are checking on build-time.

paciorek:
Per discussion with @perrydv we need to also modify NIMBLE_DIR and NIMBLE_INC_DIR via createMakevars and not configure. 

Also we should check using R-devel that if do install from source and enable-dylib=true on Mac that things are fine.

PR #868, opened 2019-03-09
allow 'x' not 'value' in 'types'

paciorek:
when using registerDistributions.


danielturek:
@paciorek So to be clear, we now allow users to specify eitehr `x` or `value` in `registerDistributions`, and either will work?

And, all the information in the `distributions_inputList.R` file, is still in terms of `value`, right?

paciorek:
Yes in the 'types' argument you can use either 'x' or 'value', but everything in the distributions object is still named 'value'. If one uses things like `getDimension` you need to use `value`, but there will be an informative message if you use `x`.

paciorek:
@perrydv you were complaining about unhelpful manual and man info on registerDistributions. I looked and in the manual there is a line about `types` being needed in special cases and in `help(registerDistributions)` there is a paragraph about `types` that does say that one specifies the type for the random variable using `value`. 

So if my change to allow 'x' instead of 'value' in the 'types' argument is not enough, please let me know of a more specific change to the manual or roxygen.

danielturek:
@paciorek That all makes sense.

PR #870, opened 2019-03-14
add more PROTECTs to C++ code generated for a nimbleRcall

### LINKED ISSUE #869: Inadequate use of PROTECT in nimbleRcall generated code
`nimbleRcall` can yield unexpected behavior in which variables do not appear to be properly copied from C++ to R for the call of an R function.  This occurs due to insufficient use of PROTECT in the generated C++.  The reproducible example we have is quite large, so it will not be included in this issue.

This bug would be more likely to manifest if the R function called by `nimbleRcall` does extensive processing (and hence the garbage collector is likely to be more active) and less likely to occur if it is fairly simple.

perrydv:
fixes #869 

This pull request inserts additional uses of `PROTECT` into C++ generated for a `nimbleRcall`.  There is a C++ function generated that creates an R expression object (`LANGSXP`) and populates it before calling the R evaluator.  Arguments to it are `SEXP` objects copied from nimble's internal C++ objects (`NimArr`).  Those needed `PROTECT` statements, as did the object returned from R before being copied into a nimble internal object.

### DIFF
--- a/packages/nimble/R/externalCalls.R
+++ b/packages/nimble/R/externalCalls.R
@@ -197,7 +197,7 @@ nimbleRcall <- function(prototype, returnType, Rfun, envir = .GlobalEnv) {
         thisNdim <- thisSymbol$nDim
         SEXPargName <- paste0(argNames[i], '_SEXP_')
         replacedArgNames[i] <- SEXPargName
-        newConvertLine <- substitute( A <- NimArr_2_SEXP(B), list(A = as.name(SEXPargName), B = as.name(argNames[i]))) ## MakeSEXPcopyLineFrom almost does this but it is at cppDef level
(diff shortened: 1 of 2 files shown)


danielturek:
@perrydv Confirming that this fixes the error / bug I was experiencing

danielturek:
@perrydv Happy to have this merged into devel, whenever you deem it ready.

paciorek:
@perrydv @danielturek do NOT merge until I give the go-ahead. I haven't yet tagged v0.7.1 nor merged devel into master.

paciorek:
Ok, devel is unfrozen - feel free to merge when you want.

danielturek:
@perrydv Is this ok to merge in?  I'm keen to, if you think it's ready.  We have the "go ahead" from Chris.

PR #871, opened 2019-03-22
CAR skipExpansionsNode

danielturek:
Do not merge.

Opening PR to trigger testing

### DIFF
--- a/packages/nimble/R/CAR.R
+++ b/packages/nimble/R/CAR.R
@@ -84,17 +84,17 @@ as.carCM <- function(adj, weights, num) {
 
 
 ## specialized conjugacy-checking of the scalar component nodes of dcar_normal() AND dcar_proper() distributions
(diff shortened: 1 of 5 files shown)


danielturek:
Also added a test to catch this case

danielturek:
I believe this fixes things up.  Reviews welcomed from @perrydv, but otherwise I'll probably merge this in on Monday.

I came to terms with the `skipExpansions` system part of conjugacy checking, and extended the system (`skipExpansiosNode`) so that it only affects the particular multivariate node in question - fixing what was causing the problem here.

I also added a test to catch this case, and make sure it's working correctly.

perrydv:
Hi @danielturek.  Thanks for jumping into this.  It is in some tricky processing that I don't know well, so let's go with your dive into it.  It passes tests and you added a new test.  LGTM.

danielturek:
Tried to remove some redundant processing from conjugacy checking

PR #872, opened 2019-04-11
added only.values=TRUE to CAR calls to eigen()

danielturek:
I think this is all fine.

Leaving this for @paciorek to merge in, before the next release.


danielturek:
I'm just merging it, tests have passed and I believe this is what we want here.

Issue #879, opened 2019-05-06
model$setInits(inits) gives terrible error with unnamed list elements


danielturek:
The title says it all.  This error should probably be caught somewhere (most likely in `model$setInits()`, and given a helpful error message.

Minimal example below:

```
library(nimble)

code <- nimbleCode({ a ~ dnorm(0, 1) })

inits <- list(a = 0, 1)   ## inits list has an unnamed element

Rmodel <- nimbleModel(code)

Rmodel$setInits(inits)
## Error: attempt to use zero-length variable name
```

PR #880, opened 2019-05-09
model$setInits() now checks for unnamed elements

danielturek:
Address #879 

`model$setInits` now checks for un-named list elements.


danielturek:
Testing passed.  I also did a fair bit of testing (using different `inits` lists) myself.

Merging this in.

PR #882, opened 2019-05-17
remove tensorflow support

paciorek:
For testing only, do not merge.

This excises tensorflow support given we haven't developed it and given the nCompiler rewrite.


paciorek:
Note `appveyor.yml` still has some modifications because of difficulties caused by tensorflow. Not clear whether we should remove those modifications.

perrydv:
Thanks for tackling this @paciorek .  The changes make sense.  The test failures look like they might be just residual typos in install_requirements and/or run_tests.

perrydv:
@paciorek I think include/base64 was only for tf and so can be removed, and the corresponding copyright entry.

PR #883, opened 2019-05-17
add check for missing returnType in user-defined dists

paciorek:
This traps a previously uncaught error that resulted in an unhelpful R error message.


perrydv:
👍 on the code, but I haven't looked at the test failures.

paciorek:
*&^$#$^&  this is failing because test-mcmc.R is producing an extra NULL in the goldFile. Not clear why this would be caused by changes in this branch and when I run test-mcmc under current devel, I also get the extra NULL. Not sure what I will do next...

paciorek:
Might have had something to do with nested testthat calls though still not clear why we weren't getting an extra NULL before this branch. In any event, a tweak to test-mcmc.R for the beetles model seems to have fixed things.

PR #885, opened 2019-05-18
reject proposed 0s in sampler_RW_dirichlet

paciorek:
This rejects when the scalar proposal for an element is exactly zero (which can happen because of underflow). 

For testing only; we may wait until also get feedback from Jean-Baptiste.


danielturek:
@paciorek Looks straightforward, and correct.  That said, I didn't do any actual testing.

PR #886, opened 2019-05-20
Fix copy aliasing bugs

perrydv:
This PR fixes some bugs with copying when the LHS and RHS involve the same variable.  There were two sources of bugs:

- Lack of an Eigen `.eval()` being inserted in some cases.
- Lack of an intermediate variable being inserted during size processing in some cases.

Examples of the bugs now fixed are in the `aliasTests` list added to `test-coreR.R`


perrydv:
@paciorek @danielturek The copying bugs I emailed you about were a bit bigger than I first realized.  This PR fixes them.

paciorek:
Ready to merge in?

perrydv:
I think so.

Issue #876, opened 2019-04-19
set enclosing environment of RCfunctions based on getNimbleFunctionEnvironment()


paciorek:
@danielturek is developing a package and when he tries to use one RCfunction within another RCfunction, R CMD check complains like this:

```
* checking R code for possible problems ... NOTE
calculateU_ns: no visible global function definition for ‘nsCrosscorr’
calculateU_ns: no visible global function definition for ‘nsCorr’
```

Apparently R CMD check evaluates the code in the package R files. This evaluation creates a set of functions. For each of these given functions, for each function call in the body of  the function, it looks through the enclosing environments of the given function to try to find the functions that are called.

When an RCfunction is created, nimbleFunction() calls RCfunction() and ultimately the RCfunction is created in a new environment by nfMethodRC$generateFunctionObject(). Because the RCfunction is defined within nimble package code, one of the RCfunction's enclosing environments is the nimble namespace. (There are a couple intervening enclosing environments - those of generateFunctionObject and of nfMethodRCclass.)

E.g., just for RCfunction created in a plain R session:
```
parent.env(parent.env(parent.env(environment(myRCfunction))))
<environment: namespace:nimble>
```

The upshot is that the RCfunction that is created in a package (that is not 'nimble' itself) does not have the package namespace in its enclosing environments. So R CMD check cannot find functions that are from that package namespace. 

I believe the solution is to pass "where" into RCfunction and on into nfMethodRC$generateFunctionObject. There we should do this:

```
environment(ans) <- new.env(parent = where)
```

I think this would be fine to do for RCfunctions created within the nimble package itself as well. But we should check carefully the series of enclosing environments that are currently being set for RCfunctions in the nimble namespace. Currently I see:
```
 parent.env(parent.env(parent.env(environment(nimble:::CAR_calcCmatrix))))
<environment: namespace:nimble>
```

and the proposal would eliminate a couple of these intervening enclosing environments.

Also we might as well have the name of the RCfunction make use of `where` as we do for nimbleFunctions. So we may want to pass in `where` into rcFunLabelMaker.

Notes:
1) because nimbleFunctions with setup code produce generators and there is no run code in the body of the generator function, this issue hasn't arisen for nimbleFunctions. Namely R CMD check never even looks at the $run code.
[...]

PR #889, opened 2019-05-21
Rc scoping

paciorek:
This adds scoping for RCfunctions and for the RCfunctions created by nimbleRcall and nimbleExternalCall using the `getNimbleFunctionEnvironment` that we introduced a while back.

This allows other packages to depend on nimble and pass CRAN checks as it places nfs and RCfunctions created as part of the package in the correct namespaces.

This fixes issue #876


paciorek:
@danielturek I believe BayesNSGP R CMD check should fully pass now.

PR #891, opened 2019-05-22
MCMC configuration print method handles samplerConf objects in control lists

danielturek:
Updates how mcmc configuration objects are printed (using `printSamplers()`), so that when a `samplerConf` object is in the control list (as will happen in the new `RJ` work), the printing still looks nice.


danielturek:
I've tried this out using at @salleuska 's example RJ code, and it works as desired.

Assuming testing passes, I'll merge this into devel.

danielturek:
Testing passed.  Merging in.

PR #895, opened 2019-05-24
new MCMC option: MCMCmonitorAllSampledNodes

danielturek:

Adds a new package MCMC option `MCMCmonitorAllSampledNodes` (default FALSE).

When TRUE, by default the MCMC will monitor all latent states that are being sampled, in addition to the top level nodes (the usual default behaviour).


danielturek:
Testing passed, merging this in.

Issue #875, opened 2019-04-19
untrapped error using TRUE/FALSE in model code


paciorek:
```
 code = nimbleCode({
     a <- pnorm(-2,0,1, lower.tail=FALSE)
 })
 m = nimbleModel(code)
defining model...
Error in replaceConstantsRecurse(x, constEnv, constNames) : 
  Error, hit end
```

paciorek:
The same error occurs with this, although here it is debatable if we want TRUE converted to 1 as a model parameter...

```
 code = nimbleCode({
      a ~ dnorm(TRUE, 1)
 })
m = nimbleModel(code)
defining model...
Error in replaceConstantsRecurse(x, constEnv, constNames) : 
  Error, hit end

paciorek:
Fixed by PR #904

PR #904, opened 2019-05-28
allow logical values in model code

paciorek:
This fixes issue #875 modifies: replaceConstantsRecurse, genReplacementsAndCodeRecurse, getSymbolicVariablesRecurse, so that logical values will pass through, treated
as numeric values are. 

@perrydv would be great if you take a look in case you think of any cases this would cause problems in for compilation. 

This PR allows this code  to work (instead of requiring that users say `lower.tail=0` and returning an unhelpful error message if they don't (see issue #875 ).
```
code = nimbleCode({
   y ~ dnorm(0,1)
   a <- pnorm(y, mu, 1, lower.tail=FALSE)
   mu ~ dnorm(0, 1)
 })
m = nimbleModel(code, data = list(y=0), inits = list(mu = 1))
cm = compileNimble(m)
cm$a
```

Note that in the process I also modified what happens when TRUE/FALSE is used as a parameter value. Previously that was erroring out as:
```
code = nimbleCode({
     a ~ dnorm(TRUE, 1)
 })
m = nimbleModel(code)
Error in replaceConstantsRecurse(x, constEnv, constNames) : 
  Error, hit end
```

Now we trap it in in `isExprLiftable`:
```
code = nimbleCode({
      a ~ dnorm(TRUE, 1)
  })
 m = nimbleModel(code)
defining model...
Error in isExprLiftable(paramExpr, types[[paramName]]) : 
  isExprLiftable: NIMBLE is not expecting a logical/boolean value; please use a numeric value in place of TRUE.
```

### DIFF
--- a/packages/nimble/R/BUGS_BUGSdecl.R
+++ b/packages/nimble/R/BUGS_BUGSdecl.R
@@ -658,7 +658,7 @@ getSymbolicParentNodesRecurse <- function(code, constNames = list(), indexNames 
     ##                indexing, not dynamic indexing
     ## - hasIndex: is there an index inside
     ## numeric constant
-    if(is.numeric(code) ||
+    if(is.numeric(code) || is.logical(code) || 
        (nimbleOptions()$allowDynamicIndexing &&
                        length(code) > 1 &&
                        code[[1]] == ".DYN_INDEXED")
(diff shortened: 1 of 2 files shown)


perrydv:
@paciorek I guess this is ok.  Checking for a logical in `isExprLiftable` might be confusing in the future.  It looks like it was a convenient place to insert a check, but maybe not the most logical place?

paciorek:
@perrydv agreed -- I only put it there because that's where it now errors out once logicals are allowed in replaceConstantsRecurse (where it was formerly erroring out with a confusing R error).

perrydv:
Sounds good.  Happy to go with what works and not insert some separate layer for this.

PR #907, opened 2019-05-29
Faster getDependencyPaths

perrydv:
getDependencyPaths is the slowest step of conjugacy checking.  I wrote a C++ version that works on a set of classic BUGS models and makes configureMCMC faster by model-specific factors, typically around 2x.  This PR will check whether the new system matches the current system across the wider range of models in our test suite.


perrydv:
This passed tests with nimbleOption `oldConjugacyChecking = FALSE`, i.e. with the new system.

Per discussion, we decided to merge this with default to the old system.  I will push a change to make `oldConjugacyChecking = TRUE` be the default and see if that passes tests, i.e. if we didn't break the old system.

paciorek:
@perrydv ok to merge in?

PR #909, opened 2019-05-29
revert addMonitors default behavior to what it has been

perrydv:
Per discussion this morning, we will revert to previous (until a recent commit) default behavior for addMonitors in configureMCMC.

This will benefit from efficiency improvement in model$getNodeNames.  In the future, we may add back this includeData = FALSE flag.

It is only relevant if a model has a top-level data node, which is not typical.


danielturek:
@perrydv As discussed, LGTM

PR #911, opened 2019-05-30
Add {d,p,q,r}logis recycing rule implementation

jpdunc23:
Includes tests for each version in `test-coreR.R`.


perrydv:
@paciorek What do you think about this.  @jpdunc23 noticed that dlogis etc. lacked compiled recycling rule behavior.  Do you want to add this for 0.8.0 or hold off since there is a lot going in?

paciorek:
@perrydv I'm fine with adding it.

PR #920, opened 2019-06-30
insert isTRUE in calc/sim/calcDiff to avoid uncompiled errors with truncation/dyn idx

paciorek:
FRO-DNM

This inserts isTRUE in calculate/simulate/calculateDiff nodefunctions so that under truncation (if a node value is NA) or dyn idx (if an index is NA) users don't see untrapped R errors from `if(NA)`. 

This also creates an `isTRUE` identity function in our C++ code. Presumably this will have negligible effect on compiled run times but worth discussion.

Before merging, check line 17 of nimbleFunction_nodeFunction to see why there is an unexpected change relative to devel.

### DIFF
--- a/packages/nimble/R/nimbleFunction_nodeFunction.R
+++ b/packages/nimble/R/nimbleFunction_nodeFunction.R
@@ -14,7 +14,7 @@ ndf_createDetermSimulate <- function(LHS, RHS, dynamicIndexLimitsExpr, RHSnonRep
         ## error gracefully if dynamic index too small or large; we don't catch non-integers within the bounds though
         if(is.null(nodeDim)) {
             nanExpr <- NaN
-        } else nanExpr <- substitute(nimArray(NaN, LENGTH),
+        } else nanExpr <- substitute(nimRep(NaN, LENGTH),
                                      list(LENGTH = prod(nodeDim)))
         code <- substitute(if(isTRUE(CONDITION)) LHS <<- RHS
                            else {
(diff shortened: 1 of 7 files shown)


paciorek:
The line 17 change was inexplicable, but I've fixed it so we use `nimRep` not `nimArray`.

paciorek:
@perrydv  can you take a quick look here? C++ changes are trivial, but just want to make sure I'm not missing something.

packages/nimble/R/nimbleFunction_nodeFunction.R
```diff
@@ -272,17 +272,17 @@ ndf_createStochCalculateTrunc <- function(logProbNodeExpr, LHS, RHS, diff = FALS
                                                )), list( e = substCode)))
         } else {
             if(discrete && lower != -Inf) {
-                substCode <- quote(if(CONDITION) {
-                                       if(LOWER <= VALUE & VALUE <= UPPER)
+                substCode <- quote(if(TRUE(CONDITION)) {
```
perrydv: @paciorek Is this one a typo?

packages/nimble/inst/include/nimble/Utils.h
```diff
@@ -165,6 +165,8 @@ void nimStop();
 
 bool nimNot(bool x);
 
+bool isTRUE(bool x);
```
perrydv: @paciorek I suggest in the .h:
static inline bool isTRUE(bool x) {return(x);}
and then nothing is needed in the .cpp and the compiler should inline the call.

paciorek:
Waiting on Travis; then will merge.

Issue #924, opened 2019-07-19
bug in crossLevel sampler that causes incorrect behavior for non-identity links


paciorek:
There appears to be a long-standing bug in the cross-level sampler (it was there in 0.6-10, possibly as long ago as when the cross-level sampler was written, unless we didn't have non-identity links at the point that it was written). The bug causes incorrect behavior for non-identity link functions. I believe that the resulting sampler behavior would generally be so egregiously bad that a user would not make use of the results. Hopefully that is the case.

The problem is in the run function of the crossLevel sampler, we call `lowConjugateGetLogDensityFunctions[[iSF]]$run`, which calls `conjugateSamplerFunction$getPosteriorLogDensity()`. In that function, we set `model[[target]]`to 1 and never set it back to its original value (see code below for an example with normal-normal conjugacy). As a result, the resulting posterior density that is used is incorrect (so the acceptance calculation is invalid) and the value of the dependent nodes (those nodes that are dependents of the top-level nodes that are 'officially' the ones being sampled) are set to 1. Note that in the step of actually sampling from the conjugate distribution, we sample reasonable values, but then there is another pass through `conjugateSamplerFunction$getPosteriorLogDensity()`  and the dependent nodes are set to 1.  So once a proposal is accepted, all the dependent nodes will be 1, and the MCMC goes haywire.

Also @danielturek - a side question -- crossLevel sampler is written for a block of (hyper)parameters. I presume it is fine for a scalar hyperparameter, even though the RW proposals for the (hyper)parameters are implicitly set up for a block of parameters?

```
Class method definition for method getPosteriorLogDensity()
function () 
{
    prior_mean <- model$getParam(target[1], "mean")
    prior_tau <- model$getParam(target[1], "tau")
    for (iDep in 1:N_dep_dnorm) {
        dep_dnorm_values[iDep] <<- model$getParam(dep_dnorm_nodeNames[iDep], 
            "value")
        dep_dnorm_tau[iDep] <<- model$getParam(dep_dnorm_nodeNames[iDep], 
            "tau")
    }
    model[[target]] <<- 0
    model$calculate(calcNodesDeterm)
    for (iDep in 1:N_dep_dnorm) dep_dnorm_offset[iDep] <<- model$getParam(dep_dnorm_nodeNames[iDep], 
        "mean")
    model[[target]] <<- 1
    model$calculate(calcNodesDeterm)
[...]

PR #925, opened 2019-07-19
getPosteriorLogDensity method of conjugate samplers doesn't modify target value

danielturek:
DO NOT MERGE.

Creating PR to trigger testing.

### DIFF
--- a/packages/nimble/R/MCMC_conjugacy.R
+++ b/packages/nimble/R/MCMC_conjugacy.R
@@ -667,9 +667,12 @@ conjugacyClass <- setRefClass(
 
             ## calculate and return the (log)density for the current value of target
             functionBody$addCode(posteriorObject$prePosteriorCodeBlock, quote = FALSE)
+            if(link != 'identity') {
+                functionBody$addCode({
+                    model[[target]] <<- origTargetValue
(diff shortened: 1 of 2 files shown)


danielturek:
This PR has passed testing, suggesting the conjugate samplers still function correctly with these changes.

Waiting for @paciorek to determine if this fixes the problem he observed (#924) in the `crossLevel` sampler.

paciorek:
Looks good, except I'll note that even in identity link cases we write back into the model even though `model[[target]]` has not changed. Is it worth checking and only doing this if necessary? Or best just to always do it?

paciorek:
@danielturek I'd like to merge this in soon. Can you look at the identity link case and see what you think vis-a-vis my comment above?

danielturek:
@paciorek Apologies that this slipped by me back in August.

I've just updated the `crossLevelFix` branch to reflect your suggestion: the conjugate sampler will not "restore" the value of the `target` node, nor call `model$calculate(...)`, when the link is `"identity"`, in which cases they should not have been modified.

Please take a look any time, and let me know if you agree.

Assuming so, I'm fine for you to merge this in.

PR #926, opened 2019-07-19
fixed corner case in checkConjugacy, when scale is identically zero

danielturek:
Fixes a corner case.

Also added testing.

I'll merge this into devel, once testing passes.


danielturek:
Tests passed, merging this into devel.

PR #929, opened 2019-07-29
add handling of `(expression)[indexing]` in addMissingIndexingRecurse

paciorek:
FRO-DNM

This looks for cases like `(x[1:2]%*%y[1:2])[1,1]` based on `(` and `[` and if found, checks no missing indices and then runs `addMissingIndexingRecurse` on the parenthesized expression being indexed.

Before finalizing this I would also like to resolve the question raised in NCT issue 141 regarding whether the stanza beginning line 562 of BUGS_modelDef.R that checks for `$` is needed.

Will need to add testing, including testing for correct trapping of missing indexing and vector version of this situation, namely something like `(y[1:2]*x[1:2])[2]`

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -566,7 +566,7 @@ addMissingIndexingRecurse <- function(code, dimensionsList) {
     ## handle cases like (x[1:2]%*%y[1:2, i])[1,1]
     if(length(code[[2]]) > 1 && code[[2]][[1]] == '(') {
         if(any(unlist(lapply(as.list(code[3:length(code)]), is.blank))))
-            stop(paste0('addMissingIndexingRecurse: the model definition includes the code ', deparse(code), ', which contains missing indices. When indexing expressions (as opposed to explicit variables), all indices must be provided.'), call. = FALSE)
+            # stop(paste0('addMissingIndexingRecurse: the model definition includes the code ', deparse(code), ', which contains missing indices. When indexing expressions (as opposed to explicit variables), all indices must be provided.'), call. = FALSE)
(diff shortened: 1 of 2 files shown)


paciorek:
One further note here - because of checking I added, this PR errors out for missing indexes of expressions, e.g.,
```
code = nimbleCode({
    mn[1:2] <- (X[1:2,1:2] %*% beta[1:2,1:2])[,1]
    y[1:2] ~ dmnorm( mn[1:2], pr[1:2,1:2])
})
m = nimbleModel(code, data = list(y = rnorm(2)),
                inits = list(X = matrix(1, 2,2), beta = matrix(2,2,2), pr = diag(2)))
cm <- compileNimble(m)
cm$calculate('y')
```

However, if I take that check out, compilation goes through and calculations are fine. So we may well not want to check for missing indexes, unless we think there are cases where missing indexing will cause problems.  Need input from @perrydv on what happens in terms of translation to Eigen code when indexes are missing.

paciorek:
- Fixed erroneous commenting of the check for missing indexes in ()[,].
- Recurse into the indexes of [,] when have things like `k[,1]`: `[k[,1],]`
- Added testing

Waiting on Travis.

paciorek:
Apparently here's the use case for needing to check $ (see line 1284 of test-nimbleLIst.R)

```
    modelCode1 <- nimbleCode({
        meanVec[1:5] <- eigen(constMat[,])$values[1:5]
        covMat[1:5,1:5] <- eigen(constMat[1:5,])$vectors[1:5,1:5]%*%t(eigen(constMat[1:5,1:5])$vectors[,])
        y[] ~ dmnorm(meanVec[], cov = covMat[1:5,1:5])
    })
```

Adding this back in and rerunning tests.

PR #938, opened 2019-10-09
added compression of scalar nodes for printSamplers(byType = TRUE)

danielturek:


danielturek:
Testing passed, merging this into devel.

Changes are only in the `conf$printSamplers(byType = TRUE)` cases.  Now, scalar nodes will be compressed, to look nicer.

PR #939, opened 2019-10-10
allow (scalar) multiplicative link in wish and invwish conjugacies

paciorek:
Do not merge: this is just the initial steps of allowing this and not functional yet.

For some functionality for BNP, we want to recognize conjugacies like this:
```
y[1:p] ~ dmnorm(z[1:p], cov = lambda * Sigma[1:p,1:p])
Sigma[1:p,1:p] ~ dinvwish(S[1:p,1:p, nu)
```
and similarly for dwish.

I've started on this by changing the invwish and wish links in MCMC_conjugacy.R to 'multiplicative' and adding 'coeff' to the posterior calculation but the heavier lift is to detect that in this situation the 'coeff' is scalar (unlike in the dmnorm linear link conjugacy case) and to calculate coeff based on the [1,1] elements of the dependent covariance and the value of the target node.

@danielturek let's discuss whether you or I should finish this off.

### DIFF
--- a/packages/nimble/R/MCMC_conjugacy.R
+++ b/packages/nimble/R/MCMC_conjugacy.R
@@ -146,7 +146,7 @@ conjugacyRelationshipsInputList <- list(
          ## changing to only use link='identity' case, since the link='linear' case was not correct.
          ## -DT March 2017
          ## link = 'linear',
-         link = 'identity',
+         link = 'multiplicative',
          dependents = list(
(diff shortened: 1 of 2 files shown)


paciorek:
Given devel had diverged a lot from this work, I made a new branch multivariate_conj (branching off bnp_moreGeneral2, which branched off devel recently) and moved this functionality into that branch.

Issue #935, opened 2019-10-04
IF2 inflexibility in what the latent nodes are


paciorek:
As written, IF2 allows a user to specify something like `alpha[2:20]` as the `nodes` even if 'alpha[1:100]` is defined in the model code, but then in the processing it decides that `alpha[1:100]` is the set of latent nodes. 

This could be confusing to the user and also prevents one from saying that `alpha[1]` is the baseline node. 

@perrydv let's discuss how much we should address this. For the moment I have put a cautionary note in the roxygen.

paciorek:
Also @perrydv  for testing purposes, do you have any code lying around where the latent states are vectors?

perrydv:
I think this is the issue that some of Nick's old code deals with, that is currently sitting on the side waiting to be worked back in.  I can point you to it if you want @paciorek.  No, I don't think I have testing code for vector latent states.

paciorek:
Yes please point me to that code.

perrydv:
I think it is in branch pfLatentSpecificationMods (and this is the only remaining piece we need from that branch).  The issue is relevant for any of the filtering algos, and this branch is pre-IF2, so look in the other filtering build*.R nimbleFunctions, e.g. buildAuxiliaryFilter.R.  There's some thick grepping  on the node names, so it might a bit of a dive to follow it.

paciorek:
Ok, I have code in a function called 'findLatentNodes' that modifies Nick's functionality (replacing grepping with manipulation of expressions) to handle cases like:

```
nodes = 'x'
nodes = 'x[2:4, 3:20]'
nodes = c('x[1]','x[2]',....,'x[20]')
nodes = c('x[,1]','x[,2]',....,'x[,20]')
```

It fills in missing indexes, determines time dimension and then expands so that there is one 'node' per time point (it might actually be a set of nodes at each time point).

Will embed use of this function in all PF algos and will modify roxygen for 'nodes' to make clear what specification is allowed.

paciorek:
Fixed in PR #940

PR #940, opened 2019-10-10
updates to PF

paciorek:
Addition of IF2 and various bug fixes/improvements to the filtering algos in light of JSS revision, including generalization of latent node detection, so this fixes issue #935.

For review/testing only; DNM.

@perrydv feel free to take a look when/if you have a chance. I plan to do additional review myself.


paciorek:
Need to deal with testing errors.

Check that dimensionality for `estimates`, `logLik` etc. work if only one param is being estimated.

paciorek:
@perrydv I may be misremembering but I thought we did not want to allow 'resample=TRUE' in the control list for the PF and PF_block PMCMC samplers. It looks like they are still there.

perrydv:
@paciorek I think you're right.  I think it was an oversight and we should remove `control$pfResample` and the local `resample` and fix relevant code to assume resample is FALSE.

paciorek:
@perrydv here's what I see in test-filtering:

1. the 'basic auxiliary' gives slightly different numerical results now (for some parameters).
2. the 'multivariate auxiliary' gives rather different numerical results (for some parameters; same for others)
3. Looks like you added {get,set}LastLogLik to auxiliary filter but not bootstrap. The PMCMC testing used the default bootstrap, so algo wasn't even running until I added that to bootstrap too. The PMCMC tests now seem fine.

For items 1 and 2 I didn't want to dive too deep until got your reaction. I didn't think you had modified the auxiliary filter algo, but rather that the modification was for the more complicated dependency situation, which shouldn't arise in the test case. But quick skim suggests you did change some lines of calculation in the auxiliary filter run code. Not sure if that would be expected to change numerical results.

paciorek:
I've dealt with 1-param IF2 case and with PFresample removal. Also added a basic IF2 test.

There is of course a lot more we could do in terms of adding tests...

paciorek:
Ok, so in the email exchange btw Nick and @perrydv in early August, the discussion was around a bug in the APF for `saveAll=FALSE`.

However, subsequent to the changes in the APF code, we are now seeing different numerical results in test_filtering.R, which is testing the `saveAll = TRUE` case.

I'm in the midst of reconciling the numerical differences via the uncompiled model...

paciorek:
@perrydv I now see that in the look-ahead step you changed the calculation (this is confirmed in an email exchange of you and Nick). So this is presumably the (or at least one) reason for the numerical changes in the auxiliary filter output for `saveAll=TRUE`. 

So I will regenerate the gold file. 

```
        ## Multiply by p(x_t+1 | x_t).
##        auxll[i] <- auxll[i] + model$calculate(calc_thisNode_self)
##        auxll[i] <- auxll[i]+calculate(model, thisNode) ## This line was included previously.  Removing it has only trivial impact on established test results.  Maybe try with more extensive test.
        ## Multiply by weight from time t.
```

perrydv:
@paiorek I am catching up on this.  What you've said makes sense.  I don't think we gave much attention to the vector case during the work in August, so that is more of an open question.

PR #942, opened 2019-10-11
updated assignment of data values to not overwrite data list

### LINKED ISSUE #941: Can't name a model data node as called `data`

danielturek:
@perrydv @paciorek 

Fixes #941

I made a change.  It's a one-liner, but make no mistake, this is **deeply rooted**.  If either of you guys are uncomfortable with this, I will not take offense.

It changes how the "data" variables are assigned into the model.  Changes the `assign()` statement to begin looking "up" one environment, so as not to over-write the local variables, in this case, the `data` list, but there's also the potential for other conflicts (with other local variables) depending on what users name their model (data) variables.

Anyway, if testing passes, then I'll have confidence in it, but you'd better both take a look, and please weigh in here.

### DIFF
--- a/packages/nimble/R/BUGS_model.R
+++ b/packages/nimble/R/BUGS_model.R
@@ -604,7 +604,11 @@ Details: If a provided value (or the current value in the model when only a name
                                               scalarize <- FALSE else scalarize <- TRUE  ## if non-scalar, check actual dimensionality of input
                                           if(length(nimbleInternalFunctions$dimOrLength(varValue, scalarize = scalarize)) != length(isDataVars[[varName]]))   stop(paste0('incorrect size or dim in data: ', varName))
                                           if(!(all(nimbleInternalFunctions$dimOrLength(varValue, scalarize = scalarize) == isDataVars[[varName]])))   stop(paste0('incorrect size or dim in data: ', varName))
-                                          assign(varName, varValue, inherits = TRUE)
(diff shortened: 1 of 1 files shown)


paciorek:
I'm confused how it ever worked, as it looks like we were assigning the data value into an object in the local frame of setData() rather than into the model itself. I must be missing something.

danielturek:
@paciorek IICU `assign` would look upward, and find the existence of the variable (say `x`), and assign into there.

But in the case of assigning into `data`, it finds the local variable first (before starting any search upwards through enclosing environments), and thus overwrites the local `data` variable (formerly).

Now, `assign` begins is search in the correct environment, that one which contains the correct `x` (or `data`) variable.

paciorek:
@danielturek thanks for explanation - I guess this behavior is specific RefClasses, since if one just has a function, assignment will create a new object in the local frame:

```
y <- 3
myfun <- function() {
   assign('y', 7)
}
myfun()
y
[1] 3
```

paciorek:
Ok, not sure what I was seeing before, but if uses `inherits=TRUE` (as we do in setData) then if 'data' doesn't exist in the local environment, it IS assigned into the parent and that is why things generally work. 

```
y <- 3
myfun <- function() {
   assign('y', 7,inherits=T)
}
myfun()
y
[1] 7
```

Sorry about the noise...

paciorek:
Closing this as PR #946 will accomplish this goal by directly assigning into the field of the class rather than via use of scoping.

PR #957, opened 2019-11-11
wrap file.path and find.package in normalizePath

### LINKED ISSUE #928: Fix path handling on shared drives with MINGW
From a nimble-users message on 22 July 2019, we learned that this R bug can affect nimble users with non-local drives and mingw compiler: https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17582

It looks like a solution is to process all paths through normalizePath().

perrydv:
with winslash = "\\", mustWork = FALSE.

Do not merge.  This is to check first if I broke anything.

It is hoped this will fix #928.

Some check on Windows need to be done.


paciorek:
@perrydv ok to merge this?

PR #958, opened 2019-11-12
recurse into inprod,mat mult, sum when looking for conjugacy

paciorek:
DNM-FTO.

Fixes NCT issue 170.

Before merging need to add testing (see file inprodconj.R).


paciorek:
Tests added. Merge when passes.

PR #961, opened 2019-11-20
simplified printSamplersByType to remove node compression

danielturek:


danielturek:
Changes only to `conf$printSamplers(byType = TRUE)`.

Testing passed, merging this in.

Issue #951, opened 2019-11-01
Warn users of inefficiency if they provide static indexing vectors as data


perrydv:
If a user is unclear on the distinction between data and constants, they might provide an indexing vector as data.  In that case, it will be interpreted as a node and therefore as a stochastic indexing situation, even if it is not.  This will create inefficiency due to larger dependency lists than necessary.  Do we want to identify such situations and warn users?

Reproducible example:
```{r}
library(nimble)

## Providing k as part of data and constants together when given as 
## constants.  This gives correct separation of data and constants, 
## with k treated as constant.
mc <- nimbleCode({
  for(i in 1:5) {
    x[i] ~ dnorm(0, 1)
  }
  for(i in 1:20) {
    y[i] ~ dnorm(x[ k[i] ], sd = 1)
  }
})

m <- nimbleModel(mc,
                 constants = list(x = rnorm(20),
                                  y = rnorm(20),
                                  k = rep(1:5, 4)))

m$getDependencies('x[1]') ## Correct

## Providing k in constants works correctly.
mc <- nimbleCode({
  for(i in 1:5) {
    x[i] ~ dnorm(0, 1)
  }
  for(i in 1:20) {
    y[i] ~ dnorm(x[ k[i] ], sd = 1)
  }
})

m <- nimbleModel(mc,
                 data = list(x = rnorm(20), ## x could go in constants
                             y = rnorm(20)),
                 constants = list(
                   k = rep(1:5, 4)))

m$getDependencies('x[1]') ## correct

## Providing k in data works but will give inefficient MCMC 
## (or whatever method) because it is handled as a stochastic indexing situation.
mc <- nimbleCode({
  for(i in 1:5) {
    x[i] ~ dnorm(0, 1)
  }
  for(i in 1:20) {
    y[i] ~ dnorm(x[ k[i] ], sd = 1)
  }
})

m <- nimbleModel(mc,
                 data = list(x = rnorm(20),
                             y = rnorm(20),
                             k = rep(1:5, 4)))

m$isData('k')
m$getDependencies('x[1]') ## all of y, as if there is stochastic indexing

```

paciorek:
Yes, makes sense to warn. This situation would also arise if user provide 'k' as 'inits'.

[...]

PR #963, opened 2019-11-20
warn users if use RHS-only as dynamic index

paciorek:
This fixes issue #951.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -180,8 +180,11 @@ modelDefClass$methods(setupModel = function(code, constants, dimensions, inits, 
     buildSymbolTable()                    ## 
     genIsDataVarInfo()                    ## only the maxs is ever used, in newModel
     genVarNames()                         ## sets varNames <<- c(names(varInfo), names(logProbVarInfo))
+    warnRHSonlyDynIdx()                   ## warns user if RHS-only nodes used in indexing (inefficient)
     return(NULL)        
 })
+
+
 
 codeProcessIfThenElse <- function(code, constants, envir = parent.frame()) {
     codeLength <- length(code)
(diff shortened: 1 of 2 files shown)


paciorek:
@perrydv @danielturek would either of you like to look at this before I merge?

danielturek:
LGTM.  Thanks for the chance to review.

paciorek:
@perrydv just a ping to see if you want to take a look at this.

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -2715,6 +2718,39 @@ modelDefClass$methods(genVarNames = function() {
...
+            ## Determine vars used in dynamic indexing.
+            vars <- lapply(decl[['dynamicIndexInfo']], function(x) {
+                if(exists('indexCode', x))
+                    return(all.vars(x[['indexCode']]))
+            })
+            vars <- unique(unlist(vars))
+            vars <- vars[vars %in% decl$rhsVars]
+            ## Evaluate indexing to determine nodes used in dynamic indexing.
+            nr <- min(50, nrow(decl$unrolledIndicesMatrix))  # avoid doing full expansion for speed
+            nodes <- lapply(seq_along(vars), function(idx) {
```
perrydv: This approach is extensive.  I guess it makes sense.  Another idea would be to issue the warning simply if an index variable *ever* appears as a LHS node.  I think the difference is that the approach here more thoroughly checks for the first (up to) 50 instances of the index variable, so it tries to handle a case where some indices of the variable are RHSonly and some are not.  Is that the idea?

Would it make sense to wrap all of this in a `try` so that if the checking itself has any glitches it is not an obstacle to the user?  Separately, would it make sense to put this entire checking step behind an option flag, with default being on.  That way, if a user reports a bug in the checking itself, we can simply tell them to set an option FALSE and they will be able to move ahead.  That would seem easier than providing a branch for them etc.

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -2715,6 +2718,39 @@ modelDefClass$methods(genVarNames = function() {
...
+                parentNode <- decl$symbolicParentNodesReplaced[[which(vars[idx] == decl$rhsVars)]]
+
+                return(sapply(seq_len(nr), function(row) {
+                    deparse(eval(substitute(substitute(e, 
+                                                       as.list(decl$unrolledIndicesMatrix[row, ])),
+                                                       list(e = parentNode))))
+                }))
```
perrydv: If we are concerned about computational cost, perhaps we could first grep if the variable name *ever* appears in `maps$nodeNamesRHSonly` and only if yes go ahead with deparsing the first 50 instances.

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -2715,6 +2718,39 @@ modelDefClass$methods(genVarNames = function() {
...
+            })
+            nodes <- unique(unlist(nodes))
+            nodes <- nodes[nodes %in% maps$nodeNamesRHSonly]
+            if(length(nodes))
+                warning("Detected use of non-constant indexes: ", paste(nodes, collapse = ', '), ifelse(nr == 50, ", ...", "."), " For computational efficiency we recommend specifying these in 'constants'.")
```
perrydv: Could we increase the level of explanation by a notch?

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -2715,6 +2718,39 @@ modelDefClass$methods(genVarNames = function() {
     varNames <<- c(names(varInfo), names(logProbVarInfo))
 })
 
+modelDefClass$methods(warnRHSonlyDynIdx = function() {
+    ## Warn if dynamic indexing involves non-constant RHS-only nodes as this causes
+    ## additional dependencies and slower computations.
+    for(i in seq_along(declInfo)) {
+        decl <- declInfo[[i]]
+        if(exists('dynamicIndexInfo', decl) && length(decl[['dynamicIndexInfo']])) {
```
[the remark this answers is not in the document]
paciorek: I'll wrap the whole checking in a condition that checks if `allowDynamicIndexing` is on.

packages/nimble/R/BUGS_modelDef.R
```diff
@@ -2715,6 +2718,39 @@ modelDefClass$methods(genVarNames = function() {
...
+            ## Determine vars used in dynamic indexing.
+            vars <- lapply(decl[['dynamicIndexInfo']], function(x) {
+                if(exists('indexCode', x))
```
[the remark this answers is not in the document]
paciorek: `exists` calls directly out to C, so shouldn't be a concern.

packages/nimble/R/BUGS_modelDef.R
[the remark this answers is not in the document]
perrydv: But see below: I think it would help to *also* wrap it in its own option so it can be turned off if any problems crop up.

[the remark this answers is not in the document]
paciorek: Yep, haven't gotten there yet ...

packages/nimble/R/BUGS_modelDef.R
paciorek: done

packages/nimble/R/BUGS_modelDef.R
paciorek: done

packages/nimble/R/BUGS_modelDef.R
paciorek: Yes, idea was as usual that we don't know how a user might split up a variable for multiple purposes. 

Not quite sure what you mean by `imply if an index variable ever appears as a LHS node`. Do you mean just to check if the index variable coincides with any variables corresponding to RHSonly nodes?

How strongly do you feel about `try`? As I've just modified it, if something goes wrong we can tell user(s) to set the option `checkRHSonlyDynamicIndex` to `FALSE` to prevent the checking from happening.

perrydv: What I meant was if we see `x[ k[i] ]`, so `k` is the index variable, then if `k` *never* appears on the LHS, it must only be RHSonly.  But I guess it is moot to think of other approaches, and in this case also more coarse.

Sure, I guess putting it all behind and option flag gives a sufficient failsafe.

PR #964, opened 2019-11-24
updates and name changes to MCMC_RJ.R

danielturek:
This PR includes the following:

- Name changes of `RJ` samplers: `RJ_prior`, `RJ_indicator`, and `RJ_toggled`.
- Name changes internal, `mcmcConf` to `conf`, and for printing purposes, although it's a slight misnomer, `samplerConf` to `samplerType`.
- Slight generalization of node expansions of indicator nodes, to handle vectors
- Updated `NAMESPACE`
- Updated testing `test-mcmcrj.R` accordingly
- Modified language in help for RJ samplers, and also User Manual, to be more consistent.

Importantly, there are no functional or substantive changes to how the system works.

I welcome thoughts and feedback, in particular from @salleuska and @perrydv


perrydv:
@danielturek The name `RJ_prior` is not intuitive to me.  It sounds like something that samples a prior, but what it means is RJ in the case that the prior probability of inclusion is fixed in the sampler.  How about `RJ_fixed_prior` or `RJ_simple` or something else?

danielturek:
Good comment.  I'm very supportive of `RJ_fixed_prior`.

@salleuska What do you think?

salleuska:
Now I see. Sure, sounds good to me,

danielturek:
@perrydv @salleuska Assuming we have consensus, I'll make this change in the PR

perrydv:
@danielturek I had made changes in the user manual RJ MCMC section on devel, and this created merge conflicts with rjmcmc_dt.  I think the only changes you made were updates to the sampler names.  In that case, I mostly over-rode your changes with the more extensive revisions I made.  If that's not right, please take a look.

salleuska:
Now the system throws an error if one tries to use a multivariate node. 

Not sure about the message I left at lines 297 & 335 `is multivariate; only univariate nodes can be used with reversible jump sampling.` since I think we never mentioned "univariate nodes" anywhere.
Alternatively it may say something like `reversible jump sampling does not support multivariate nodes`. @danielturek @perrydv  let me know if you have better options

paciorek:
I changed the message to "only univariate priors".

@danielturek @perrydv @salleuska I'd like to merge this in. Any concerns?

danielturek:
LGTM

PR #966, opened 2019-12-04
Accumulators in MCMC

danielturek:
Added accumulators to the MCMC.  This includes:

- Implementation
- Documentation
- Testing

Details:

Accumulator variables are specified using either:

```
conf <- configureMCMC(Rmodel, accumulators = c("x", "y"))
## or:
conf$setAccumulators("x", "y")
```

There are also similar methods for resetting and printing accumulators:

```
conf$printAccumulators()
conf$resetAccumulators()
```

After running an MCMC, a matrix containing the accumulated sums and sums-of-squares is retrieved using:

```
as.matrix(Cmcmc$mvAccumulators)
```

The total number of accumulated MCMC iterations is accessed using:

```
Cmcmc$getAccumulatorIterations()
```

Accumulators give the expected behaviour w.r.t. `mcmc$run(niter, reset = FALSE)`, whereby accumulation continues.


danielturek:
@paciorek Assuming testing passes, I'm hoping this will make it into the upcoming release.

danielturek:
@paciorek @perrydv 

Failures with `could not find function is.na.vec` is causing travis failures across the board here.  Is there something I should be doing differently in terms of the testing?

paciorek:
No, my fault. Not sure why I left devel in such a state of disarray. I'm fixing devel now and once it passes testing you can merge that into this branch.

danielturek:
@paciorek @perrydv Testing passed.  Welcome any degree of code review here.

danielturek:
@perrydv @paciorek Any chance of this PR getting reviewed, in advance of the upcoming release?

perrydv:
Are we worried about the numerical issues of estimating variance by naive running sums and sums of squares?  A quick search gives this as a starting point on the issue and known approaches: https://www.johndcook.com/blog/standard_deviation/

danielturek:
Yes, it's a known issue, the [wikipedia article](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance) also addresses this the same.  I figured I'd go with something simple, and be upfront about the calculations.  My vote is "go with as is, rather than over-complicate", but of course I'm open for discussion.

paciorek:
Side note: we recognized this same thing as an issue when we talked about doing a variance calculation on the fly for WAIC, IIRC.  I'm ok with it if we do note the issue where a user will see it.

perrydv:
What do you see as the future extensibility of accumulators?  Would user-defined accumulators fit cleanly in the current design?

danielturek:
Not easily extensible to general calculations, specified on the fly... but easy to extend (via source code) to other functions.  At present, it's hard-coded for sums and sums-of-squares.  But it could be easily extended to include a shifted version, to stabilize the variance calculation.

perrydv:
I think it would be worth some thought on extensibility.  If extensibility is not an issue or need, then the name "accumulators" might be too general, and they could be named for mean and variance estimation.  If it is a need (autocorrelations?), then it seems worth designing with that in mind.

paciorek:
@danielturek I'd like to push ahead with release and there are enough open questions about the interface for the accumulators that I think we should leave it out of this release.

danielturek:
Fair enough, no worries.

paciorek:
This is likely to be superseded by more general work to handle 'derived quantities' in an MCMC. See NCT issue 285.

PR #968, opened 2019-12-06
fix to printSamplersByType for both univariate and multivariate case

danielturek:
Fixes case for posterior predictive sampler, which operates on both univariate and multivariate nodes.


danielturek:
Testing passed on this fix to sampler printing method.  Merging.

PR #971, opened 2019-12-10
updates to conf print for multiple multivariate nodes

danielturek:


danielturek:
Improved printing (again) for MCMC configuration objects.  Testing passed, merging in.

PR #972, opened 2019-12-10
set useNewConfigureMCMC TRUE by default

perrydv:
See #910


perrydv:
@paciorek @danielturek Looks like the useNewConfigureMCMC does still pass testing.

PR #980, opened 2020-02-12
added checks for niter and nburnin in mcmc$run method

danielturek:
Adds checks in the `mcmc$run` method, for:

`niter` < 0
`nburnin` < 0
`niter` < `nburnin`

Any of these result in a helpful error message.


danielturek:
Testing passed, merging in

PR #981, opened 2020-02-13
adaptFactorExponent is now user-settable option in RW samplers

danielturek:
Modified the RW sampler, and also `calculateAdaptationFactor` nimbleFunction (which is used in the block samplers, and also the Wishart sampler), so that the exponent controlling scale-decay is user-modifyable (default is the originally hard-coded value of 0.8)


danielturek:
Tests passed, merging this in.

PR #985, opened 2020-02-22
allow multiplicative link in conjugacy system

paciorek:
FTO - DNM.

@danielturek would be great if you could look over this. 

This adds a multiplicative link capability, in particular for wish, invwish distributions;
note that this causes additional computations because computation of
scale involves matrix manipulations.

We need this functionality for some conjugacies in BNP.

Next steps will be to modify our conjugacy link processing to
identify if offset and scale need to be calculated in cases
where they are possible.

### DIFF
--- a/packages/nimble/R/MCMC_conjugacy.R
+++ b/packages/nimble/R/MCMC_conjugacy.R
@@ -146,7 +146,7 @@ conjugacyRelationshipsInputList <- list(
          ## changing to only use link='identity' case, since the link='linear' case was not correct.
          ## -DT March 2017
          ## link = 'linear',
-         link = 'identity',
+         link = 'multiplicative',
          dependents = list(
              ## parentheses added to the contribution_R calculation:
              ## colVec * (rowVec * matrix)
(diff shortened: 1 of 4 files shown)


paciorek:
Note: test-bnp errors are related to getSamplesDPmeasure and I'll ask Claudia to address those if they persist after merged into bnp_moreGeneral2.  We'll be merging there before merging into devel.

paciorek:
Note to self: this is merged into bnp_moreGeneral2 as it was needed for Claudia and I to move forward. Still want to get review from @danielturek .

danielturek:
All looks reasonable to me.  Although I admit I haven't given much thought about the new `cc_checkScalar` function.  @paciorek do you imagine user-defined functions could show up here?

paciorek:
For functions/operators not in a short list ("[", "*", "+", "-", "/",  "exp", "log", "^", "pow", "sqrt"), it defaults to assuming the output is not scalar since we don't have type processing at this context and a user-defined function could produce any dimensional output.

PR #989, opened 2020-03-14
configureMCMC print default=TRUE, using show method

danielturek:
Opening PR to trigger testing.  Then I'll plan to merge, to see how we like this.


danielturek:
Testing passed, merging

PR #993, opened 2020-03-22
add lkj 'd' and 'r' functions on Cholesky scale

paciorek:
No sampler set up for lkj dist, though. FTO at this point.


paciorek:
This branch now also includes `RW_lkj_corr_cholesky`.  Would like some review from @danielturek . Some comments on the new sampler:

1. It is based on the transformation and associated calculation of the induced prior on the transformed scale given in the [Stan reference manual](https://mc-stan.org/docs/2_22/reference-manual/cholesky-factors-of-correlation-matrices-1.html). Rather than computing the Hastings ratio, I decided to compute logMHR based on a symmetric proposal on the unconstrained scale (the `y`s in the Stan docs) and the induced prior on the unconstrained scale. 
2. As part of the sampler I have a nf method for calculating the transformation to the unconstrained space. This is part of what we need for the `lkj` transformation in the parameter transformation system.
3. The bookkeeping for managing the state of the sampler is a bit involved. I spent a lot of time with it and it may be hard to do a code review given the tedious intricacies, but having another eye on it would be helpful. There is a test that does the update in an inefficient manner that avoids much of the bookkeeping and compares results after 2 iterations of an MCMC. 
4. I decided to do univariate RW updates for each unconstrained element. It's not as bad as it may sound computationally, because updating one unconstrained element then modifies x[i,j] and the other elements in the same column (x[i,(j+1):i]. And the prior on the constrained scale is fast as it only uses the diagonal elements. However it is the case of course that modifying one element means recalculating all stochastic deps of the entire matrix that has the lkj prior. So there is an argument to be made for using a block sampler. Changing to that should be reasonably straightforward.

danielturek:
@paciorek Your approach looks similar to what I'd do.  I'll trust the underlying math is correct, which I did not consider at all. Just a few comments:

- The block sampler might well be worth doing (rather than univariate updates of the unconstrained components), since I think the full dependency calculation for each univariate component would really bog things down.  That's exactly how, using a block sampler on the unconstrained scale, I wrote the `RW_wishart` sampler, and you might refer to that code, if you change over to block sampling.
- Once we reach any sort of "stable state" for the design of the parameter transformations system, we can work this `lkj` transformation code into the `parameterTransform` nimbleFunction, then ideally work that system into both the `HMC` and the `lkj` samplers.

Purely stylistic comments:
- Thank you for keeping the alignment of the assignment lines correct.
- May I request removal of all *entirely blank* lines, within the sampler function definition, to be removed?  If you prefer, they can be replaced with double commented blank lines, as:
```
    [sampler code here]
    ##
    [code continues here]
```

paciorek:
Would be good to document example usage of LKJ in a model given that it is on the Cholesky scale and for a correlation matrix, so there is some work to set up an overall model. 

@danielturek we have a short section 5.2.4.1.1 where we discuss improper priors. I'm inclined to make a short section 5.2.4.1.2 that shows a quick example of LKJ usage. It would look like this in a model: 

```
for(i in 1:p)
   sigma[i] ~ dunif(0, 10)
U[1:p, 1:p] ~ dlkj_corr_cholesky(shape = 1)
Ustar[1:p, 1:p] <- sigma[1:p] * U[1:p, 1:p]
y[1:p] ~ dmnorm_chol(mu[1:p], Ustar[1:p, 1:p], prec_param = 0)
```

paciorek:
Note there is still an open question about the name `dlkj_corr_cholesky`. It's long, but the use of 'cholesky' instead of 'chol' is meant to help distinguish that our usual use of 'chol' relates to how parameters are provided, whereas this use of 'cholesky' relates to the random variable itself.

danielturek:
@paciorek I don't think I'm totally following the example you gave, specifically the line with a vector times a matrix in the model code:
```
Ustar[1:p, 1:p] <- sigma[1:p] * U[1:p, 1:p]
```

So I did a test myself to see how this operates.  Unless I'm missing something, it does seem to give a run-time error in compiled execution, and also non-sensical results, so I believe the example is worth re-thinking.  Try running the code below, and please let me know if I'm misunderstanding something:

```
library(nimble)

code <- nimbleCode({
    prod[1:p, 1:p] <- sigma[1:p] * U[1:p, 1:p]
})

p <- 5
constants <- list(p = p)
data <- list()
inits <- list(sigma = 1:p, U = matrix(1:(p^2), nrow=p, byrow=TRUE))

Rmodel$calculate()
Rmodel$sigma
Rmodel$U
Rmodel$prod

Cmodel$calculate()   ## ERROR
Cmodel$sigma
Cmodel$U
Cmodel$prod          ## INCORRECT
```

paciorek:
Thanks @danielturek . I had meant to come back to the question of vectorized multiplication in this fashion but then forgot to.  So I think we need to fill a matrix column-wise with the standard deviations, and then do direct multiplication so that things are O(n^2) not O(n^3).

danielturek:
@paciorek Sounds good, keep in touch, I'll be curious to see the updated example.

I'm fine with the name `dlkj_corr_cholesky` also, I'm glad we discussed it, and in the scheme of things it's better to have a (longer and) descriptive name for the distribution.

paciorek:
Ok, I am finalizing this, as well as adding the lkj transform for lkj-distributed upper-triangular matrices to the parameter transforms in the ADoak work.

I've added a block RW sampler for the lkj distribution, per Daniel's suggestion. It is now the default.

As far as usage in a model, here is what I have. I think the nf approach will be most efficient, though a bit complicated to explain to users. Doing it in model code would be tricky.

```
uppertri_mult_diag <- nimbleFunction(
    run = function(mat = double(2), vec = double(1)) {
        returnType(double(2))
        p <- length(vec)
        out <- matrix(nrow = p, ncol = p, init = FALSE)
        for(i in 1:p)
            out[ , i] <- mat[ , i] * vec[i]
        return(out)
    })

code <- nimbleCode({
    for(i in 1:n)
        y[i, 1:p] ~ dmnorm(mu[1:p], cholesky = U[1:p, 1:p], prec_param = 0)
    U[1:p,1:p] <- uppertri_mult_diag(Ustar[1:p, 1:p], sds[1:p])
    Ustar[1:p,1:p] ~ dlkj_corr_cholesky(1.3, p)
})
```

PR #998, opened 2020-04-15
added nimble system option MCMCuseConjugacy

danielturek:
Opening PR to trigger testing


danielturek:
Testing passed.  Merging this into `devel`.

Issue #1002, opened 2020-05-02
use `system2` not `system` to fix issue with nimble on R 4.0 on Windows


paciorek:
As of R 4.0, we're having failures because of redirection of output when using `system`.  Solution is to use `system2`. I'm doing this on branch 'fix-win-r40'.

paciorek:
@perrydv do we need to retain the `forceO1` and `stopCompilationBeforeLinking` options? 

Why do I ask?  `system2` takes the command (one word) as first argument and any arguments as a character vector. Those two options involve sending multiline commands to `system` (separated by semicolons), so we'd have to iterate over the multiple lines and pull the "g++" out of them, and then send individually to `system2`.  I'd prefer not to muck around with them if these are options that are not really in use.

perrydv:
Hey @paciorek I'm ok having these disabled. In this case it could be helpful to leave a vestige of the code so that if they are needed in the future we can fix them up.  You could just comment out the sections.

paciorek:
Ok branch "fix-win-r40" should fix this. I've tested on my Windows VM. Up for discussion how/if to get this out to users quickly.

paciorek:
This is merged in to v0.9.1. Still need to merge into devel, so leaving this open.

paciorek:
Now merged into devel by merging in 01c445.

PR #1003, opened 2020-05-02
update to `system2`

paciorek:
WIP: This fixes issue #1002 . 

Important - this branched off 'master' in case we want to do a release that solely fixes this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ notifications:
 branches:
   only:
     - devel
+    - master
 
 latex: false  # Work around tlmgr issues.
 
--- a/UserManual/src/chapter_InstallingNimble.Rmd
+++ b/UserManual/src/chapter_InstallingNimble.Rmd
@@ -66,18 +66,18 @@ On Windows, you should download and install `Rtools.exe`
 available from [https://cran.r-project.org/bin/windows/Rtools/](https://cran.r-project.org/bin/windows/Rtools/).
 Select the appropriate executable corresponding to your version of R
 (and follow the urge to update your version of R if you notice it
-is not the most recent).  This installer leads you through several
-'pages'.  We think you can accept the defaults with one exception:
-check the PATH checkbox (page 5) so that the installer will add the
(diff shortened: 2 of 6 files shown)


paciorek:
Hmm, perhaps because of branching off master, this is not triggering Travis.

danielturek:
@paciorek I think you could add one line to the `.travis.yml` file to fix this, as:

```
branches:
  only:
    - master
    - devel
```

paciorek:
@perrydv @danielturek: I've looked around some at our various commits, PRs and issues and have a rough plan I wanted to run by you.

Idea is to release v 0.9.1 as a bug fix release by using devel (and of course merging in fix-win-r40 branch) with the important exception of reverting commit 43831, which is the commit adding the realized conjugacy link. This seems to be the only commit on devel that is anything other than fairly minor bug fixes or other minor changes. I was also thinking of adding in some of the minor PRs still pending (977, 982, 988, 992, 994, 997). 

Once that was out the door, I would put commit 43831 back in using "git cherry-pick" (just found this functionality), and we would continue on as is, with me starting to do work to prepare for 0.10.0.

We need to figure out if CRAN will complain about the clang11 issue. I have an email in to the maintainer of the prophet package (a Facebook product) who has the same issue but released on April 29. 

Thoughts?

danielturek:
@paciorek All sounds reasonable.  Yes, `git cherry-pick` is useful and flexible, just be cautious of merge conflicts which you'd then have to resolve and commit the manual merge fixes; so, for that workflow, I'd recommend branching off devel, *and then* "cherry-picking" the specific commit you want, and resolving any possible merge conflicts *on that new branch other than devel*, just in case any possible conflicts are subtle or not obvious how to resolve.  Then, assuming all goes well after the cherry-pick, merge that branch back into devel.

paciorek:
Actually, note to self, we can just have devel lead into 0.10.0 because the only thing in v0.9.1 that is not in devel is the fix-win-r40 commit, which we can merge into devel. So can avoid the cherry-pick, I think.

PR #1006, opened 2020-05-06
default MCMC configuration will not monitor top-level data nodes

danielturek:
Opening PR to begin testing.


danielturek:
Testing passed.  Merging this in.

danielturek:
This fixed issue number 202 "data nodes in the middle of a model can result in data nodes being monitored in an MCMC" in `nimbleCoreTeam` repository.

PR #1011, opened 2020-05-25
Update environments set up by nimbleFunction

perrydv:
This tests some re-arrangements of how environments are set up to make it easier to write packages using nimble.


danielturek:
@perrydv Is there a reason that the "name" field for `sampler_BASE` is "nimble_sampler_BASE" ?  Sticking with convention for other NFs, would it be better to have it as `name = "sampler_BASE"` ?

perrydv:
Hi @danielturek, sure, I just pushed that change.

The proposed changes have passed testing.  That's a relief as this is a change that affects every nimbleFunction everywhere.  

We may want to add some testing on defining nimbleFunctions within R functions, if we don't already have such a test.

perrydv:
@paciorek @danielturek This PR now passes tests.  It was originally marked do-not-merge.  Now it is ready.  Any reason not to go ahead?  It is a deep change but one needed for easier packaging using nimble and it would make sense to get it into use by devel users.

danielturek:
@perrydv LGTM

paciorek:
Fine by me.

paciorek:
I will merge this in and then (consistent with @perrydv suggestion to do remove-getLoadingNamespace on top of this) consider merging in remove-getLoadingNamespace.

PR #1013, opened 2020-06-08
fix  UNPROTECT count for nimbleRcall when returnType is void()

perrydv:
This fixes a bug in the generated C++ that results in run-time warnings from R that the PROTECT and UNPROTECT calls are not matched.  It only occurred for returnType void.  It is a single-line fix.


perrydv:
@claudiawehrhahn @paciorek This is the fix needed in nimbleRcall for the BNP need we've been discussing.

PR #1016, opened 2020-06-10
reduce memory needs of compiled getParam

perrydv:
This is ready to merge. It has three changes to the `getParam` implementation.

1. Modify keyword template for `getParam` (and getBound, since it is similar) to rm() an intermediate object that may be large, after it is no longer needed.
2. Modify `makeParamInfo` to operate by declaration, which should typically be faster for large vectors of nodes.
3.  In the case that the `paramID` for all nodes is the same, which will be common (for example when all nodes are from the same distribution), return a sparse representation encoded by `c(-1, paramID)` rather than the equivalent of `rep(paramID, <number of nodes>)`

### DIFF
--- a/packages/nimble/R/nimbleFunction_Rexecution.R
+++ b/packages/nimble/R/nimbleFunction_Rexecution.R
@@ -192,13 +192,25 @@ makeParamInfo <- function(model, nodes, param) {
   }
   if(length(unique(types)) != 1 || length(unique(nDims)) != 1)
     stop('cannot have an indexed vector of nodes used in getParam if they have different types or dimensions for the same parameter.')
-
+    
   ## on C++ side, we always work with double
   if(types[1] %in% c('integer', 'logical')) types[1] <- 'double'
-  paramIDvec <- rep(1L, length(nodeIDs))
-  sourceIDs <- split(seq_along(nodeIDs), nodeDeclIDs)
-  for(i in seq_along(declID2nodeIDs)) { #unsplit() would be another approach to this step.
-    paramIDvec[sourceIDs[[i]] ] <- paramIDs[i]
+
+  if(length(paramIDs) == 1) { ## We could shortcut on this case earlier
+    paramIDvec <- paramIDs
+  } else {
+    if(length(unique(paramIDs)) == 1) {
+      # They are all the same.
+      # We encode this in a sparse way
+      paramIDvec <- c(-1L, paramIDs[1])
+    } else {
+      ## Otherwise, create a full vector of paramIDs
+      paramIDvec <- rep(1L, length(nodeIDs))
+      sourceIDs <- split(seq_along(nodeIDs), nodeDeclIDs)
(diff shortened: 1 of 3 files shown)


perrydv:
@paciorek This appears ready to merge as part of our memory reduction efforts.

paciorek:
I tested the current devel, which includes this PR, against 0.9.1 on Sally's health IRT example. Results are surprising.

Timing for devel:
configureMCMC: 764 seconds
buildMCMC: 1638 seconds
compile MCMC: 196 seconds

Timing for 0.9.1:
configureMCMC: 530
buildMCMC: 1158
compile MCMC: 591

So speed is much better with these changes for compilation, but seems to be rather worse for configuration and building of MCMC. @perrydv does this make any sense?
Note that these times do not include my changes to make BNP setup more efficient, which are not in devel nor 0.9.1. The BNP changes improve configureMCMC and buildMCMC by 10x relative to the devel timing above. 

As far as memory, I only looked at top at the end of compiling the MCMC, but I don't see a big change -- 3.8 GB under 0.9.1 and 3.6 GB under devel.  Again, @perrydv I'm not sure if that makes sense to you or not.

perrydv:
@paciorek I got pretty different results.  I used OS X, with a new R session for each case.  This was the IRT example in unconstrained coordinates.  I focused on configureMCMC and buildMCMC.  (For compiling MCMC, did you use `nimbleOptions(clearNimbleFunctionsAfterCompiling = TRUE)`?)

Timing for devel, with some major components noted:

- configureMCMC : 442
    - checkConjugacy_new: 57
    - addSampler: 12
    - findClusterNodes: 367
- buildMCMC : 1000
    - adding samplerFunctions: 967
        - getDependencies: 22 + 110
        - checkCRPconjugacy: 433
            - findClusterNodes: 360
        - findClusterNodes: 366

Timing for 0.9.1, with the same major components noted:

- configureMCMC : 481
    - checkConjugacy_new: 55
    - addSampler: 12
    - findClusterNodes: 409
- buildMCMC : 1099
    - adding samplerFunctions: 1064
        - getDependencies: 22 + 108
        - checkCRPconjugacy: 478
            - findClusterNodes: 407
        - findClusterNodes: 420

`getDependencies` has been the target of some major efficiency hacking in the past.  Now as we see increasingly big models it is an issue again and we might be able to push more of it into C++.  But it seems we should get your improvements to findClusterNodes (and related) in place first.  I'm not sure what might explain the different results.  Did you run both in the same R session?

paciorek:
Ok, I timed on one of our other Linux machines and this time I get:

devel:

- configureMCMC: 1064
- buildMCMC: 2308
- compile MCMC: 327

0.9.1:
 - configureMCMC: 1072
 - buildMCMC: 2309
 - compile MCMC: 997

So there seems to be a big improvement in compilation of the MCMC. But when I looked at memory use in `top`, it seeemed to be 3.7 GB in both cases. 

Note that my previous numbers were from separate R sessions. I was running things at the same time, I think, so maybe that had some effect.

PR #1019, opened 2020-06-12
CAR complex eigen fix

danielturek:
Updated calculation of eigenvalues of the `Cmatrix` normalized adjacency matrix, as done in the `dcar_proper` distribution.  Specifically, when `Cmatrix` is singular and has repeated 0-valued eigenvalues, these are now calculated (correctly) as 0, rather than `NaN`.

Also includes testing eigenvalue calculations, and moves all testing for `dcar_proper` and `dcar_normal` to a new testing file `test-car.R`.

This PR fixes issue number 229 in the `nimbleCoreTeam` repository.


danielturek:
Noting that testing might fail, on account of mismatch in the "MCMC gold file" of output from test-mcmc.R

danielturek:
Also, @perrydv 

When eigenvalues are initially (incorrectly) calculated as `NaN` by `nimEigen`, we get the following warning messages:

```
## Run-time warning: matrix used in call to nimEigen() has a complex eigenvalue
```

This is generated from line 153 of `inst/include/nimble/EigenTypedefs.h`.  Would there be any way to suppress these warnings?

danielturek:
Testing passed.  Merging into devel.  I'll create a separate "cleanup" issue about the warning messages from `nimEigen`.

PR #1024, opened 2020-06-19
force intermediates of index range expressions to be int

perrydv:
Do not merge.

This is a change that will be useful for AD but is not AD-specific, so I am running tests on it.

The change is that when we have `x[1:b[1] ]`, where `b[1]` could be any expression, we create an intermediate variable for `b[1]`.  Clearly that could be of type `int` but by default behavior that loses the context that it is an index range, it will be of type `double` if `b` is double.  This change would use context information from `:` to force the intermediate to be int.  This would be useful in AD to ensure index ranges are not swept into AD-tracking (which would fail).


paciorek:
@perrydv says we can merge this.

Issue #1015, opened 2020-06-10
incorrect erroring out with 'multiple definitions for nodes'


paciorek:
I just happened to try this weird model in the process of debugging something else.  The error occurs because `R[1:5,1:5]` appears twice in `maps$nodeNames` in `genExpandedNodeAndParentNames3`.

```
code <- nimbleCode({
    R[1:n,1:n] <- sigma2 * exp(-dist[1:n, 1:n] / rho)
    rho ~ dgamma(2, 3)
    sigma2 ~ dgamma(2, 3)
    for(i in 1:n) 
        y[i] ~ dnorm(z[i], var = R[i,i])
})
n <- 5
locs <- runif(n)
dd <- fields::rdist(locs)
model <- nimbleModel(code, constants = list(n = n), inits = list(dist = dd, z = rnorm(5)))
```
Error msg: 
```
defining model...
Error: There are multiple definitions for nodes:R[1:5, 1:5]
If your model has macros or if-then-else blocks
you can inspect the processed model code by doing
nimbleOptions(stop_after_processing_model_code = TRUE)
before calling nimbleModel.
```

paciorek:
There is a bug in `makeVertexNamesFromIndexArray2` in terms of determining contiguity in matrices and arrays.

Suppose we have `vars_2_vertexOrigID['y']` being:
```
1, 2
3, 1
```
then we decide that `y[1,1]`, `y[2,2]` is contiguous block ,even though it obviously doesn't actually make up a block. 

This occurs in any case where the elements that are the subset of interest are symmetric in terms of their row and column indexes, including

```
1 2
2 3
```
and analogs in higher-dimensional arrays.

It looks like we need to do a post-hoc check of the output of `makeSplitInfo` that the indexes for each potential block actually fill out the entire block and not just that they are individually contiguous. I think for each the vertex IDs, if all of the indexes are reported as contiguous we then need to determine if the number of elements is equal to the product of the number of unique index values across the different dimensions. If not, we need to change the contiguity value in `info` for that vertex ID to be 0.  I haven't done this yet but will make a branch to do so, including adding some basic tests of  `makeVertexNamesFromIndexArray2`.

paciorek:
This was fixed but never closed.

PR #1026, opened 2020-06-21
fix detection of contiguous blocks in `makeVertexNamesFromIndexArray2`

paciorek:
Fixes issue #1015. Includes tests of `makeVertexNamesFromIndexArray2`

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -1544,8 +1544,6 @@ makeVertexNamesFromIndexArray2 <- function(indArr, minInd = 1, varName) {
         paste0(all[1,], seps, maxStrs)   ## paste minimum-separator-maximum
     })
 
-
-    }
     dimStrings[['sep']] <- ', '
     newNames <- paste0(varName, '[',  do.call('paste', dimStrings), ']') ## paste together pieces from different dimensions
     list(indices = as.integer(names(splits[[1]])), names = newNames)
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -1507,6 +1507,18 @@ makeVertexNamesFromIndexArray2 <- function(indArr, minInd = 1, varName) {
     ## each entry is (min, max, 0/1 for vector, 0/1 for contiguous)
     info <- lapply(splits, lapply, makeSplitInfo)
 
+    ## Detect and fix cases of non-contiguous indices such as splits being `1`: [1 2] and `2`: [1 2] or `1`: [1, 2] and `2`: [2 1]
+    for(i in seq_along(info[[1]])) {
+        ## Check if a vector and all indexes seen to be contiguous
+        tmp <- sapply(info, `[[`, i) 
+        if(sum(tmp[3,] == 1) > 1 && all(tmp[4,] == 1)) {  # matrix or array seen as contiguous
+            if(length(splits[[1]][[i]]) != prod(tmp[2, ] - tmp[1, ] + 1))  # insufficient elements to fill out the block
+                for(j in seq_along(info))
+                    if(info[[j]][[i]][3] == 1)
+                        info[[j]][[i]][4] <- 0   # set all non-scalar indexes to be non-contiguous
+        }
+    }
+
     ## From here on is the construction of string labels from the info
     dimStrings <- lapply(info, function(x) {
         all <- do.call('rbind', x)   ## This makes a table with a row for each unique element of indArr
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -1503,34 +1503,49 @@ makeVertexNamesFromIndexArray2 <- function(indArr, minInd = 1, varName) {
     ## etc.
     splits <- lapply(arrayWithIndices, split, indArr)
 
-    ## info is a list of summaries from the splits
-    ## each entry is (min, max, 0/1 for vector, 0/1 for contiguous)
-    info <- lapply(splits, lapply, makeSplitInfo)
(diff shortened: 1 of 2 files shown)


paciorek:
@perrydv no urgency but just flagging this in case you want to take a look, since I think you had originally written this functionality.

perrydv:
Hi @paciorek I've got my head into this and see what is going on and what your solution is.  I'm worried about efficiency.  We've been talking about scaling costs in time and memory, and what goes on in `makeVertexNamesFromIndexArray2` is I think one costly model-building steps in both, I think.  You can see it forms a set of objects the size and shape of a model variable to pull out sequences of indices in each dimension of each unique indArr label.  For large variables, this is costly.  If I follow, adding the kind of iteration you did might be very costly for a large array of singleton nodes.  I think it would be better to add more checks in `makeSplitInfo` to set the contiguity flag (4th element) to 0 when it should be.  The purpose of the "first condition" and "second condition" is to try to cheaply catch violations of `splitIndices` having a contiguous-consistent form like `1 1  2 2 3 3` or `1 2 1 2 1 2` or other expansions from higher dimensional arrays (e.g. `1 1 2 2 3 3 1 1 2 2 3 3`).  Unfortunately the lower + upper triangular case gave `2 3 4 1 3 4 1 2 4 1 2 3` and this tricked the cheap checks and was marked as contiguous when it shouldn't have been.  Clearly there could be more laborious checks, but those might involve formation of large objects to construct the various valid forms.  I'm fairly sure I was trying to avoid doing that, and that this whole section of code was already subject to performance squeezing.  What could really make sense would be to push some or all of these steps into C++.  In a way they would make more sense there, in the sense that using R to achieve good performance on this task was not pretty.  Not hard, but would take some time and care.

paciorek:
`makeSplitInfo` operates on individual indexes, so without monkeying with how it is set up, it wasn't clear to me how to modify it (without changing it more drastically) to catch these problematic cases that involve examining multiple indexes at once.

Your case of a large array of singleton nodes would not exercise the code I added because of the `indArr[indArr < minInd] <- NA` check. I need to think through further but I think the case we maybe should be worried about is if the array is defined as a multivariate node and then we use many singleton elements of that node elsewhere in the model.  I can check that case to see whether the code I added meaningfully changes computation time.

perrydv:
I'm not sure what you mean about "individual indices".  I think examples of how `makeSplitInfo` is used are `makeSplitInfo(c(1, 2, 3, 1, 2, 3, 1, 2, 3))`, `makeSplitInfo(c(1,1, 1, 2, 2, 2, 3, 3,3))`, and (in this case), `makeSplitInfo(c(2, 3, 4, 1, 3, 4, 1, 2, 4, 1, 2, 3))`.  Yes, I think you are right about the useful where your new code could be costly: 
```
code <- nimbleCode({
    R[1:n,1:n] <- exp(-dist[1:n, 1:n])
    for(i in 1:n)
        for(j in 1:n)
            y[i, j] ~ dnorm(0, sd = R[i,j])
})
n <- 4
dist <- matrix(1:16, nrow = 4)
model <- nimbleModel(code, constants = list(n = n), inits = list(dist = dd))
```
```

perrydv:
Also I am not sure if it is sufficient to look at info[[1]] only.

paciorek:
Hmm, am I missing something?  In this case of a matrix, `makeSplitIndices` operates separately on the row and column indices (that is what I meant by 'individual indices' - sorry to be confusing in terminology). But in this case we need info on both row and column indices together to decide if we have a full block. 

As far as `info[[1]]`, I think the lengths of `info[[1]]`, `info[[2]]`, ... should always be the same and equal to the number of unique elements in `indArr`, because `info[[1]]` is about the row information for each subset being considered, `info[[2]]` is about the column info for each subset being considered, etc.

perrydv:
OK, right, separately on row and column indices.  I think in theory if there is a non-contiguous block, at least one of those will fail to follow a satisfactory pattern.  It's just that we are cheap in checking that because it would be too costly to check it fully.  This is the first time someone has had a node pattern that breaks it.  

With `info[[1]]` etc., I haven't thought it through fully, but loosely could there be a case where the problem we see here occurs in x[1, 1:4, 1:4]`, so that info[[1]] will just be rep(1, 16), I think?

This is what I mean by pushing into C or C++.  Doing this in R is an exercise in large sets of indexing using operations designing to avoid for loops that would work well using for loops in fast execution.  I know we need a practical short-term solution, but doing it in full detail could be painful.

paciorek:
I'm pretty sure about the `info[[1]]` issue.

Ok so I tested with a 1000x1000 node where the user uses every individual element separately on the RHS elsewhere in a model. With `calculate=FALSE`, the model building takes 120 sec. of which 30 sec. is the original implementation of `makeVertexNamesFromIndexArray2`.
With my addition to the code, there is an additional 40 sec. So non-negligible. 

Furthermore, simply having a RHS-only node that is used in an element-wise way results in use of `makeVertexNamesFromIndexArray2` and incurs the extra cost.

I think I see ways I could get the cost of my new code down quite a bit. If I can do that, my inclination is to leave it at that as a practical solution rather than pushing to C++ given other priorities. We could leave this issue open to remind us that we might come back and push a chunk of `makeVertexNamesFromIndexArray2` down to C++ in the future.

paciorek:
Ok, I think I will be able to get the cost of the extra checking down to be negligible (and to be zero cost when working with vectors, since the current checking is fine for those).

paciorek:
@perrydv Thanks for pushing on efficiency here. I re-organized and the speed now is as good as before because the new check is fully vectorized. For larger cases, it's actually a bit faster now as `info` is now a list of matrices rather than a list of lists. Turns out it is faster to create `info` in this way, and it also eliminates the step of creating `all` later.

@paciorek remember to check logic of the new code on a couple test cases once this passes testing.

PR #1029, opened 2020-06-22
remove `where = getLoadingNamespace()`

paciorek:
We've determined this isn't needed (though it doesn't hurt). Remove so doesn't confuse developers of nimble-depending packages.


perrydv:
@paciorek I suggest we try this on top of #1011

perrydv:
@paciorek I'm not seeing where but it looks like this test failure is just due to some typo so travis isn't parsing the source code correctly.

paciorek:
This was merged into #1011 and merged in via that route.

PR #1044, opened 2020-07-14
Support NA in compiler

perrydv:
Several user issues recently have requested support for NA in compiled code.

It is not always clear that one should "want" to use NA in a model context, but supporting it for nimbleFunctions is a more general need.

Doing so is not entirely straightforward.  We can support NA for integer and double types, but not for logicals, in compiled code.  In compiled code, logicals are implemented at C++ bools, and there is no way to extent beyond `true` and `false` values for those.  That does not mesh naturally with R, in which "NA" *defaults* to logical.  In this PR, NA defaults to integer in compiled code.

Some ancillary aspects of compiler support for flexible numeric operations were noticed in exploring the NA issue but will be dealt with separately.  In particular, it would be nice to issue a warning if NA is assigned to a logical.  This looks like it fits into a larger need to issue warnings for "information loss" assignments, like "int <- double".  We currently issue such warnings for these if the LHS is a name but not an indexed expression.  It would also be nice to special-case the warning if there is NA on the RHS in order to provide more helpful information.

There is still a need to make default = NA work.  That is not part of this PR.

### DIFF
--- a/UserManual/src/chapter_RCfunctions.Rmd
+++ b/UserManual/src/chapter_RCfunctions.Rmd
@@ -393,6 +393,16 @@ for(i in 1:10)
    x[i] <- foo(y[i]) 
 ```
 
+#### How NIMBLE handles `NA`
+
+NIMBLE supports use of `NA` with some caveats.  In the compiled version of a nimbleFunction:
+
+  - `NA` values in a logical scalar or vector will not work;
+  - Assigning `NA` to a new variable will make that variable have type double (in R, this would be a logical!);
+  - Non-trivial use of `NA` in integer variables may fail by having the value become a large (negative) number;
+  - Use of `NA` in doubles should generally work.
+
+These issues arise because NIMBLE uses R's encoding of NA values in C(++) but uses native compiled math and type-casting in C++, which do not always preserve R's NA encodings.  In summary, NA should generally work for doubles and should be used cautiously (i.e. after you test that what you need to work actually works) for integers.  For some needs, NaN is a suitable alternative to NA.
 
 
 #### Changing the sizes of existing objects: *setSize* {#sec:chang-sizes-exist}
--- a/packages/nimble/R/genCpp_generateCpp.R
+++ b/packages/nimble/R/genCpp_generateCpp.R
@@ -96,10 +96,17 @@ nimCppKeywordsThatFillSemicolon <- c(
 
 ## Main function for generating C++ output
 nimGenerateCpp <- function(code, symTab = NULL, indent = '', showBracket = TRUE, asArg = FALSE) {
-    if(is.numeric(code)) return(if(is.nan(code)) "(nimble_NaN())" else code)
-    if(is.character(code)) return(paste0('\"', gsub("\\n","\\\\n", code), '\"'))
+    ## The literal token NA should not appear in code as numeric or character, since it is logical in R.
+    ## Hence the checking of is.na cases for is.numeric and is.character are defensive only and not expected to be necessary.
+    if(is.numeric(code)) return(if(is.nan(code)) "(nimble_NaN())" else if(is.na(code)) 'NA_REAL' else code)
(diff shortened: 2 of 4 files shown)


paciorek:
@perrydv I was looking at our current test for use of NA in compiled code and adding a new test, and it's not clear to me that things are fully as we want:

```
## doesn't seem to be what we want
nf <- nimbleFunction(
    run = function() {
        ans <- NA
        return(ans)
        returnType(double())
    })
cnf <- compileNimble(nf)
cnf()
## [1] -2147483648

## this seems fine
nf <- nimbleFunction(
    run = function() {
        return(NA)
        returnType(double())
    })
cnf <- compileNimble(nf)
cnf()
## [1] NA
```

Note that this seems fine to assign NA into an existing vector:

```
 nf <- nimbleFunction(
	run = function(y = double(1)) {
            y[1] <- NA
            returnType(double(1))
            return(y)
        })
expect_true(is.na(nf(rnorm(3))[1]), 'NA in vec uncompiled')
cnf <- compileNimble(nf)
expect_true(is.na(cnf(rnorm(3))[1]), 'NA in vec uncompiled')

**Comment by perrydv** [comment]
@paciorek I think the issue is casting from integer to double.  This works:
```
nf <- nimbleFunction(
    run = function() {
        ans <- NA ## Currently this makes ans be integer
        return(ans)
        returnType(integer())
    })
cnf <- compileNimble(nf)
cnf()
```

And this works
```
nf <- nimbleFunction(
    run = function() {
        ans <- 2.34 ## For ans to be double
        ans <- NA
        return(ans)
        returnType(double())
    })
cnf <- compileNimble(nf)
cnf()
```

The problem occurs when an int NA is cast to double by assignment or return.

There is no problem when a double NA is cast to int.

We could make assignment to NA default to double for a new variable.  But for a variable that is already int, we would still have this problem.  It would take a deeper dive into NA internals to see if there is an easy way through this.  At the moment it seems like a tradeoff between supporting NA in most cases but having potentially catastrophic failure if it is cast to a non-NA value, vs continuing to not support NA.

perrydv:
This discussion from "Rcpp for everyone" is helpful:

[https://teuder.github.io/rcpp4everyone_en/240_na_nan_inf.html#points-to-note-when-handling-na-with-rcpp](https://teuder.github.io/rcpp4everyone_en/240_na_nan_inf.html#points-to-note-when-handling-na-with-rcpp)

paciorek:
I guess NAs will mostly arise in vectors, so it seems like things are mostly fine. I suppose it would be best to do this, as you say: "We could make assignment to NA default to double for a new variable." 

But not sure if we should do that for the release or just merge this in.

perrydv:
It looks like it's not really vector vs. scalar.  It's integer vs. double.  The problem is when NA is assigned to an integer and then cast to a double (e.g. as part of returning a double).

paciorek:
Ok, fair enough. My point was that often users would be using vectors of doubles and assigning an NA into an element. And it seems like that use case is fine. But I suppose that a user might create vector of NAs de novo in a nimbleFunction, which would by default would be integers, and then would often try to assign doubles into that vector, which would create problems. 

Does that sound like a correct assessment?

paciorek:
@perrydv shall I merge this and open a NCT issue that we may want to make assignment to NA default to double?

perrydv:
I went ahead and made the change and drafted a section for the manual.  Please review.  Let's see if any tests break.

perrydv:
I think the test failure is a case where the test expected failure for something that now works.  I will check.

perrydv:
Ok @paciorek The new tests related to NA pass.  If you want to see what works and doesn't, the tests do include some known failures.

paciorek:
Looks like we can merge in, right @perrydv ?

perrydv:
Yes.  I anticipate someday we'll get user questions around this, but still it's an improvement.

PR #1046, opened 2020-07-15
fix access to member data of >2D in a nested nimbleFunction

perrydv:
This PR fixes cases like the following, which are not working:

```
nf1 <- nimbleFunction(
  setup = function() {
    A <- array(0, dim = c(2, 3, 4))
    setupOutputs(A)
  },
  run = function() {}
)

nf2 <- nimbleFunction(
  setup = function() {
    mynf1 <- nf1()
  },
  run = function() {
    ans <- mynf1$A[,,1]
    returnType(double(2))
    return(ans)
  }
)
mynf2 <- nf2()
cmynf2 <- compileNimble(mynf2)  
```

The problem only occurs if `A` has dimension > 2.

### DIFF
--- a/packages/nimble/R/genCpp_eigenization.R
+++ b/packages/nimble/R/genCpp_eigenization.R
@@ -835,7 +835,7 @@ eigenizeName <- function(code, symTab, typeEnv, workEnv) {
     
     if(needStrides) {
         newArgs <- list()
-        newArgs[[1]] <- code$name
+        newArgs[[1]] <- as.name(code$name)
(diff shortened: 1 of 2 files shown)


paciorek:
@perrydv Apparently this didn't get merged into 0.10.1. Do we want it for 0.11.0? If so, and if you think it is ready, I can check on the test failures, which I am guessing are unrelated to this PR.

perrydv:
@paciorek Flagging that I finally circled back and fixed this.

PR #1050, opened 2020-07-17
full revision of `getSamplesDPmeasure`

paciorek:
Also includes update to chapter_bnp.Rmd with info on 'more general' functionality.
And the pdf of the manual is updated to reflect that.

The changes in this branch were really done in `bnp_outputG_list` but merging that into devel was presenting ugly merge conflicts so I copied the changes into this branch.


paciorek:
@claudiawehrhahn I'm seeing three test failures in test-bnp.

1) The first set of code after:
```
test_that("Test that weights of getSamplesDPmeasure function are correctly computed", {
```
is failing because `aux$trunc` does not exist.

2 and 3) Two failures in the code after:

```
test_that("check use of epsilon parameters in getSamplesDPmeasure", {
```

Same problem. It looks like you no longer return `$trunc` as output of `getSamplesDPmeasure`.

If you're able to fix this, please do in `bnp_outputG_list2`.

PR #1060, opened 2020-08-08
add as.list for a modelValues

perrydv:
Back in January we drafted this `as.list` method for a `modelValuesBaseClass` and `CmodelValues`, i.e. uncompiled and compiled modelValues objects, for a specific purpose.  These allow `as.list(mcmc$mvSamples)` and `as.list(Cmcmc$mvSamples)` to work.  `as.list` returns a list with variable-named elements, each of which has the size and shape of the variable plus one dimension to index MCMC iteration.  MCMC iteration is by default the first index but ` iterationAsFirstIndex = FALSE` makes it the last index. We regularly hear of use cases where this would be helpful, so this PR opens discussion on adding it to the package.  There are arguments to control which variables are requested (default: all that are monitored) and whether to arrange the MCMC iteration as the first index (default: TRUE).

This PR would add `as.list.modelValuesBaseClass` and `as.list.CmodelValues` to the package but would not integrate them with functions like `nimbleMCMC` or `runMCMC`.  I'm thinking we could discuss that for a later release.

This PR includes a test and minor changes to documentation to draw attention to the `as.list` feature.
.

### DIFF
--- a/UserManual/src/chapter_DataStructures.Rmd
+++ b/UserManual/src/chapter_DataStructures.Rmd
@@ -148,17 +148,30 @@ customMV$a
 
 
 Often it is useful to convert a modelValues object to a matrix for use
-in R.  For example, we may want to convert MCMC output into a matrix
-for use with the `coda` package for processing MCMC samples. This
-can be done with the `as.matrix` method for modelValues
-objects. This will generate column names from every scalar element of
-variables (e.g. "b[1, 1]" ,"b[2, 1]", etc.). The rows of
-the modelValues will be the rows of the matrix, with any matrices or
-arrays converted to a vector based on column-major ordering.
+in R.  For example, we may want to convert MCMC output into a matrix,
+for use with the `coda` package for processing MCMC samples, or into a
+list. This can be done with the `as.matrix` and `as.list` methods for
+modelValues objects, respectively. `as.matrix` will generate column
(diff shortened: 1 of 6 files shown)


danielturek:
@perrydv Overall I like the idea of this a lot.  I will happily discuss this further, and integrate usage into `runMCMC` at a later time.

One request though: I find it unintuitive when arguments assume the default value of `TRUE`, and if we want MCMC iteration number to be the *last* iteration, we have to do some backwards logic and distill the need to specify the argument `iterationAsFirstIndex = FALSE`.  I would **much** prefer the intuitive approach of removing that argument (in name), and instead having the argument `iterationAsLastIndex`, with default value `FALSE`, and when we want the MCMC iteration as the *last* index, we specify `iterationAsLastIndex = TRUE`.  Can we please strongly consider that?

perrydv:
Hi @danielturek Yes, good point.  I agree.  I remember the coding thought process that led to the draft as it is, but I agree with your take on what will feel more natural.  Unfortunately I'm heading out, so would you mind making the change (including with checks in roxygen and user manual spots)?  Alternatively I can make it with a week or two delay.

danielturek:
@perrydv Done.  The modelValues `as.list` argument name has been changed from `iterationAsFirstIndex` to `iterationAsLastIndex`, also changing it's default value.  Documentation updated as well.

PR #1061, opened 2020-08-09
added new setMontiors and setMonitors2 methods to MCMC configuration objects

danielturek:
The title says it all


danielturek:
Testing passed.  Merging this in.

PR #1063, opened 2020-08-15
partially fix scoping issues with user-defined distributions

paciorek:
WIP:

1) makes a copy of the auto-generated "r" function for a user-def dist in the GlobalEnv to avoid scoping issues.
2) updates `prepareDistributionInput` so that user-def dists can be defined inside functions.

Note, this PR does not address the larger issue (e.g. NCT 236) that user-defined dists, functions, samplers need to be in the GlobalEnv and will not work if solely in functions, which makes things complicated for parallelization, where we tell users to put everything inside the function.


paciorek:
Hmm, this is complicated by the fact that it violates CRAN policy for a package to modify GlobalEnv. I'm removing the assignment.

PR #1065, opened 2020-09-06
MCMC control list extraction

danielturek:
Modified MCMC samplers to use new control list extraction function `extractControlElement` function, to prevent errors caused by R's partial matching of list element names.


paciorek:
@danielturek Do you really want `extractControlElement` exported. If so we'd need to add Roxygen info for the Rd. I'm removing the `@export` unless you tell me otherwise.

danielturek:
@paciorek Please see email discussion that I just forwarded to you (from the users' list, with David Pleydell, a contributing user).  I would prefer to keep `extractControlElement` exported, unless there's a better way to support full usage, or a better way to manage the MCMC list extraction.  I'm happy to discuss, but for the meantime, perhaps the "export" shouldn't be removed just yet.

paciorek:
Fine by me. Are you up for adding `@export` back in and with Roxygen documentation?

perrydv:
@danielturek @paciorek IIUC, David was making a new nimbleFunction by copying and modifying one in our source code.  I'm not sure that's a good reason for every function used in setup code to be exported.  Instead, someone copying and modifying could insert "nimble:::" where needed to gain access to non-exported functions.

danielturek:
@perrydv @paciorek Well, I'm no longer (and admittedly, never was) sure what the right solution is here.  As is, the `extractControlElement` provides a single entry point for MCMC sampler setup functions, for handling errors like the one which came up.  But, unless this function is exported, users can't copy and re-use, or make extremely minor tweaks to samplers - which seems like a big easy bonus to allow, without them jumping through other hoops. And, I don't think there are any other functions used in setup code which aren't exported, at least not commonly (or ever) used in setup of MCMC samplers.  So it felt good to me to have this function exported.  Open for discussion on better options.

perrydv:
Thanks for that explanation.  I agree that sounds general enough that it would be good to export it.

danielturek:
@paciorek Are you ok to re-add `export(extractControlElement)` to the `NAMESPACE` file?

paciorek:
`NAMESPACE` is populated automatically based on having `@export` in the R code file.
But if it's exported we need documentation. That could be real documentation or just defaulting to our "artificial" documentation for stuff we don't really want to document, in order to satisfy CRAN.

@danielturek  So just let me know if this is something where we should have a least a bit of real documentation. If so, it would help if you would write the roxygen.

danielturek:
@paciorek just noting I believe we have quite a lot of exported functions, with no documentation.  But I'm happy to add something for `extractControlElement`, I'll do that now. But, for example, copied directly from `BUGS_utils.R`:

```
#' @export
is.model <- function(obj, inputIsName = FALSE) {
    if(inputIsName) obj <- get(obj)
    return(inherits(obj, 'modelBaseClass'))
}

#' @export
is.Rmodel <- function(obj, inputIsName = FALSE) {
    if(inputIsName) obj <- get(obj)
    return(inherits(obj, 'RmodelBaseClass'))
}

#' @export
is.Cmodel <- function(obj, inputIsName = FALSE) {
    if(inputIsName) obj <- get(obj)
    return(inherits(obj, 'CmodelBaseClass'))
}
```

danielturek:
@paciorek Roxygen documentation for `extractControlElement` is now added, and appears on devel branch.

paciorek:
Yeah, `is.model` is an example of where we just point to the "nimble-internal" artificial documentation that I was referring to. It's set up in `prep_pkg.R`.

PR #1068, opened 2020-09-16
fixed bug in runCrossValidate for merging of MCMC configurations

danielturek:
Fixes a bug in `runCrossValidate`, to correctly merge the sampler configuration lists from two different MCMC configurations.

Used `conf$addSampler` as was proposed by @perrydv, to do this in a robust manner.

Also added a test (using the dyes example), where cross validation now gives results that seem reasonable.  The test checks for the exact numerical cross validation results.


danielturek:
@paciorek I suspect this will initially fail testing, due to the travis bug.  But once that's fixed, this PR should pass (as it works for the cross-validation tests), and I'm hoping you can merge it into the upcoming release.

perrydv:
@paciorek Does this test failure look like just the travis issue we've been seeing?

paciorek:
Yeah, that is it.  It didn't finish the other tests after numericTypes, but presumably those won't be an issue.

I should be fixing the precision error this morning.

paciorek:
Yes, that's what I see.

danielturek:
This is being addressed in PR  #1298.

Pending testing results.

PR #1070, opened 2020-09-21
Remove SMC tools from NIMBLE

dochvam:


paciorek:
@dochvam I see that `test_mcmc`, `expect_compiles` and other utilities are in `test-noSMC.R`. `test_mcmc` and `expect_compiles` and probably the others already live in `test_utils.R` so just wanted to check your thoughts on why such utilities are in `test-noSMC.R`.

dochvam:
@paciorek I have no idea why the redundancy--maybe a bad habit from working on migrating to nimbleSMC. Looks like you fixed, so thanks!

paciorek:
@dochvam @perrydv @danielturek in `noSMC` we don't have any indication for the user that `buildBootstrapFilter` and the various other PF methods exist in `nimbleSMC` (right, Ben?).

Ben did put a message in for when users try to add a PF-based MCMC sampler.

Do we want dummy `buildBootstrapFilter` etc. functions that indicate these have been moved?

On somewhat similar lines, how do we want to handle the manual. It would probably be a bit of a pain to have a separate nimbleSMC manual to point to. I'm somewhat inclined to just note at the beginning of the SMC section (and in an appropriate place in the MCMC chapter) what has been moved to `nimbleSMC`.

paciorek:
Another comment here is that our title for the nimble package is:

```
MCMC, Particle Filtering, and Programmable Hierarchical Modeling
```

which is now only accurate if we are talking about the NIMBLE system. I think this is ok.

danielturek:
@paciorek @perrydv I almost feel like this title should be fully revised; but, as always, I'm quite flexible.

perrydv:
I am ok leaving the title as is, at least for now.  I like the idea of adding informative messages for the `buildBoostrapFilter` etc.

paciorek:
Also one implication of having stubs for `buildAuxiliaryFilter` etc. is this message when one loads `nimbleSMC`:

```
Attaching package: ‘nimbleSMC’

The following objects are masked from ‘package:nimble’:

    buildAuxiliaryFilter, buildBootstrapFilter, buildEnsembleKF,
    buildIteratedFilter2, buildLiuWestFilter
```

PR #1071, opened 2020-09-24
Double not integer

DRJP:
Suggested addition to the user manual highlighting that BUGS code cannot pass integer vectors to run functions.


perrydv:
This fixes #1072

paciorek:
I find the suggested language hard to understand. In particular I don't know what "pass via BUGS code to run code" means. Users don't declare types in BUGS code, so I think that this is referring to the output type of user-defined distributions (and user-defined functions). Is that right? If so, we should reword to make that point explicitly.

perrydv:
Hi @paciorek I'd say go ahead and revise and many thanks to @DRJP for the PR the get the ball rolling in a concrete way.  I think the main point for users is that all variables in a model will be doubles.  That means that if one creates a user-defined function or distribution, both arguments and return types should be doubles.  An exception is for scalars, which should cast automatically across types.  The first draft of language in the PR does seem somewhat narrower, coming from the use case that raised the confusion, so let's make it more general.

DRJP:
Thanks guys for taking a look at this. For sure, reword as you see fit. 

"pass via BUGS code to run code" - by this I meant that you are using a user defined function inside a BUGS model. In the examples I've looked at this has been at deterministic nodes only (e.g. [script posted on nimble-users](https://groups.google.com/g/nimble-users/c/7UNfrFE916c)) so I am only _assuming_ that other node types would behave similarly. As far as I can tell, this would (almost?) always be a function that has no setup code, so we're just talking about run code only functions. The compilation error occurs when input vectors to the run function are declared integer(1) in the function - so the goal is to warn users against doing that. Curiously, setting the returnType to integer(1) (or integer(3) in my case) does not cause compilation errors - which I imagine means that a cast to double is successfully applied somewhere. 

It could also be worth adding the word "cast" to the document, in case anyone with C or C++ experience is trying to find how casting is done in nimble. The details are there in "11.3.2.1.1 Avoid changing types of a variable within a nimbleFunction", but someone searching the document for "cast" will not find that section.

paciorek:
Ok, in devel, I added a note in the User-defined chapter that vectors must be doubles (the second sentence here is the new one):

```
  - NIMBLE uses doubles for numerical calculations, so we suggest simply using doubles in general, even for integer-valued parameters or values of random variables. In fact, non-scalars must be declared as doubles.
```

I think it makes sense to put this in the user-defined chapter and not in the chapter on writing nimbleFunctions, since it really has to do with using nimbleFunctions in model code.

@perrydv I'll leave it up to you to respond to the suggestions about casting.

danielturek:
@paciorek How about a (brief) mention of this in both sections?  I know we don't want the user manual to be "bloated", but it seems relevant to both.  But no worries if we decide otherwise.

paciorek:
@danielturek  Not opposed, but it wasn't clear to me where in the nimbleFunction chapter it belongs, given it only applies to user-defined dists/funs. Do you have a suggestion?

danielturek:
@paciorek Perhaps 11.4 "Declaring argument and return types" ?  But if it's sufficiently tangential (being about functions for use in models) that's ok, it does make sense.

paciorek:
Ok, I've added a note in 11.4. Closing this without merging as I just inserted relevant text directly into devel.

PR #1075, opened 2020-10-12
Safe run mcmc

### LINKED ISSUE #1074: Add try to runMCMC
Evidently if a step such as `samplesSummary` fails, which can happen if there are any NAs arising from bad initialization or mixing, then `runMCMC` fails.  Can we wrap steps inside `runMCMC` in `try` to provide some failsafe behavior and not lose all the samples?

danielturek:
Fixes #1074 .


danielturek:
Testing passed, merging into devel.

PR #1078, opened 2020-11-03
removed samplerAssignmentRules system

danielturek:
Deprecates MCMC `samplerAssignmentRules` system.


danielturek:
@paciorek @perrydv Now that testing has passed, I would welcome a second set of eyes on this PR, and agreement that this is what we're looking for.

perrydv:
Only two minor comments @danielturek.  In the comment around line 213 of configure_MCMC (but line number may have changed), there is a comment before a section of code describing it as "use old (static) system ...".  This might confuse anyone trying to understand the code.  Could we change that comment to say this is the current (now only) assignment system?  Also I would vote for just removing some of the larger chunks of code instead of commenting them out.  We can keep them handy in our own archives in case we want to rework this at some point (and of course they're in the repo history).  Some of the small commented-out steps (like single function arguments) may be more helpful to remind us of changes.  Otherwise LGTM.

danielturek:
Removed large commented code chunks, as suggested.  Will see if testing passes, then merge into devel.

PR #1079, opened 2020-11-06
Auto block warning

danielturek:
Using the `autoBlock` MCMC option gives the following output:

`The autoBlock option is no longer actively maintained.  In some cases, it may not operate correctly with more recent system features and/or distributions.`

In addition, this same information was added to the User Manual, and the Roxygen documentation for `autoBlock`.

### DIFF
--- a/UserManual/src/chapter_MCMC.Rmd
+++ b/UserManual/src/chapter_MCMC.Rmd
@@ -204,6 +204,8 @@ The default MCMC configuration includes monitors on all top-level stochastic nod
 
 #### Automated parameter blocking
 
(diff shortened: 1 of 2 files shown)


danielturek:
@paciorek @perrydv Testing passed, and I'd welcome a second set of eyes on this.

paciorek:
I do like the idea of putting this into the 0.10.1 bug fix release given the nature of the warning.

danielturek:
Fine with  me.  @perrydv thoughts?

@paciorek if we reach consensus, then I'm fine with you actually doing the merge of this PR, any time.

perrydv:
LGTM

danielturek:
Merging this into devel.

PR #1080, opened 2020-11-18
fix makeParamInfo when there is only one declID involved

perrydv:
Do not merge.

This is for testing of a small change to makeParamInfo.


perrydv:
@paciorek I think this is ready to go.  I think I got the MCMC testing update right, but please check.

danielturek:
LGTM.  Nice additions to the testing.

PR #1087, opened 2021-01-09
dynamically adjust conjugacy to actual presence of coeff and offset in setting up conjugate samplers

paciorek:
This only sets up storage for, calculates, and uses coeff and offset in conjugate sampler run code if coeff and offset are actually present in the model structure. 

This addresses NCT issue 187.


paciorek:
I did a quick benchmark with this basic model that has all conjugacy and no offset/coeff needed and there was some speedup but not huge: roughly 2.6 seconds for devel and 2.3 seconds on this branch. Would probably make more of a difference with multivariate conjugacies where not using this approach involves calculating a matrix of coefficients.

perrydv:
Sounds good to me @paciorek.  I skimmed the code but didn't get my head into the details.  The benchmark you ran may not sound like much but that's order 10% and so definitely worth it.

PR #1092, opened 2021-01-28
Disallow any scientific notation when deparse, as.character or paste convert numbers to text

### LINKED ISSUE #1091: incorrect node names for large models in maps
It looks like nodes that have round indexes, e.g., `mu[100000]`, `mu[200000]` etc. are not being correctly labeled in the maps.

```
code <- nimbleCode({
for(i in 1:100003) 
y[i] ~ dnorm(0,1)
})
m <- nimbleModel(code)
m$modelDef$maps$nodeNames[100000]
## [1] "y[1e+05]"
> m$modelDef$maps$nodeNames[100001]
## [1] "y[100001]"

nn=m$getNodeNames()
nn[100000]
## [1] "y[1e+05]"
m$simulate(nn[100000])
m$y[100000] ## so it's ok so long as we refer to the node by its true name in the maps when calling 'simulate'
## [1] 0.2830629
## actually ok regardless - we must be transforming the character string version of the name correctly
m$simulate('y[100000]')
m$y[100000]
## [1] -0.9506755
```

This presumably *could* have many consequences, but only if some code is generating "y[100000]" when the node name is "y[1e+05]". I haven't determined where there might be inconsistencies, but presumably in the user's case there are.  

@danielturek can you check the example above on your system. Perhaps there is an R version effect or something else going on that explains the different behavior on your system vs. mine.

perrydv:
Fixes #1091 

In the issue we discussed potential problems from deparse and as.character.

Turns out we already funnel deparse calls into our own version to expand the default character limit, so it was easy to add `control = "digits17"` to reduce conversion from numbers to scientific notation as text.

Changes from `as.character` to `format(x, scientific = FALSE)` were done where needed one-by-one.

Turns out the immediate culprit was in `paste`.  This was addressed by changing the `scipen` options in R and reverting it via `on.exit`.


danielturek:
@perrydv LGTM

I'm pleased with how (relatively) simple the changes ended up being.  I (myself) think it's ok with temporarily changing the `scipen` option, but will wait to hear what Chris thinks.  I trust you hunted down all instances of `paste` and `as.character`.

About weeding out some older (now unused) functions, that code is OG indeed, pre-dating the first commit to `nimble-dev/nimble`.

paciorek:
It's not clear to me to what extent the changes primarily rely on scipen versus on modifying arguments to individual function calls, but that's ok. I've said my piece about changing options and trust Perry has come up with good approach.

Is there a reason not to simply remove `parseEvalCharacter{,Many}`? In some cases I do like to preserve old code in comments, but preserving unused functions doesn't seem useful. I might be missing something.

Do we want at least a basic test that a model with `x[100000]`, `x[1000000]`, etc. ends up with the correct node names?

Also, I'm wondering if there are places we could catch if a user (or perhaps even our own code, unintentionally) creates and tries to use a node name like `x[1e+05]`, such as 

```
i = 100000
mynode <- paste0('x[', i, ']')
model$getDependencies(mynode)
```

perrydv:
Thanks to both of you for looking at this.  The reliance on scipen is minimal, and when it is used it is always localized, guaranteeing that it will be reset to its previous value upon exit.  This is done by `on.exit`, which should cover exiting by any means, including erroring out.  Well, such a change is done for the entire `nimbleModel` process, so I guess that is not really minimal, but it should still be considered localized.

@paciorek I agree we should remove those commented-out code blocks.  As I was searching relevant code, I saw those (uncommented), realized they are dreck that is never used, and commented them to get through testing and ensure I wasn't making a mistake.

And I agree we should put in new tests.   I will try that unless you want to.

perrydv:
Hi @paciorek I was writing tests for this and here are some observations about a case we discussed: What if a user provides a malformed node name like `"x[1,1]"` or `"x[1e+5]"`.  The recommended practice is to put it through `model$expandNodeNames` (or other model processing) to get the canonical name format.  If that isn't done, it looks like in compiled code:

```
model[['x[1e+5]']] # works
values(model, 'x[1e+5]') # does not work
model$calculate('x[1e+5]')) # works
nimCopy(from = model, to = mv, rowTo = 1, nodes = 'x[1e+5]', logProb = TRUE) # does not work
```

These just depend on whether the partial evaluation / compilation setup happens to process the node names in a way that corrects them.  I think this is really a distinct issue because it would apply as well to `"x[1,1]"` (should be `"x[1, 1]"`), but is it something you think we should address?  I think the expectation in the past has been that a nimbleFunction programmer should ensure that node names are valid before using them in run code.  (I've entered the literals, but the same thing happens if the name is in a variable.)

perrydv:
@paciorek I realized I hadn't tried the case you suggested:
```
m$getDependencies('x[1e+05]')
```
This does work.

paciorek:
@perrydv I don't feel strongly either way about whether we should more consistently handle a user giving us a malformed node name. Ideally of course we would handle it consistently either with it working or with giving them an informative error. But not our top priority. We should probably make NCT issue of it.

perrydv:
Sounds good.  I agree.  There would be some overhead to adding support like that.  For the moment, I will add tests for what is on this branch and then we can merge.

perrydv:
i added tests and put them in `test-expandNodeNames`

paciorek:
@perrydv @danielturek ok to merge?

danielturek:
@paciorek Ok with me.

perrydv:
👍

PR #1093, opened 2021-02-03
Change Eigen (not latest) with a few warnings suppressed

perrydv:
On ADoak we have an updated version of Eigen (not now the latest) with a few tweaks to suppress compiler warnings generated from R CMD SHLIB's compiler invocations.  This PR moves that version of Eigen over to devel.  It removes compiler warnings on my system.  We could update further or other ideas, but this is a practical step for now.


perrydv:
@paciorek @danielturek What do you think of merging this?  It will make other work easier.

paciorek:
Fine by me. 

I am still getting a slew of warnings on Linux.

perrydv:
@paciorek At some point let's see if we can quiet warnings down on linux too.  I'm going to merge this because because it makes work on the C++ code much easier on OS X.  Also setting up ccache might help quiet warnings (and make builds faster).

PR #1095, opened 2021-02-04
add model restoration to nimDerivs of a nimbleFunction including a model

paciorek:
This addresses NCT issue 268 by restoring model state (including logProbs) when `order=0` is not included in the order of `nimDerivs` of a nimbleFunction method and when double-taping.

The approach is to check of the deriv is being taken of a nf method and if a model can be located within the environment of the nimbleFunction. If so, and in the circumstances mentioned above, we save model values at the start of `nimDerivs_nf` into a list in an environment within the nimbleFunction environment and restore before leaving `nimDerivs_nf`. 

@perrydv feel free to take a look, but I'm happy to proceed without your review.

### DIFF
--- a/packages/nimble/R/nimbleFunction_Rderivs.R
+++ b/packages/nimble/R/nimbleFunction_Rderivs.R
@@ -668,28 +668,31 @@ nimDerivs_nf <- function(nimFxn = NA, order = nimC(0,1,2),
 
   ## Initialize information for restoring model when needed
   e <- environment(derivFxn)
-  ## Detect if a model is involved
-  isModel <- sapply(names(e), function(x) is.model(e[[x]]))
-  if(any(isModel)) {
-      if(!exists('restoreInfo', e)) {
-          ## First nimDerivs call: save model state and initialize derivative status.
-          e$restoreInfo <- new.env()
-          wh <- which(isModel)
-          if(length(wh) != 1) stop("nimDerivs_nf: unexpectedly found no or multiple models.")
-          e$restoreInfo$model <- e[[names(e)[wh]]]
-          e$restoreInfo$currentDepth <- 1
-          e$restoreInfo$deepestDepth <- 1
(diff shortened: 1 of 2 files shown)


perrydv:
@paciorek This makes sense for immediate practical purposes but might not be sufficiently general long-term.  The nimDerivs call might not end up using the model, in which case copying should be harmless but unnecessary (but ok since this will primarily be for debugging).  And in principle there could be multiple models (although I'm not sure I've seen a use case doing that).  And the nimDerivs might call a function that calls a method of another nimbleFunction that calls model$calculate, in which case the model wouldn't be found in the environment from which nimDerivs is called. That would be a weird case because usually the outermost `nimDerivs` will need the `model`, `updateNodes`, and `constantNodes` arguments, in which case the model would then be in the local environment anyway, but it is conceivable these would not be needed.  How about if we *require* the `model` argument to `nimDerivs` if a model will be used anywhere in the called functions (even if `updateNodes` and `constantNodes` are not needed) and then simply use that provided argument as the source for copying and restoring?   But if you want to go with what you drafted so we can move ahead with testing steps of interest and revisit if necessary later, :+1:.

paciorek:
I'm fine with requiring the model, but would we enforce that? Without thinking hard about it, it seems that enforcing this would run into the same issues you've identified with my approach in terms of us not realizing that the relevant model is in the environment of some other nimbleFunction. I guess we could just document that if you are using nimDerivs where a model is used somewhere you should provide the model, and if you don't then we don't guarantee correct behavior in terms of restoring model state.

perrydv:
Yeah, that's similar to updateNodes and constantNodes.  You have to use them correctly.

paciorek:
I've switched to using a user-provided model. I did leave in a check for a model in the nimbleFunction environment, so we will warn users if they don't provide the model in most of the cases that would probably arise.

PR #1096, opened 2021-02-05
allow nimNumeric, nimMatrix and nimArray into models

perrydv:
Not sure whey these were being error-trapped out of model code.  It appears they can work if they are let through error-trapping and also on the "never replace" (i.e. never make into a lifted node) list.

Am I forgetting any reason not to support these?  We might want to add some tests before merging.

### DIFF
--- a/packages/nimble/R/BUGS_BUGSdecl.R
+++ b/packages/nimble/R/BUGS_BUGSdecl.R
@@ -92,9 +92,11 @@ nimbleOrRfunctionNames <- c('[',
                             paste0(c('d','r','q','p'), 't'),
                             paste0(c('d','r','q','p'), 'exp'),
                             'nimC', 'nimRep', 'nimSeq', 'diag',
+                            'nimNumeric','nimMatrix','nimArray',
                             'length')
 
-functionsThatShouldNeverBeReplacedInBUGScode <- c(':','nimC','nimRep','nimSeq', 'diag')
+functionsThatShouldNeverBeReplacedInBUGScode <- c(':','nimC','nimRep','nimSeq', 'diag',
+                                                  'nimNumeric', 'nimMatrix', 'nimArray')
 
 #' BUGSdeclClass contains the information extracted from one BUGS declaration
 BUGSdeclClass <- setRefClass(


danielturek:
@perrydv In my mind, these were only for nimbleFunction DSL usage, not for in models, but we've had 2 cases recently of users using `numeric` or `matrix` in model code, so I guess why not?  That said, in those cases I feel as thought neither user actually **wanted** a multivariate deterministic node declaration, as it would bog down the graph with the multivariate node, but I guess it's worth supporting (versus not supporting, for no good reason).

Also noting, the earlier user (who used `numeric` in the model) truthfully didn't want it at all, as later the individual elements of the (`numeric`-declared) vector were assigned into with specific (scalar) values.

paciorek:
Thoughts on what the use cases are for these? I'm not seeing, but haven't thought about it carefully.

perrydv:
@paciorek It was this case in a message to nimble-users on 5 Feb 2021.  The user preferred to have a compact declaration to populate an entire matrix at once:

```
code1 <- nimbleCode({
  p1 ~ dunif(0,1)
  p2 ~ dunif(0,1)
  # Version 1, this produces an error:
  P[1:2,1:2] <- matrix(c(p1*p2,p2, p1,p1), nrow = 2, ncol = 2)
  # Version 2, this works:
  # P[1:2,1] <- c(p1*p2,p1)
  # P[1:2,2] <- c(p2,p1)
  x[1,1:2] <- x0[]
  for(t in 1:4){
    x[t+1,1:2] <- P[1:2,1:2]%*%x[t,1:2]
    y[t] ~ dnorm(x[t+1,1], var = p1*p2)
  }
})
Rmodel <- nimbleModel(code1,
                      dimensions = list(x = c(5,2)),
                      data = list(y=c(6,8,11,12), x0 = c(3,5)),
                      inits = list(p1 = 0.5, p2 = 0.5))
mcmc.out <- nimbleMCMC(Rmodel, niter = 100, nchains = 2, monitors = c("p1", "p2"))
```

paciorek:
Ah, I guess that makes sense. I do worry that users will be inclined to populate a matrix or vector of constants in model definition rather than via `inits`, with resulting inefficiency. We see this in users coming from BUGS/JAGS. But I guess we also see that in loops, e.g.,

```
for(i in 1:n) 
  z[i] <- 0
```

I wonder if we want to issue a warning when we detect deterministic assignment where RHS is constant(s). Just brainstorming; haven't thought it through.

perrydv:
The testing build failures appear to be Travis-CI glitches.  I restarted them.

perrydv:
Yeah that's a good point @paciorek .  OTOH the inefficiency might not be that much.  If there's a lot of initialization within the model, it would involve model declarations and nodes to process and compile, but presumably they wouldn't be used much in algorithms.

The case of `nimNumeric` really seems redundant because simply `c` (aka `nimC`) should work.  The only need for `nimMatrix` or `nimArray` is to structure the values.  But it could be confusing to leave out `nimNumeric`.

I guess there were earlier reasons we left these out of the models, but right now it seems reasonable to me to make them work.

paciorek:
Interestingly, the nimble-users message from Feb 8 also seemingly wanted to use `matrix()` in a model. Her message wasn't about about that issue, though. Side note: she nicely used a user-defined nimbleFunction that called `matrix` to get around the limitation discussed here.

paciorek:
@perrydv ok to merge?

perrydv:
Ok with me, thanks @paciorek

PR #1098, opened 2021-02-13
Clean up commented-out code from no-longer-experimental changes

perrydv:
We often comment-out code as we are working on changes.  At the time, it is to make it easier to see the changes without more git branching and to retain in-file documentation what was changed in case of new errors.  This PR removes a bunch of such changes that are now old and don't seem helpful.  Feel free to put anything back that is useful to understanding current code.  I tread more lightly on MCMC-related code.

### DIFF
--- a/packages/nimble/R/BUGS_getDependencies.R
+++ b/packages/nimble/R/BUGS_getDependencies.R
@@ -1,12 +1,7 @@
-
-
-
 ## returns a list of all deterministic dependents up to the first stochastic dependent,
 ## omitting any nodes in 'omit', or down the path of 'omit' nodes
 ## works only in terms of vertex IDs, as in the igraph object.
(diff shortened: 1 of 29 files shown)


paciorek:
@danielturek can you check and sign off?

danielturek:
@paciorek @perrydv Thanks for the chance to review this.  I added one thing back in, the commented out code in the RW samplers, with methods for "get scale history expanded", which is quite useful sometimes, it changes the recorded "scale history" (and prop cov history for the block sampler) to be returned on the clock of MCMC iterations, rather than on the clock of adaptation cycles, allowing direct comparison with the actual MCMC samples, which is useful sometimes. The code remains commented out, so that it's not compiled every time the RW or RW_block sampler is used, but I'd like to keep it in there for when it's needed.

I made this change, and put that (commented) code back in.  Otherwise, this PR LGTM.

PR #1099, opened 2021-02-13
WIP: add nimbleSMC testing to travis for nimble

paciorek:


paciorek:
Note to remember: somehow the gold file for nimbleSMC test-filtering has an extra line in it that is no longer printed (even by the currently released nimbleSMC) when running test-filtering. This means that activating SMC testing currently in devel will cause devel to fail on Travis.

We need to:

1. update the gold file when we next update nimbleSMC on CRAN
2. activate nimbleSMC testing on nimble devel branch.

paciorek:
Comments about how testing of nimble and nimbleSMC should (I think) be synchronized:

- Travis testing of nimble should use the CRAN nimbleSMC. This is because we want changes to nimble not to break the released nimbleSMC.
- Travis testing of nimbleSMC should use the CRAN nimble. This is because we want changes to nimbleSMC to work with the released nimble.
- If we are jointly changing nimble and nimbleSMC in Git, it gets tricky. E.g., if we change a nimble API and need to update how nimbleSMC uses nimble. Presumably we deal with this on an ad hoc basis, testing locally.

paciorek:
This is ready to go. The failure is because of the extra line issue mentioned two comments above. Merge into devel after nimbleSMC next version released on CRAN.

PR #1100, opened 2021-02-17
in newMatrixClass, get returnType as the scalar of input type

perrydv:
This tweaks a single line in C++.

For the newMatrixClass, which is used to implement nimMatrix via nimNewMatrixD (and others), it sets the return type from the scalar type of the input, not simply defaulting to double.

This does not fix a bug.  This change would be convenient for AD, so I want to see if it causes an problems via testing.  The change seems slightly more correct than what was there, although nearly equivalent.

### DIFF
--- a/packages/nimble/inst/include/nimble/nimbleEigen.h
+++ b/packages/nimble/inst/include/nimble/nimbleEigen.h
@@ -1639,7 +1639,8 @@ template<typename Index, typename DerivedInput>
   int dim1, dim2, totalLength, inputLength, inputRows;
   bool init; // would be a bit silly to call with init = FALSE, but it is allowed to simplify code generation
   bool recycle;
-  typedef double result_type;
+  //  typedef double result_type;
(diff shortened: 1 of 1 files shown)


paciorek:
@perrydv ok to merge?

perrydv:
@paciorek No, let's not put this in the release.  It doesn't fix anything that's currently broken.  It is for a future need for AD work (but not directly in AD code).  And I need to do more in this vein so I'd like to keep this open for changes and testing (to make sure I don't break something with general changes that will facilitate AD).

paciorek:
@perrydv thoughts on this PR wrt 1.0.0?

perrydv:
@paciorek This is a change that is in ADoak and so is going into 1.0.0 already.  The idea with this PR was to get the change into devel (it is not specific to the AD code, but is needed to support it) to see if anything bad happened, but we never did merge it, so it is moot.

paciorek:
Closing as moot.

Issue #1076, opened 2020-10-17
separate if statements fail in definition-time if-then-else


paciorek:
This is weird. Haven't delved into it except to create the reprex. 

This fails, but the nested if-else version is fine.

```
code = nimbleCode({
    if(A && B) {
        y~dnorm(0,1)
    }
    if(!A && B) {
        y ~ dunif(0,1)
    }
    if(A && !B) {
        y~dgamma(1,1)
    }
    if(!A && !B) {
        y <- 7
        }
})

A = B = TRUE
m = nimbleModel(code)
```

```
defining model...
Error in code[[i]] : subscript out of bounds

Enter a frame number, or 0 to exit   

1: nimbleModel(code)
2: md$setupModel(code = code, constants = constants, dimensions = dimensions, 
3: codeProcessIfThenElse(code, constants, userEnv)
4: codeProcessIfThenElse(code[[i]], constants, envir)
```

```
code = nimbleCode({
    if(A && B) {
        y~dnorm(0,1)
    } else if(!A && B) {
        y ~ dunif(0,1)
        } else if(A && !B) {
            y~dgamma(1,1)
        } else {
            y <- 7
        }
})

A = B = TRUE
m = nimbleModel(code)
```

paciorek:
It looks like `else` is required:

This works, but fails if remove the `else {}`.

```
code = nimbleCode({
    if(A) {
        y~dnorm(0,1)
    } else {}
   z <- 7
})

A = FALSE
m = nimbleModel(code)
```

PR #1104, opened 2021-03-07
fix error when model if-then-else has no else

paciorek:
This fixes issue #1076.

@perrydv I think this is just a long-standing oversight.


perrydv:
Yes, agreed it is an old oversight.  Thanks @paciorek. 👍 to merge.

Issue #1084, opened 2020-12-03
Code checking incorrectly triggered for particular nimbleFunctionList usage


perrydv:
The following incorrectly generates a warning that `foo` needs to be defined:

`myNimbleFunctionList[[1]]$foo(x)`

where `foo` is a method in the virtual nimbleFunction for `myNimbleFunctionList`.  The following does not generate the warning:

`i <- 1`. `myNimbleFunctionList[[i]]$foo(x)`, nor does

`myNimbleFunctionList[[1]]$run(x)`.

paciorek:
Here is a reprex, which throws a warning about `foo` and `foo2` but not `foo3`. I will look into this.

```
nfbase= nimbleFunctionVirtual(
    name = 'nfbase',
    run = function() {returnType(double(0))},
    methods = list(
        foo = function() { }
    )
)

nfextend = nimbleFunction(
    setup = function() {},
    run = function() {
        returnType(double(0))
        return(3)
    },
    methods = list(
        foo2 = function() {},
         foo3 = function() {}
    )
)

nfmore = nimbleFunction(
    setup = function() {
        fxns = nimbleFunctionList(nfbase)
        fxns[[1]] <- nfextend()
    },
    run = function() {
        returnType(double(0))
        fxns[[1]]$foo()
        fxns[[1]]$foo2()
        i <- 1
        fxns[[i]]$foo3()
        out = fxns[[1]]$run()
        return(out)
    })
```

paciorek:
Notes on walking the parse tree:

```
out1 = nimble:::RparseTree2ExprClasses(quote(foo$cc(7)))
out2 = nimble:::RparseTree2ExprClasses(quote(foo[[2]]$cc(7)))
out3 = nimble:::RparseTree2ExprClasses(quote(foo[[i]]$cc(7)))
out4 = nimble:::RparseTree2ExprClasses(quote(foo[[x+3]]$cc(7)))
out5 = nimble:::RparseTree2ExprClasses(quote(foo[[myfun[[2]]$innercc()]]$cc(7)))
```

When detect `$`, the 2nd arg in the exprClass is the method name.

paciorek:
It looks like this was addressed, probably in PR #1035 but this was never closed.

PR #1105, opened 2021-03-13
robustify checking of uncompilable elements in nfs

paciorek:
Formerly we were pulling out the names in nimbleFunction run code and relying on ordering of those names to try to find nf methods and not warn about those in `nf_checkDSLcode` (since we can't know what the methods of potentially not-yet-defined nfs are). As shown in issue #1084 this check was not robust.

This PR converts the run code to an `exprClass` and walks the tree to find methods based on finding `$`. 

This fixes issue #1084

### DIFF
--- a/packages/nimble/R/RCfunction_core.R
+++ b/packages/nimble/R/RCfunction_core.R
@@ -78,9 +78,11 @@ nfMethodRC <- setRefClass(
             if(code[[1]] != '{')
                 code <<- substitute({CODE}, list(CODE=code))
             ## check all code except nimble package nimbleFunctions
-            if(check && "package:nimble" %in% search()) 
-                nf_checkDSLcode(code, methodNames, setupVarNames)
             generateArgs()
+
+            if(check && "package:nimble" %in% search()) 
+                nf_checkDSLcode(code, methodNames, setupVarNames, names(arguments))
+
             generateTemplate() ## used for argument matching
             removeAndSetReturnType(check = check)
             ## Includes for .h and .cpp files when making external calls:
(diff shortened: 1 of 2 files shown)


perrydv:
@paciorek I'm not sure if this covers recursion through the exprClass tree thoroughly.  I think `exprClasses_processSpecificCalls` is the simplest example we have.  Would it make sense to imitate that more closely?

perrydv:
Also I meant to check if any new code checking like this is toggled by an option.  That would be nice so that if it misfires and needs to be disabled for some use case, we can tell a user how to turn it off.

paciorek:
@perrydv I looked at `exprClasses_processSpecificCalls`. It's not dissimilar to the code I wrote. For the recursion, it uses a for loop over args rather than my `lapply`. Other than that. there is not actually much to `exprClasses_processSpecificCalls`. It calls out to specific handlers to modify code while my code is looking for `$`.

Also, just realized myself that the run-code checking is behind the `checkNimbleFunction` option, so that is dealt with.

paciorek:
@perrydv just wanted to ping you on the recursion question here.

paciorek:
@perrydv I'd like to merge this in for the release; can you take a look?

perrydv:
OK @paciorek I probed this with a couple of kinds of cases where I've seen issues traversing parse trees in the past, and it looks ok.  I do think it would be good to change line 84 to:

```
if(check && "package:nimble" %in% search() && isTRUE(nimbleOption("checkDSLcode")))
```

so there can be a failsafe (set the option FALSE) in case of a misfired code check.

paciorek:
`check` is already based on the `checkNimbleFunction` nimbleOption, so I think the failsafe is already taken care of.

Issue #1035, opened 2020-06-26
Two problems arising from nimbleFunctionList methods with no arguments


perrydv:
In the example below:

1. The static code processing to give warnings misfires, giving an unnecessary warning.
2. Name mangling of `j` does not reach into the indexing of the nimbleFunctionList, only when invoking the nullary methods.  This may be the case of (1).

```
library(nimble)

nfv <- nimbleFunctionVirtual(
  run = function() {returnType(void())},
  methods = list(
    f0 = function () {returnType(void())},
    f1 = function (x = double()) {returnType(void())}
  )
)

nf <- nimbleFunction(
  contains = nfv,
  setup = TRUE,
  run = function() {print("run")},
  methods = list(
    f0 = function() {print("f0")},
    f1 = function(x = double()) {print("f1")}
  )
)

usenf <- nimbleFunction(
  setup = function() {
    nfList <- nimbleFunctionList(nfv)
    for(i in 1:3)
      nfList[[i]] <- nf()
  },
  run = function(j = integer()) {
    nfList[[j]]$run()
    nfList[[j]]$f0()
    x <- 1.2
    nfList[[j]]$f1(x)
  }
) ## Misfired warning about needing to define j, possibly because of the next issue

usenf1 <- usenf()
cusenf1 <- compileNimble(usenf1)
printErrors()
## for both f1 and run, name mangling of j did not occur.
```

A workaround is to assign `j` to a local variable and then use that as the list index.

paciorek:
Incorrect warning about `j` is being fixed in PR #1105 , as it relates to issue #1084 .

Lack of name-mangling appears to be a bug in `nf_substituteExceptFunctionsAndDollarSigns`, where we check `if(length(code) == 1)`. However something like `nfList[[j]]$f0()` is of length 1, so we never recursed into it to find `j` to name-mangle.

I am fixing in `issue_1035`. PR to follow.

PR #1107, opened 2021-03-15
fix name-mangling in context of nimbleFunction lists

paciorek:
This fixes issue #1035, which fixes an error in name-mangling in creation of C++ related to nimbleFunction lists.


perrydv:
I think this is ok.  You're relying on `&&` to not evaluate the second condition if the first condition is FALSE.

paciorek:
Yes, my understanding is that is reasonable R practice.

PR #1109, opened 2021-03-20
Add univariate elliptical slice sampler

danielturek:
The title says it all.  Tested and works for both univariate and multivariate normal nodes, with same numeric results as previous `ess` sampler.

@perrydv I'm happy to merge this as-is, but I think it raises some issue with `getParam` and generated C++ code.  See lines 732-734 (with the comment "this should work"), which if commented in, triggers a segfault at runtime.


danielturek:
Fixes issue number 108 of nimbleCoreTeam repository.

danielturek:
@perrydv I filed an issue demonstrating this segfault I mentioned, it's issue #1111.

danielturek:
@paciorek Please take a look at this PR, for adding the univariate case to the ess sampler.

I'm fine with merging it, but wanted to ask for your review.

paciorek:
@danielturek I don't think this works correctly in your use case of concern of having a scalar and multivariate ESS in the same MCMC. `temp` ends up being a scalar in the C++ code and while the MCMC runs, it gives incorrect answers for the multivariate sampling. Presumably we need  two different `temp` variables, one a scalar and one a vector, defined in the setup. And then an `if` in the setup that uses the appropriate `temp` depending on whether `d` is one or not.

danielturek:
@paciorek Egg on my face.  Apologies, I try to not have errors like that.  Should be fixed now (pushed changes), which now uses a `nimbleFunctionList`.

paciorek:
Minor comment - should we have the name of the list and the related nf's contain the wording 'getMean' to be clear what they do? 'essNF' is rather generic. Just a thought; I'm fine with it as is.

danielturek:
@paciorek I considered the same, at the time of writing, and opted to skip the `methods` list and just use the `run` function, just to save some lines of code.  I'm fine with any such changes, but I guess I figured leave as-is.  Let me know if you feel strongly, I would not be offended.

paciorek:
Oh, I meant something like essNFList -> ess_getMean, essNF_univariate -> ess_getMean_univariate, not whether there are non-run methods. No strong feelings.

paciorek:
@danielturek I think I will go ahead and merge this in, as any name changes would not be user-facing anyway.  Looks like there are conflicts I'll need to resolve first.

PR #1110, opened 2021-03-20
handle additional cases of multiple/overlapping node definitions

paciorek:
This adds error trapping to address NCT issue 252. Up to this change we only detect when two identical nodes are defined twice, not when two nodes overlap (e.g., two multivariate nodes or a multivariate and scalar node).

@danielturek @perrydv Let me know of any further thoughts about removing the message mentioned in issue 252. For the moment I've commented it out. 

```
                    "If your model has macros or if-then-else blocks\n",
                    "you can inspect the processed model code by doing\n",
                    "nimbleOptions(stop_after_processing_model_code = TRUE)\n",
                    "before calling nimbleModel.\n"
```


perrydv:
@paciorek I'll just suggest per usual that we protect this new error-trap with a nimbleOptions toggle so that we can give people a way to get around it if it ever misfires.

PR #1115, opened 2021-03-25
Added samplesAsList option to runMCMC and nimbleMCMC

danielturek:
The title says it all.  Adds a new option `samplesAsList` to `runMCMC` and `nimbleMCMC` functions, with default value `FALSE`.  When `samplesAsList = TRUE`, the `samples` return object will be a named list, with named elements corresponding to variables being monitored.  Details:

The number of MCMC iterations will be the first dimension of each named element.  Thus, for 10 MCMC iterations, scalar variables are returned as a length-10 vector, length-3 vector variables are returned as a 10x3 arrays, and 3x4 array variables are returned as 10x3x4 arrays.  This matches the behaviour of `jagsUI`.

To achieve this, I made changes to the `modelValues` utility functions `as.list.modelValuesBaseClass` and `as.list.CmodelValues`, adding `scalarize = TRUE` to the `dimOrLength` call.  See the modified code.  I think this makes more sense, and my guess is that the missing-ness of this only reflects the total lack of usage of these `as.list` functions.  But perhaps I'm missing a part of the bigger picture here.

Also, I had to wonder, why the duplicate definitions of these functions?  Would it be simpler, and functionally identical, to say:
```
as.list.modelValuesBaseClass <- as.list.CmodelValues <- function(…) { ... }    ?
```

Our behaviour does **not** mimic that of `jagsUI`, in the case of multiple chains.

In the case of 3 chains, we will return `samples` as a list of length 3 (elements named "chain1", "chain2", and "chain3"), where the first element is a named list, with elements corresponding to the samples (of each variable) from chain 1, etc ....  In contrast, `jagsUI` will return a named list as before, with the samples from each chain concatenated.  This is a slight difference in functionality from `jagsUI`, which we could discuss.

This approach can be wasteful in terms of memory when `samplesAsList = TRUE`, since `runMCMC` will still collect samples **both** in the forms of an array, and also as a list (using this update).  This is because the **array** form of the samples is used internally for both the (optional) summary, and also the (optional) WAIC calculations, even if not returned to the user.  I'm ok with this, but flagging it for discussion.

### DIFF
--- a/packages/nimble/R/MCMC_run.R
+++ b/packages/nimble/R/MCMC_run.R
@@ -29,6 +29,8 @@
 #'
 #' @param samplesAsCodaMCMC Logical argument.  If \code{TRUE}, then a \code{coda} \code{mcmc} object is returned instead of an R matrix of samples, or when \code{nchains > 1} a \code{coda} \code{mcmc.list} object is returned containing \code{nchains} \code{mcmc} objects.  This argument is only used when \code{samples} is \code{TRUE}.  Default value is \code{FALSE}.  See details.
 #' 
+#' @param samplesAsList Logical argument.  If \code{TRUE}, then samples are returned as a named list object, where each element corresponds to the monitored variable of that name.  List elements are vectors or arrays.  The first dimension of each element corresponds to the number of samples collected, and the subsequent dimensions of each element will match the dimensions of the variable itself.  When multiple chains are run, a list consisting of \code{nchains} named lists is returned, with each nested list containing the samples collected during one chain.  This argument is only used when \code{samples} is \code{TRUE}, and cannot be using concurrently with \code{samplesAsCodaMCMC = TRUE}.  Default value is \code{FALSE}.
+#' 
 #' @param summary Logical argument.  When \code{TRUE}, summary statistics for the posterior samples of each parameter are also returned, for each MCMC chain.  This may be returned in addition to the posterior samples themselves.  Default value is \code{FALSE}.  See details.
 #'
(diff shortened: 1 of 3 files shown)


paciorek:
I am worried about the memory use but given it's not the default maybe ok. 

Would it be possible to create the list form only after doing WAIC and the summary and to first remove the matrix form before creating the list form so as to avoid the extra memory?

Note that once we update WAIC with Josh's work, we will note need the array form for that.

perrydv:
I like the idea of waiting to create the list until after WAIC, in the event WAIC is wanted but coda matrix format is not wanted.

I think we should support both list and coda returns. That's a hassle but the reason is that users may have different needs for different outputs and if they are using nimbleMCMC there is no easy way to get back in to request a different format.

The drafted new argument would change the API.  If a user has existing code with unnamed arguments, what used to be their summary argument will now be their samplesAsList argument.  Both are logicals so I'm not sure there is a good error-trap.  One alternative would be to add the new argument at the end, although that is somewhat inconsistent.  Another would be to convert the samplesAsCodaMCMC argument to a samplesFormat or Format argument with a set of options including TRUE/FALSE (which would provide backward compatibility) and text label options like "Coda" or "List".  If multiple options are provided, multiple formats can be returned.  To error-trap old uses, we can put a "samplesAsCodaMCMC" argument last, in case existing code uses it as a named argument, and handle it with a warning to update their code.  I'd prefer this approach.  It would become redundant with the "samples" argument (because with text labels, "none" could be an option), but keeping both would avoid breaking code, unless you can see a way to consolidate them and catch cases of existing uses.

danielturek:
@perrydv Fair questions about backwards-compatibility.  I'm ok with tabling this PR for now, and waiting until WAIC is finalized.  Then reaching some decisions about arguments and user interface.

PR #1116, opened 2021-03-25
return no nodes when no latent nodes and latentOnly is TRUE

paciorek:
This fixes NCT issue 212. There was an oversight that if there are no latent nodes, then nothing was being done to set elements of `validValues` to `FALSE`.


paciorek:
@danielturek I'd like to merge this in for the release; can you sign off on this?

danielturek:
@paciorek I gave it a try for a few cases, and seems to work fine.  Changes look good to me.

PR #1117, opened 2021-03-29
check for missing names in data and inits lists

paciorek:


paciorek:
This fixes NCT issue 179.

PR #1120, opened 2021-04-10
check for incomplete declarations in model code

paciorek:
As seen in a recent user post, we were not error trapping cases like this (including such non-declarations inside `if` or `for`).

```
mycode <- nimbleCode({
    model
    y~dnorm(0,1)
})
m <- nimbleModel(mycode)
```

This PR catches such cases. Initially I thought we should be looking for cases where `length(code)==1` in `codeProcessIfThenElse`, but `{}` is legit and of length 1, albeit useless, and `f()` is (and should be) trapped elsewhere.

@danielturek @perrydv this seems straightforward, but a second eye would be nice, particularly as all models use this code.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -190,6 +190,9 @@ modelDefClass$methods(setupModel = function(code, constants, dimensions, inits, 
 
 codeProcessIfThenElse <- function(code, constants, envir = parent.frame()) {
     codeLength <- length(code)
+    if(class(code) == "name")
+        stop("Incomplete declaration found: '", deparse(code), "'.")
+        
     if(code[[1]] == '{') {
         if(codeLength > 1) for(i in 2:codeLength) code[[i]] <- codeProcessIfThenElse(code[[i]], constants, envir)
         return(code)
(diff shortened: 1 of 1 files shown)


paciorek:
@danielturek @perrydv I think this is the last PR for 0.11.0 and I've gone through all my pre-CRAN steps, so if one of you has a chance to take a quick look at this today or tomorrow, that would be great.

perrydv:
@paciorek This change makes sense to me but I have a few minor comments.  This function is not a super logical place for this check but I guess it's ok there.  Should we still have a trap with `if(codeLength == 1)` and return some useful message?  It looks like in that past that just returned the code unprocessed.  Looks like your trap will catch a single stray name, but what if it is a stray `"junk"` or `5` instead of `junk`?  Would the class then be character or numeric instead of name?  A couple of other code objects that are not names but have length 1 are `quote({})`, which I think we'll get if someone writes `for(i in 1:10) {}` by accident, and `foo()`, which might have as a typo line by itself.  I think we can prefer `is.name(x)` to `class(x) == "name"`.

paciorek:
Thanks, this is helpful. I'll think through those various cases.

Agreed this is a funny place for the check but it is where we first encounter the potential issue. 

`5` is already handled via separate error trapping. I suspect `"junk"` is as well but need to check.

I had decided not to simply trap `codeLength == 1` because  `{}` is in some sense legit and because trapping `foo()` is done later with a more informative message.

perrydv:
Sounds good.  If you've thought through the cases then it looks good to me.

PR #1124, opened 2021-04-24
WIP: pass `where` into `nf_checkDSLcode`

paciorek:
This addresses the issue raised by @dochvam in email that `nimbleEcology` install was triggering warnings from `nf_checkDSLcode` when a nimbleFunction (without setup code, but should be the situation regardless of setup code) calls another nimbleFunction.

The issue is that `nf_checkDSLcode` was looking its enclosing environment (the nimble namespace) and during package install for nimble-depending packages, that doesn't end up looking in the relevant environment containing the functions in the package.

I have not thought through yet whether this change could have unintended side effects. That said, the passing of `where` is now consistent with what we are doing when we call `nimbleFunction`.

### DIFF
--- a/packages/nimble/R/RCfunction_compile.R
+++ b/packages/nimble/R/RCfunction_compile.R
@@ -146,16 +146,16 @@ RCvirtualFunProcessing <- setRefClass(
 RCfunction <- function(f, name = NA, returnCallable = TRUE, check, where = NULL) {
     if(is.na(name))
         name <- rcFunLabelMaker(envName = environmentName(where))
-    nfm <- nfMethodRC$new(f, name, check = check)
+    nfm <- nfMethodRC$new(f, name, check = check, where = where)
     if(returnCallable)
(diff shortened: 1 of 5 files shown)


perrydv:
@paciorek I think this makes sense but am a bit worried about unintended consequences.  What are creative cases where it could break?  Are there cases where during secondary package install it *should* look in the nimble namespace to find valid DSL function names and would no longer do so?

paciorek:
Given the enclosing env't for the secondary package namespace should be the nimble namespace, I think this should be fine, but I should check with the particular example to verify that I am thinking about this correctly.

paciorek:
I've verified that the enclosing environment for the secondary package ns is the imports of the secondary package, which would include nimble. I'm going to merge this in and then make sure that packages that depend on nimble can be installed.

paciorek:
I've checked that nimbleEcology and nimbleSMC install and that example code runs (the nimbleEcology vignette code and nimbleSMC test code). Will merge and during pre-release checks, will check reverse dependencies of nimble.

PR #1127, opened 2021-05-20
added calculate call to posterior_predictive_branch sampler

danielturek:
Fixes the `posterior_predictive_branch` sampler.


danielturek:
@paciorek testing passes now.

PR #1132, opened 2021-06-12
WIP: add lkj tranform to parameter transform system

paciorek:
In conjunction with PR #993 , this adds lkj transform to the parameter transform system.

There is testing, but it won't run, because the lkj distribution will be first merged into devel via PR #993 and is not actually available in ADoak or ADoak-addLKJ yet. The tests should pass once everything is integrated together, as I did check them under a hacked version of things. 

One good thing to do would be to compare NIMBLE HMC on a basic lkj test case to Stan. I have Stan code for this and have compared Stan to NIMBLE's non-HMC lkj samplers in the lkj branch that will be merged into devel.


danielturek (approved):
@paciorek This addition looks good.  Thank you for the clear code comments throughout, they help a lot with following what's going on.

PR #1133, opened 2021-06-16
fix issue 1131 by truncating result of deparse in dynamic index warning

paciorek:
This uses a new `safeDeparse` utility


paciorek:
Omitting in favor of PR #1167

PR #1136, opened 2021-06-22
WIP: extend and robustify WAIC

paciorek:
This PR implements all the work Josh and I have done to do online WAIC calculation and add marginal WAIC and grouped WAIC.

It takes the code Josh wrote for his thesis (which was all embedded in `buildMCMC`) and reorganizes to create a `buildWAIC` nf with relevant methods per dev meeting discussion 2021-02-08. Also changes some naming and reorganizes the return list of WAIC-related quantities. Finally, it sets up `configureMCMC`, `buildMCMC`, `runMCMC` and `nimbleMCMC` to work with the new implementation.

Off-line WAIC with `calculateWAIC` is retained for back compatibility (if one sets `enableWAIC` and turns off online WAIC via `control$onlineWAIC = FALSE`.

Still to do: 

- update the manual 
- add tests. 

Open questions that I'd like feedback from Perry and Daniel on eventually (not needed at the moment):
- the return value is an extensive list containing some pieces that are not used in many cases; other approaches?
- naming of return elements, control list elements, and methods could use a second eye
- should I be predefining C++ code for the `waicList`? Perhaps related, if I recompile the MCMC without recompiling the model, I get a compilation error related to the `waicList`.
- the WAIC is a single nimbleFunction not a nimbleFunctionList. I'm not entirely clear if there is a need for it to be a nimbleFunctionList. 
- how should we modify the printing out of MCMC configuration information to mention WAIC (and possibly if marginal and grouping is used)
- to allow users to assess convergence of WAIC, we may want to allow users to get WAIC for each chain when using multiple chains. For now, I get an aggregated WAIC. Could also let users specify that they want to see how WAIC varies within increasing number of MCMC samples (similar to use of `convergenceSet`) but that is going to make the inputs and outputs even more complicated.


paciorek:
One thing to think about is that while online calculation is great, a user has to commit to their burnin in advance. We could comment on this in our docs. If we really wanted to we could also add marginal WAIC and grouping to the `calculateWAIC` "post-hoc" functionality.

paciorek:
@guhauhsoj this PR takes your work in the `update_waic` branch and rearranges it to use a nimbleFunction specific to WAIC that is used within `buildMCMC`. I did a bunch of renaming and some reorganizing of inputs and outputs, but the core calculations are as you had them. That said, there's a chance I might have introduced an error. If you have a chance and interest please take a look and see if you have any comments.

I also added some text to the chapter_MCMC.Rmd of the manual and info to the roxygen related to WAIC.

paciorek:
Some thoughts from a discussion amongst Perry, Daniel, Chris:

- Piling everything into a big output list seems not optimal. How about having a small core list of main outputs, obtained by `getWAIC` and then a different `get` method that returns the more voluminous stuff in a separate list format. 
- Another suggestion was to provide an argument to `runMCMC` and perhaps `nimbleMCMC` controlling how much was returned, but I think we tentatively decided not to 'pollute' these functions with more arguments. At least for `runMCMC` a user can still call methods of the mcmc object that could call out to methods of the waic object. And we've always said that `nimbleMCMC` is only for high-level use.

paciorek:
@danielturek What would you think about one additional argument to `runMCMC` (but not to `nimbleMCMC`) that says whether the user wants per-chain WAIC or an aggregated WAIC. This would allow a user to get at uncertainty in WAIC in terms of posterior mixing if desired. Not sure what it would be called. Perhaps `perChainWAIC`. It would default to FALSE (i.e., aggregated WAIC).

Another option would be to change the  current`WAIC` arg (currently a logical) to become a character string, one of "none", "all-chains", or "per-chain".

paciorek:
I just pushed a commit that splits the output list into a core list and a 'details' list. @danielturek @perrydv when you get around to looking at this (no hurry at all), see what you think. 

I have not yet updated the roxygen or manual to reflect splitting the output given I'm not sure what we'll ultimately decide on.

danielturek:
@paciorek I'm still wading / working through this (there's obviously a lot to look at).  Brief comments for now, before the release, I wonder if maybe some review (and possible changes) to the naming conventions might be worthwhile - really  thinking about simplicity and consistency.  Ideas, among others, that came to mind:

Changing `waicControl` into `controlWAIC`, to better match with `enableWAIC`, and also the original `control` list.

Otherwise - Can we take this opportunity to *fully* remove the `burnIn` argument from the `calculateWAIC` method?  It's been unsupported and deprecated essentially since it was released, then immediately changed to `nburnin`.

Why the re-naming in the `calcualteWAIC` method, of `logAvgProb` to `logAvgProbCalc`, and `WAIC` to `WAICCalc`?

Why the line
```
contains = nimbleFunctionVirtual()
```
in the `buildWAIC` nimbleFunction?

Keep the `marginalizeNodes` control list element called that, rather than changing it to `latentNodes` in `buildWAIC` ?

danielturek:
Other aliases we might also add, to the Roxygen documentation: `WAIC`, and `enableWAIC`.

danielturek:
Maybe change the name of the `nItsMarginal` argument to `niterMarginal`, for better consistency with other MCMC arguments?

danielturek:
@paciorek @perrydv

Noting that `buildMCMC` executes the line:
```
waic <- buildWAIC(model, mvSaved, conf$waicControl)
```
regardless of whether WAIC is enabled.  I understand this is necessary for compilation of the MCMC, that the `waic` object be properly created.  But as a result, all of the code in `buildWAIC` is **compiled for every MCMC**, even if we're not using WAIC.

Could it instead look like (sorry if I mess up some syntax):
```
waicList <- nimbleFunctionList(WAIC_virtual_NF)
if(enableWAIC)   waicList[[1]] <- buildWAIC(model, mvSaved, conf$waicControl)  ## same as current
if(!enableWAIC)  waicList[[1]] <- buildDummyFunctionThatCompilesQuickly()      ## new
```
Then, both `buildWAIC` and `buildDummyFunctionThatCompilesQuickly` inherit from `WAIC_virtual_NF`, to make this work, and technically both contain the same methods.

Do we think this is worthwhile?

I'm coming from the angle that so many people seem to use the MCMC, it would be a shame to slow down compilation for *everyone* when only (what seems like) a small subset make regular use of WAIC.

danielturek:
Small preference that `online` and `thin` variables be accessed from the `waic` object using access methods:
```
methods = list(
    getOnline = function() {
        returnType(double())
        return(online)
    }
    getThin = .....   ## same
)
```

This also alleviates the need for the `setupOutputs()` call that's presently in `buildWAIC` setup function, which we basically never use.  And would lead to a simpler implementation of `buildDummyFunctionThatCompilesQuickly` from my previous comment, where the access methods could look like:
```
methods = list(
    getOnline = function() {
        returnType(double())
        return(0)
    }
    getThin = .....   ## same
)
```

danielturek:
I'm not sure, does
```
export(waicDetailsList)
```
need to be added to the `NAMESPACE` file also, in addition to `export(waicList)` ?

paciorek:
Responses to Daniel's helpful comments:

1) change to `controlWAIC`: done
2) removing `burnIn`: done.
3) As far as re-naming in the calculateWAIC method, of logAvgProb to logAvgProbCalc, and WAIC to WAICCalc, I think Josh might have done this. I'll revert the naming. Good catch.
4) As far `contains = nimbleFunctionVirtual()`, I think at one point I thought I needed that. I'll defer until we discuss point #8 below.
5) Keep the marginalizeNodes control list element called that: done.
6) Other aliases: done.
7) `niterMarginal`: done
8) I like the idea of avoiding unnecessary compilation by having a nimbleFunctionList. Let's discuss on Tuesday.
9) I'm fine with getThin() and getOnline. I understand the logic, but I'm guessing there are various other places in the system where we don't have getters.
10) NAMESPACE fix: done.

paciorek:
@danielturek just want to re-raise this with you from a comment above:

What would you think about one additional argument to `runMCMC` (but not to `nimbleMCMC`) that says whether the user wants per-chain WAIC or an aggregated WAIC. This would allow a user to get at uncertainty in WAIC in terms of posterior mixing if desired. Not sure what it would be called. Perhaps `perChainWAIC`. It would default to `FALSE` (i.e., aggregated WAIC).

Another option would be to change the current WAIC arg (currently a logical) to become a character string, one of "none", "all-chains", or "per-chain". Though if we changed the meaning of the WAIC arg, I guess we would want to do the same for that arg to nimbleMCMC.

paciorek:
Daniel and I agreed to add `perChainWAIC` as an arg to `runMCMC` with default of FALSE. This is now in the waic_update_nf branch. Per-chain WAIC only works for offline WAIC as doing it for online would involve a bunch of recoding and would complicate the WAIC output. Plus one can always get per-chain WAIC by simply running a loop.

paciorek:
@danielturek @perrydv This is a (hopefully) full and nearly final draft of WAIC, including implementation of  ideas for user-friendly offline WAIC that we discussed back in July. I still need to look it over to think about corner cases in terms of how a user might access WAIC, but given the short timeline to release I thought I'd point you to it if you'd like to take a look. I suspect Daniel particularly will like to see the full interface. Any feedback or thoughts on cases I might have missed are very welcome. 

Ignore the test failures; I don't expect them to be substantive.

The key changes are:
- online WAIC is the default if one has `enableWAIC=TRUE` in `configureMCMC` or `buildMCMC`. For online WAIC there is now the multitude of different cases we handle (marginal vs. conditional, grouped data nodes vs. ungrouped) that Josh and I worked on and that I put into this branch back in July.
- one can get our former offline behavior with `enableWAIC = TRUE` and `controlWAIC = list(online = FALSE)`. This modality now uses an offline WAIC nf (`buildOfflineWAIC`) that is built within `buildMCMC`. The original WAIC calculation code formerly in `mcmc$calculateWAIC` has been moved to the `run` method of the `buildOfflineWAIC`. 
- I have added a standalone function (not nimbleFunction) called `calculateWAIC` that can be applied to an MCMC **or** a matrix of samples **without** having to enabled the WAIC at the time of building the MCMC. This function calls `buildOfflineWAIC` and runs the resulting compiled nf to do the calculations and return the simple WAIC output that we currently produce. Side note: this will fail gracefully if the needed variables were not monitored in the MCMC. I believe this is the additional functionality we decided we wanted at our July conversation about WAIC.
- When we run multiple chains, one can get WAIC for each chain individually if using offline WAIC per the previous note in this PR.

roxygen and manual are updated for online WAIC but not yet for the new post hoc offline WAIC functionality.

danielturek:
I tried to look at all code in this PR.  Mostly minor comments below.

Looks like changes to `UserManual/src/UserManualRefs.bib` have evidence of an old merge conflict at the very end (line 258)

User manual `chapter_MCMC.Rmd` line 457, ".... or a matrix containing posterior **samples**..."

Line 646 of the same file has the wrong name for the NIMBLE option, it should read "MCMCenableWAIC".

Commented lines of code in MCMC_WAIC.R?  Beginning on like 105, ending on line 158?  Looks like this belongs in nimbleList_core.R ?

Logic behind the default iterations for calculating marginal WAIC being 1000?  It's fine with me, I have no problem with this, just wondering what that's based on?

Maybe this is too in the weeds.  But, if a model has dependence structure as:
theta -> latent1 -> latent2 -> data
and we only want to marginalize over latent1 for calculating marginal WAIC, then I think the code in MCMC_WAIC.R lines 197-200 won't handle this correctly.  In this case, the `marginalizeNodes` vector over which to integrate for calculating marginal WAIC will include latent2, not what the user wanted.  (affecting calculations on lines 264-265).

The user-facing options look good.  Complicated, but providing many options.  This is an extensive revision of WAIC (understatement), and it's a lot to digest.  But I have no additional comments about the options provided to users (which seem to be encompassing of anything I can think of), and the additions of the dummy WAIC function in  buildMCMC, and also the option for post-MCMC (offline) calculation of WAIC from the MCMC, or from a matrix of posterior samples, are good additions.

LGTM, pending comments above.

paciorek:
@danielturek thanks, I've fixed all of these issues. 

As far as 1000 it is ad hoc based on a small number of example models Josh was trying. But we do give the user the ability to change it and to realize it is too few.

Good point about two sets of latents. We could think about redefining 'data' nodes to be all downstream stochastic nodes including latent2 in your nomenclature (i.e., how well can we predict latent2 rather than the actual data, which wouldn't be affected by changing latent1). But I'm not sure whether WAIC as formally defined includes this case. So I think we should leave ti as it is and add a warning if we find additional latent nodes.

paciorek:
@danielturek @perrydv I'll merge this in unless and lingering thoughts.

paciorek:
@guhauhsoj We've finally merged in all the WAIC work -- thanks again for your work on this. You'll see that while the user interface is different than when you last added code, it's all built around the core numerical implementation that you provided.

PR #1137, opened 2021-06-25
fix issue 1128

paciorek:
This adds a check when building the CRP sampler that the number of stoch deps as well as deterministic deps is constant across the groups (i.e., the dependents of each element of the clustering variable).


paciorek:
Arggh, apparently I fixed this a second time and it is already in devel.

PR #1148, opened 2021-08-19
add error trap for incorrect [[ in nf run code

paciorek:
This addresses an untrapped error when incorrectly using `[[` in nf run code, per NCT issue 300.


paciorek:
The CI error on Windows doesn't seem to be substantive, so merging this in.

PR #1151, opened 2021-08-27
warn of dynamic indexes in a vector like `a[1:L]`

paciorek:
This partially addresses NCT issue 315 by giving a more interpretable error message than before.

It intercedes before the existing error: `R function `:` has arguments that cannot be evaluated; either the function must be a nimbleFunction or values for the following inputs must be specified as constants in the model` and instead says:

```
"Dynamic indexing found in a vector of indices, ", deparse(code), ". Only scalar indices, such as 'idx' in 'x[idx]', can be dynamic. One can instead use dynamic indexing in a vector of indices inside a nimbleFunction."
```


paciorek:
I've enhanced this to improve the error message and making sure no check is inserted in C++ if have constant values on either side of `:`.

Important note: this will affect all uses of `x:y` in nimbleFunction run code (e.g, `x[1:n,1:n]%*%y[1:n,1:n]`. So we need to check testing carefully and I'll want a review from @perrydv .

paciorek:
One open question here is that if we do have constants and there is negative indexing, I do insert a run-time check, which of course emits a message everytime the line of code is run. Probably fine and consistent with how we would handle a dynamic case that always has negative indexing (i.e., emit a message every time it occurs), but another option is to just report at compilation time and omit the run-time message.

Issue #1131, opened 2021-06-09
processing of too-long model declaration lines


danielturek:
Something is going off during compilation of model declaration lines that are "too long".  A warning is issued at compilation time, which looks indicative of a larger problem.

This may also have something to do with processing of dynamic indexing... not sure.

Reproducible example:

```
library(nimble)

code <- nimbleCode({
    log.mu <- indA[1]*beta[1]*covs[1,scale[1],1] + 
        indA[1]*indA[2]*beta[2]*(covs[1,scale[1],1]^2) +
        indA[3]*beta[3]*covs[1,scale[2],2] +
        indA[3]*indA[4]*beta[4]*(covs[1,scale[2],2]^2) +
        indA[5]*beta[5]*covs[1,scale[3],3] +
        indA[5]*indA[6]*beta[6]*(covs[1,scale[3],3]^2) +
        indA[7]*beta[7]*covs[1,scale[4],4] +
        indA[7]*indA[8]*beta[8]*(covs[1,scale[4],4]^2) +
        indA[9]*beta[9]*covs[1,scale[5],5] +
        indA[9]*indA[10]*beta[10]*pow(covs[1,scale[5],5],2) +
        indA[11]*beta[11]*covs[1,scale[6],6] +
        indA[11]*indA[12]*beta[12]*(covs[1,scale[6],6]^2) +
        indA[13]*beta[13]*covs[1,scale[7],7] +
        indA[13]*indA[14]*beta[14]*(covs[1,scale[7],7]^2)
})

data <- list(covs = array(0, c(10,3,7)))
inits <- list(scale = rep(1,7))
Rmodel <- nimbleModel(code = code, data = data, inits = inits)

Cmodel <- compileNimble(Rmodel)
## compilation finished.
## Warning message:
## In if (is.na(code)) stop("Character NA is not supported for compilation.") :
##   the condition has length > 1 and only the first element will be used
```

Upon inspection, the `code` element that's being processed here in the function `nimGenerateCpp` has value:
```
> code
[...]

PR #1154, opened 2021-09-02
robustify `nimble::deparse` to allow truncation to only first element of result

paciorek:
This fixes issue #1131 

This avoids getting a multi-element character vector when the deparse result is longer than 500 chars. I also use this functionality in all warning/error messages where `deparse()` is used in our codebase. 

More discussion of how we might robustify further in `deparse`  and `nimDeparse` are in issue #1131 .


paciorek:
Omitting in favor of PR #1167

PR #1158, opened 2021-09-15
fixed error with non-constant mu vector in dcar_proper sampler

### LINKED ISSUE #1157: dcar_proper posterior appearing biased with covariates in monte carlo study
Hello,

I've been working on proper CAR models in Stan, and part of my process has been to compare results to Nimble. Since the results differ across software, I've been using simulated data to check for biases in my models, and all the results for Nimble are turning out biased. Specifically regression coefficients within a proper car model. I'm very new to Nimble, so I know there's  a good chance that I'm just doing something wrong here. I'd be glad to hear if I've made a mistake.

I've created a reproducible example based on the Nimble ICAR example (<https://r-nimble.org/examples>). I simulate spatially auto-correlated log-rates (using `sim_sar`, the SAR model) with a covariate, then draw some count data using `rpois()`.  The data is simulated like this:
```
x <- sim_sar(w = W, rho = rho.x, sigma = 0.2)
phi <- sim_sar(w = W, mu = alpha + x * beta, rho = rho.y, sigma = 0.2)
y <- rpois(n = n, lambda = E * exp(phi))
```
Then I fit the CAR models and collect the posterior mean of the regression coefficient on each iteration, and also collect results from `lm(log(y/E) ~ x)`, which should also be unbiased. 

```
library(nimble)
library(CARBayesdata, quietly = TRUE)
library(sp, quietly = TRUE)
library(spdep, quietly = TRUE)
Nsim <- 30
#' draw from simultaneous autoregressive (SAR) model
sim_sar <- function (m = 1, mu = rep(0, nrow(w)), w, rho, sigma = 1, ...) {
    stopifnot(inherits(w, "matrix"))
    stopifnot(ncol(w) == nrow(w))
    N <- nrow(w)
    if (missing(mu)) {
        mu <- rep(0, N)
    }
    I <- diag(N)
    S <- sigma^2 * solve( (I - rho * t(w)) %*% (I - rho * w) )
    x <- MASS::mvrnorm(n = m, mu = mu, Sigma = S, ...)
    return(x)
}

data(GGHB.IG)
data(respiratorydata)
respiratorydata_spatial <- merge(x = GGHB.IG, y = respiratorydata, by = "IG", all.x = FALSE)

W.nb <- poly2nb(respiratorydata_spatial, row.names =  rownames(respiratorydata_spatial@data))
nbInfo <- nb2WB(W.nb)
CM <- as.carCM(nbInfo$adj, nbInfo$weights, nbInfo$num)
nregions <- nrow(respiratorydata_spatial)

code <- nimbleCode({
    # priors
    alpha ~ dnorm(0, 10) #/#
    beta ~ dnorm(0, sd = 100)
    sigma ~ dunif(0, 100)   # prior for variance components based on Gelman (2006)
    prec <- 1 / sigma^2
    rho ~ dunif(0, 1)
    # latent process
    s[1:N] ~ dcar_proper(mu[1:N],
                         C = C[1:nC],
                         adj = adj[1:L],
                         num = num[1:N],
                         M = M[1:N],
                         tau = prec, 
                         gamma = rho)    
    # likelihood
    for(i in 1:N) {
        mu[i] <- alpha + beta * x[i]
        lambda[i] <- E[i] * exp(s[i])
        y[i] ~ dpois(lambda[i])
    }
})

#/# prep sim data
W <- spdep::nb2mat(spdep::poly2nb(respiratorydata_spatial), style = "W")
rho.x <- 0.75
rho.y <- 0.6
beta <- -1
alpha <-  0
n <- nrow(W)
E <- respiratorydata_spatial$expected
inits <- list(alpha = 0, beta = -1, prec = 6, s = rnorm(n))
data <- list()
constants <- list(N = nregions,
                  L = length(nbInfo$adj),
                  C = CM$C,
                  M = CM$M,
                  nC = length(CM$C),
                  adj = nbInfo$adj,
                  num = nbInfo$num, #x = x,
                  E = E
                  )

res <- lapply(1:Nsim, function(i) {
    x <- sim_sar(w = W, rho = rho.x, sigma = 0.2)
    phi <- sim_sar(w = W, mu = alpha + x * beta, rho = rho.y, sigma = 0.2)
    y <- rpois(n = n, lambda = E * exp(phi))
    constants$x <- x 
    data$y <- y
    # par(mfrow=c(1,3)); hist(x); hist(y/E); plot(x, log(y/E))
    model <- nimbleModel(code, constants = constants, data = data, inits = inits)
    cModel <- compileNimble(model)
    conf <- configureMCMC(model, monitors = c('alpha', 'beta', 'sigma', 's'))
    MCMC <- buildMCMC(conf)
    cMCMC <- compileNimble(MCMC, project = cModel)
    samples <- runMCMC(cMCMC, niter = 15000, nburnin = 7500)
    f <- lm(log(y/E) ~ x)    
    c(nimble = mean(samples[,"beta"]), lm = coefficients(f)["x"])
}
)
r = do.call("rbind", res)
readr::write_csv(as.data.frame(r), "data/test-nimble-bias-mean-centered.csv")
#r <- readr::read_csv("data/test-nimble-bias-mean-centered.csv")

message("Sampling from log-linear Poisson model with beta = -1 (nsim = ",nrow(r), ")")
message("Sample mean of Nimble estimates: ", mean(r[,"nimble"]))
message("Sample mean of lm estimates: ", mean(r[,"lm.x"]))

par(mfrow = c(1,2))
hist(r[,"nimble"])
hist(r[,"lm.x"])
```
These are the results I've gotten:
```
> message("Sampling from log-linear Poisson model with beta = -1")
Sampling from log-linear Poisson model with beta = -1
> message("Sample mean of Nimble estimates: ", mean(r[,"nimble"]))
Sample mean of Nimble estimates: -0.801299816222076
> message("Sample mean of lm estimates: ", mean(r[,"lm.x"]))
Sample mean of lm estimates: -1.01871752123125
```

I also tried OpenBUGS, and found that results are different from Nimble but also appear biased. The amount of bias here is consistent with other results I've gotten. Obviously if I'm making a mistake here with Nimble, I could be making the same or similar mistake with OpenBUGS. However, when I do not use a covariate (intercept only model), the results from Nimble and Stan are identical. 

Am I doing something wrong here?  Thanks for your time!

```
> sessionInfo()
R version 4.1.1 (2021-08-10)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.3 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] spdep_1.1-8      sf_1.0-2         spData_0.3.10    sp_1.4-5        
[5] CARBayesdata_2.2 nimble_0.11.1   

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.7         compiler_4.1.1     pillar_1.6.2       LearnBayes_2.15.1 
 [5] class_7.3-19       shapefiles_0.7     tools_4.1.1        boot_1.3-28       
 [9] nlme_3.1-152       lifecycle_1.0.0    tibble_3.1.3       lattice_0.20-44   
[13] pkgconfig_2.0.3    rlang_0.4.11       Matrix_1.3-4       igraph_1.2.6      
[17] DBI_1.1.1          parallel_4.1.1     expm_0.999-6       e1071_1.7-8       
[21] coda_0.19-4        dplyr_1.0.7        raster_3.4-13      gtools_3.9.2      
[25] generics_0.1.0     vctrs_0.3.8        classInt_0.4-3     grid_4.1.1        
[29] tidyselect_1.1.1   glue_1.4.2         R6_2.5.0           fansi_0.5.0       
[33] foreign_0.8-81     gdata_2.18.0       deldir_0.2-10      purrr_0.3.4       
[37] magrittr_2.0.1     gmodels_2.18.1     splines_4.1.1      MASS_7.3-54       
[41] codetools_0.2-18   ellipsis_0.3.2     units_0.7-2        assertthat_0.2.1  
[45] KernSmooth_2.23-20 utf8_1.2.2         proxy_0.4-26       crayon_1.4.1     
```

The results:

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head>
	
	
	<title></title>
	<meta name="generator" content="LibreOffice 6.4.7.2 (Linux)"/>

	
</head>

<body>

nimble | lm.x
-- | --
-0.82232725857513 | -1.03677864295126
-0.876502927693985 | -0.974760506404128
-0.858686565413772 | -1.1000797449563
-0.786424043640384 | -0.929469727903681
-0.688338525708843 | -0.958121857059611
-0.646006763105514 | -0.764433142958986
-0.709161374581548 | -0.892886985978557
-0.835793984533608 | -1.14589608020523
-0.812737439486002 | -1.12577896910568
-0.86187317932852 | -1.0782901574334
-0.855261315967987 | -0.956724296180544
-0.889154185893735 | -1.02506647983045
-0.840048451024205 | -1.08381443465891
-0.80990753317543 | -1.00541709721234
-0.759248973226371 | -0.975112288165628
-0.858613957576768 | -1.14430364775781
-0.700019037284245 | -0.92014704006507
-0.859785451942923 | -1.12990489599549
-0.790357999019719 | -1.03179045854255
-0.745481917835726 | -1.02967259799018
-0.812743181991503 | -1.00116485232969
-0.796261269960158 | -0.972140732732058
-0.772647389490486 | -0.928329688788328
-0.859329771644722 | -1.01867771933016
-0.754810459171743 | -0.962827845998812
-0.831211743010605 | -1.13312305460461
-0.854018029248226 | -1.07643180240894
-0.816860284505026 | -1.05400433443646
-0.711645793700346 | -0.941255309575998
-0.823735678925053 | -1.16512124537676

</body>

</html>

danielturek:
Fixes a bug in MCMC sampling of `dcar_proper` distributions, when the mean vector `mu` is non-constant.

The original (buggy) version correctly implemented the equations in the GeoBUGS User Manual, Version 1.2 (September 2004), which contains a typo in the conditional specification of the proper CAR distribution.

Fixes #1157


danielturek:
Noting this PR might not initially pass testing, since testing may well include posterior results for proper CAR models.

paciorek:
Thanks, @danielturek . Would you like to fix the equation for the conditional density for the CAR in the manual or should I?

danielturek:
Done.  Nice eye @paciorek

paciorek:
I have an efficiency request here. Can you call `getParam` for `mu` once (instead of twice) and then index into it for the target and for the neighbors. I did a quick check and changing to that looks like it might save 10% of the MCMC time on Connor's example.

danielturek:
@paciorek I just made these changes.  Only one call to `getParam` now.  See what you think.

paciorek:
@danielturek did you forget to push to github?

danielturek:
@paciorek Sorry.  Done.  Pushed just now.

paciorek:
Three further thoughts (sorry to belabor things)

1. So the 'plusOne' business seems hard to read. Is there a reason not to assign the full vector of means to a variable and then index that as needed to get the `targetMu` and `neighborMus`? I don't think there will be extra cost since there presumably is at least temporary vector created by the output of `getParam`, regardless of whether we create one explicitly. Sorry to back-seat drive, but given I'm looking at it, that's my reaction.
2. Can we take this opportunity to remove the for loop in the setup code and use `%in%`:
```
neighborIndices[1, ] <- which(targetDCARscalarComponents %in% neighborNodes)
```
3) Also, do you know why we are using 1-row 2-d arrays instead of a vector for `neighborsC`, and `neighborIndices`?

danielturek:
No worries about pushing back on these design/implementation choices.

1. Logic is that the total number of means could be huge, but largest number of neighbors comparatively small.  I think this makes sense.
2. Could use `%in%`, but trying to be defensive against non-consecutive declarations of CAR process nodes.  Maybe that's not possible, but this is explicit, and it's in setup code, so this seemed safer to me.  Also noting, the code you drafted would not work, would be incorrect, due to LHS indexing - so more care and code would be necessary regardless.
3. Option is between creating vectors of at least +2 length extra (to avoid scalar vs. vector ambiguity, while also accounting for the num neighbors = 0 case) or otherwise to use arrays, as was done here.  If using vectors is known to be much better for performance reasons, then this design choice could be changed.

paciorek:
Ok, thanks for responses.

Regarding item 1, I don't see how your approach saves anything. In this code

```
targetNeighborMus <- model$getParam(targetDCAR, 'mu')[targetNeighborIndices[1,1:numNeighborsPlusOne]]
```
based on looking at the generated C++, an intermediate variable of length N is getting created based on the call to getParam, so doing the indexing within the RHS is no different than assigning the result of getParam to an explicit variable and then grabbing the subsets out of the full vector.

paciorek:
Side note: I was curious if it mattered that we call `getParam` on every individual target, so I did some monkeying around (I needed to create a version of `$run` that took an input parameter, which required creating `sampler_BASE2`) and if we instead call `getParam` before looping through the componentSamplerFunctions and pass the mean vector into the individual samplers it does save a bit of time on this example (9.2 seconds vs. 9.8). Not sure how that would scale as N changes. Not advocating we do this now, but something to consider more at some point.

danielturek:
@paciorek Thanks for explaining this.  I understand your comments, about each component sampler getting the entire mean vector, into a length-N vector, then subsetting as necessary.

I agree, this could be avoided by one call to getParam in the encompassing CAR sampler function, then passing the relevant neighbor mean components to each component (scalar) sampler function.  Given the small amount of testing you did, my inclination is to leave this as-is - which is working, simpler, and very similar in your speed test.

How do you feel about that?  Or, is there another suggestion that you made, which I missed?

paciorek:
I'm happy to skip using `getParam` once outside of the individual sampler functions, but I would still vote that we use `getParam` inside the individual sampler function run code to extract the entire mean vector to a local variable in the run code and then subset that to get the target and the neighbor means, avoiding the "plusOne" approach of concatenating the target and neighbor indices, which I think makes the code harder to read. As I mentioned in either your approach or my suggestion, there is a temporary N-long vector being created in the C++, so using getParam and then extracting the target+neighbor values in one line of code doesn't actually save us anything.

danielturek:
@paciorek Ok, that makes sense.  Are you able to make this change in the CAR component sampler code?

paciorek:
@danielturek Yes, I'll make that change.

paciorek:
Ok, we now set a variable that contains the entire vector of mu values in the run code and then index that.

I also realized that we could use `match` in place of the for loop in setup code and it should achieve the same robustness but much faster. I do think that with a for loop over target elements and then the for loop in setup code that we could have non-negligible time spent on that loop and the use of match should speed things up by at least an order of magnitude if not more.

@danielturek feel free to take a look.

danielturek:
Just pushed a minor change of variable name.

(The previous name `targetNeighborIndices` was meant to suggest "target index and neighbor indices together".  Changed this name back to `neighborIndices`)

paciorek:
Good catch. Thx.

Issue #1108, opened 2021-03-20
missing indexing in demotion of matrix to vector causes conjugacy processing error


paciorek:
This is a reprex of the issue raised on nimble-users on March 18, 2021.

Note the missing indexing in the definition of `a`. 

```
code <- nimbleCode({
    mu~dnorm(0,1)
    y~dnorm(mu +a[2],3)
    a[1:3] <- (myfun())[,1]       ## this fails
    ## a[1:3] <- (myfun())[1:3, 1]  ## this is fine
})

myfun = nimbleFunction(run=function() {
    returnType(double(2))
    out = nimMatrix(0, 3,3)
    return(out)
})

m = nimbleModel(code,data=list(y=0))
configureMCMC(m)
# Error in parse(text = x, keep.source = FALSE)[[1]] : 
#  subscript out of bounds
```

Traceback indicates the issue is somewhere in conjugacy processing, presumably because we are expecting full indexing when tracing through the deterministic expression(s).

paciorek:
In contrast, having:

```
    a[1:3] <- b[,1]  
```

is fine.

paciorek:
Also note that if we have `myfun()[,1]` instead of `(myfun())[,1]`, we do trap this at the model definition stage.

And in cases where a different flow through conjugacy checking occurs, this is also fine, e.g.,

```
code <- nimbleCode({
    mu~dnorm(0,1)
    a[1:3] <- (myfun())[,1]       ## this is fine
    y[1:3] ~ dmnorm(a[1:3], pr[1:3,1:3])
    yy~dnorm(y[2],1)
})

myfun = nimbleFunction(run=function() {
    returnType(double(2))
    out = nimMatrix(0, 3,3)
    return(out)
})

m = nimbleModel(code,data=list(yy=0),inits = list(b=matrix(rnorm(8),4,2)))
configureMCMC(m)
```

I'm just going to add a rudimentary check to `cc_expandDetermNodesInExpr` that if missing index is found, we just say that for this particular case, for conjugacy checking, that there may be missing indexing.

paciorek:
This is now fully addressed in the PRs mentioned above.

PR #1156, opened 2021-09-08
allow cases like `myfun()[,1]`

paciorek:
In PR #929 we added handling of ``(expression)[indexing]`. This adds handling of `myfun()[,1]` and `myfun(b[,1])[,1]`, as discussed in NCT issue 317 by mimicing cases like `(X[1:2,1:2] %*% beta[1:2,1:2])[,1]`.

I will add a few tests.

PR #1162, opened 2021-09-16
Improve logging and add check for unused constants

paciorek:
This PR:

1. adds a check for constants not used in the model but provided in `constants`
2. Overhauls how we emit standard logging messages
3. improves checking of missing indexes in `myfun()[,1]` and `(myfun())[,1]` to allow these cases. It also ensures that indices in arguments to `myfun` in  `myfun(k[,1])[,1]` are checked. See issue #1108 for some discussion of this.

I didn't intend for the commits for item 3 above to be in this branch but somehow they ended up here.

It's not clear why this can't be automatically merged (I can merge via the commandline with no conflicts), so I'm putting in the PR to get tests to run and will then merge manually.


paciorek:
With bullet item 3 it turns out I did end up with a weird merge conflict with stuff being in devel that I thought was there but then didn't seem to be there and then was there. I think things are ok now.

paciorek:
Yeah item 3 was in commit 9875c38, which was merged into devel last night via PR #1156

paciorek:
I pulled the constants check out for now because it was flagging variables used in definition time if-then-else; see NCT issue 319.

Issue #1134, opened 2021-06-21
emit warning or error if using negative indexing in a nimbleFunction


paciorek:
Per user email: 

```
test = nimbleFunction(
  run = function(x=double(1)) {
    n = length(x)
    nimPrint(sum(x[1:n]))
    nimPrint(sum(x[n:1]))
  }
)

cTest = compileNimble(test)

test   (1:6)    # Prints 21 twice  - the expected output
cTest(1:6)    # Prints 21 and 6  - somewhat surprising output 
```

paciorek:
I'm working on adding an assertion into the C++.

One question is whether we just want to report and error but not actually error out. I guess so given that's what we do with mis-matched dimensions in matrix multiplication, but not fully seeing why we don't actually error out.

paciorek:
See PR #1151 for full proposal for how to handle this for any use of `x:y` in DSL code.

paciorek:
I'm actually now throwing an error in `sizeColonOperator` if there is negative indexing with constants, since there is no point in compiling it and then throwing messages every time the line of code is executed.

paciorek:
Also note that there is no check in C++ if have constant values on either side of `:`.

paciorek:
In PR #1166 we now catch negative indexing with constant indexes. There is not an obvious way to insert run-time error trapping only in cases of user code that needs it (and avoiding inserting a trap in cases like `1:niter`) so negative indexing could still be a problem in some cases.

paciorek:
Error trapping of dynamic negative indexes is now discussed in NCT issue 315 so closing this.

PR #1166, opened 2021-09-18
error trap constant negative indexes and add run-time check for dynamic negative indexes

paciorek:
In `sizeColonOperator`, this inserts an error trap for negative constant indexes and, if indexes are not constant adds a check in run-time code for negative indexes. 

This fixes issue #1134


paciorek:
Ok, I didn't think this through. Since the size processing handles node functions as well, this is injecting run-time checks in a lot of our core generated code. And even worse, there are times that we generate code where we expect negative indexing, such as `1:0` is usually the case for 

```
for (i in seq_along(initFunctionList))
```
in `my_initializeModel$run`. Or in our overall MCMC, we have `1:niter` and don't want to check that probably.

It's not clear how to know within `sizeColonOperator` when we want the check and when we don't...

I'm going to drop the dynamic check, but leave the setup code check for constant negative indexes.

Issue #1171, opened 2021-10-01
dynamic conjugacy processing fails if first dependent equals 1 exactly for a multiplicative or 0 exactly for an additive case if values provided in constants


paciorek:
Per nimble-users example, this model results in no calculation of the 'coeff' in the conjugate sampler run code because the coeff for the first dependent is baked in as 1 and cc_checkLinearity sees that and thinks there is no coefficient needed.

```
code <- nimbleCode( {
  beta ~ dnorm(0,0.001)
  for (i in 1:n) 
  y[i] ~ dnorm(alpha+beta*x[i], tau) 
})
m <- nimbleModel(code, constants = list(n = 4, x = 1:4), data = list(y = rnorm(4)))
```

Similar issue would arise with the offset if we had `alpha[i]+beta*x[i]` and `alpha[1]=0` is defined in constants.

Need to look in `cc_checkLinearity` to see if we can simply indicate that there is a coeff needed whenever we have multiplication by anything, regardless of whether that is 1 exactly. And similar for offset and 0.

Perry also suggested the possibility of turning off baking in constants, but I'd like to avoid that big a change for what should just be an oversight in the conjugacy checking.

paciorek:
Ok, here is a proposed solution. Basically I think what we need is that we only avoid an offset if there are no additions that add anything to expressions involving the target. And we only avoid a coeff if there are no multiplications that multiply by an expression involving the target. These cases need to be distinguished from adding 0 or multiplying by 1, since that can occur when constants are used. 

I think we can do this by extending the `cc_linearityCheck` framework to include the notion of `NULL` that could be used for offset or scale. NULL means, respectively, no addition and no multiplication. Whenever we find just the target when calling `cc_linearityCheck` we return NULL for both offset and scale. Then the key thing is that when NULL values get combined (e.g., `cc_combineExprsAddition` with other things to build up the `offset` and `scale`, we treat the NULL values as zero or one respectively. Thus NULL never gets passed along unless we have offsets of two NULLs as input (I think - need to check this). Then in `cc_linkCheck` the only time we omit the offset is if the final offset is `NULL` and similarly for the coeff if final scale is `NULL`.

The result should be the only time we omit the offset is in cases like `beta` or `x[i]*beta` and only time we omit the coeff is in cases like `beta` or `alpha + beta`.

Let me know of any thoughts, in particular @danielturek, given that most of the `cc_check` code was originally yours.

[...]

PR #1172, opened 2021-10-06
fix 0 or 1 as constants in conjugacy processing

paciorek:
This fixes issue #1171 . 

@danielturek @perrydv here's the fix for the 0/1 constants conjugacy issue. Let me know of any concerns.

### DIFF
--- a/packages/nimble/R/MCMC_conjugacy.R
+++ b/packages/nimble/R/MCMC_conjugacy.R
@@ -1346,12 +1346,23 @@ cc_vectorizedComponentCheck <- function(targetNode, expr) {
 ##############################################################################################
 ##############################################################################################
 
+cc_replace01 <- function(expr) {
+    ## replace {0,1} with {0,1}+1i so distinguished from offset,scale of 0,1 when match target
+    if(is.numeric(expr) && (expr %in% c(0, 1))) {
+        return(expr + 1i)
+    if(is.call(expr))
+        for(i in 2:length(expr))
+            expr[[i]] <- cc_replace01(expr[[i]])
+    return(expr)
+}
+
 cc_checkLinearity <- function(expr, targetNode) {
 
     ## targetNode doesn't appear in expr
     if(!cc_nodeInExpr(targetNode, expr)) {
         if(is.call(expr) && expr[[1]] == '(') return(cc_checkLinearity(expr[[2]], targetNode))
-        return(list(offset = expr, scale = 0))
+        # add +1i to tags 0s and 1s as not being from exact match to target
+        return(list(offset = cc_replace01(expr), scale = 0))  
     }
 
     ## expr is exactly the targetNode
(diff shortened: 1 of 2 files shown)


paciorek:
@danielturek I'm getting a curious test failure with this PR in the 'various conjugacies' test in `test-mcmc.R`, around line 427. The test failure seems to be because with the 'new' posterior predictive branch sampler, we assign that sampler instead of the samplers we used to assign.

However:

1. Testing on devel, 0.12.0, and 0.11.1 did not fail in GitHub Actions (but do fail now that I run the 'various conjugacies' test directly). So if that addition of posterior_predictive_branch affected the 'various conjugacies' test, it should have shown up a long time ago.
2. In PR #1086 you say explicitly that you needed to adjust some tests to add 'data' in order to have the same samplers assigned during testing that had been assigned before the posterior_predictive_branch sampler existed. This makes complete sense and would be consistent with 0.11.0, 0.11.1, 0.12.0, and devel all passing the GH Actions testing over the last 8 months or so. Yet I can't see any evidence of your addition of 'data' in the test files, either now or in commit 8e84f42b5, associated with PR #1086.

I'm very confused. Any insights you might have?  Also possible I'm missing something obvious as I am wrapping up at the end of a long day.

danielturek:
@paciorek I'm not sure about the current test failure, and I'll try to look at what's going on today.  But regarding the changes of "data" in the test files, associated with the MCMC posterior predictive branch sampler (hereafter: "branch sampler"), introduced in PR #1086 , I'll draw your attention to 2 commits which were in that merge into devel:

[commit 3a883246bdff6acef7f1d0595bb3750f397a0f3b](https://github.com/nimble-dev/nimble/pull/1086/commits/3a883246bdff6acef7f1d0595bb3750f397a0f3b)

which changed some of the data, so as to get the original conjugate samplers, but I believe that approach fixed **some** of the issues in testing from addition of the branch sampler, but not all (I'd have to go back to that commit, and re-run though testing to see what failures persisted, and after that approach, I subsequently modified the testing file test-mcmc.R to not use the branch sampler, see 

[commit 11bc7e82babb252e65f2774c47741de88d947f94](https://github.com/nimble-dev/nimble/pull/1086/commits/11bc7e82babb252e65f2774c47741de88d947f94)

which changes nimble option to not use the branch sampler in testing:
```
nimblePPBranchSamplerSetting <- getNimbleOption('MCMCusePosteriorPredictiveBranchSampler')
nimbleOptions(MCMCusePosteriorPredictiveBranchSampler = FALSE)
```

So, I'd have to keep looking through old commits and tests, to see what these failures were, but that's a small start.

It's still a mystery what's going failing with this current PR for fixing the conjugacy 0/1 issue.  To be looked into further.

paciorek:
Ok, so I looked a bit more and that option is critical. I think the main (only?) mystery now is why that option doesn't seem to be having an effect when running test-mcmc on this PR's branch. If I run the line 427 test separately on this branch after setting the option, it passes, as we expect.

paciorek:
@danielturek I think I figured it out. The new conjugacy processing is (correctly) changing things from a conjugate sampler to a RW sampler because of some constants that we have in this weird model.  Thanks for your help pointing out the nimbleOption setting.

PR #1177, opened 2021-10-23
add nimbleSMC testing to GH actions CI

paciorek:
This replaces PR #1099. We want to test nimbleSMC as part of nimble CI.


paciorek:
I just manually ran `test_package('nimbleSMC','^filtering$',reporter='summary')` (which is how nimbleSMC testing would be run from nimble's `run_tests.R` with devel branches of nimble and nimbleSMC and test passed. So hopefully goldfile issue is ok.

paciorek:
I'm merging in given that I have some hope the goldfile issue is ok.

PR #1178, opened 2021-10-24
Fix ordering of findSamplersOnNodes method of MCMC configuration object

danielturek:
Do not merge.  Opening PR to trigger testing.

Fixes a bug in ordering from `conf$setSamplers`, and other MCMC methods which use the internal MCMC configuration method `findSamplersOnNodes`.  This specifically occurs when the samplers are specified as a vector of node names.

With this fix, the order specified by the vector of node names is now preserved.

### DIFF
--- a/packages/nimble/tests/testthat/test-bnp.R
+++ b/packages/nimble/tests/testthat/test-bnp.R
@@ -4048,7 +4048,7 @@ test_that("Testing handling (including error detection) with non-standard CRP mo
   paramSamplers <- conf$getSamplers('thetaTilde')
   expect_identical(sapply(paramSamplers, function(x) x$name), rep('CRP_cluster_wrapper', n))
   ids <- sapply(paramSamplers, function(x) x$control$clusterID)
-  expect_identical(as.integer(ids), as.integer(c(2:5, 1)))
+  expect_identical(as.integer(ids), as.integer(c(2:n, 1)))
 
   ## Model F: non conjugate, non-identical data
   ## doesn't detect presence of some conjugacies because only check first set
@@ -4978,7 +4978,7 @@ test_that("Testing handling (including error detection) with non-standard CRP mo
   expect_identical(sapply(paramSamplers, function(x) x$name), rep('CRP_cluster_wrapper', n*2*2))
   ids <- sapply(paramSamplers, function(x) x$control$clusterID)
   expect_identical(as.integer(ids), rep(rep(1:n, each = 2), 2))
-  
+
 
   ## no dependence on sigma in mu prior
   code <- nimbleCode({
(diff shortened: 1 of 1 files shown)


paciorek:
Many moons pass...

Yes, agreed that we can simply fix the test to conform to the new, better ordering. I've made a commit to that effect.

paciorek:
@danielturek I'm getting a big tangled up here. Please look at the test in `test-bnp.R` in lines 3427-3463.

The sampler order goes through the `thetaTilde` nodes "by row". I.e., `thetaTilde[1,2]` is the "second" node.
So then we get that `conf$getSamplers()` has this ordering:  `xi`, `thetaTilde[1,1]`, `thetaTilde[1,2]`, etc. (i.e., thetaTildes are by row).
But if we do `conf$getSamplers('thetaTilde')`, we have this ordering `thetaTilde[1,1]`, `thetaTilde[2,1]`, etc. (i.e., thetaTildes are by column.)

At root of this is that I think there is an inconsistency between column ordering from `expandNodeNames` and `getNodeNames`. The latter is used when setting up the samplers.

Not sure what we want here, but going by row in one call and by column in the other doesn't seem right. And given we generally do things by column (I think), it's weird that the sampler order is to go by row within the thetaTilde nodes. But we'd probably not want to change that. 

Thoughts?

Once we untangle this, I can fix the testing errors in `test-bnp.R`, all of which I suspect relate to this issue of row vs. column ordering.

danielturek:
@paciorek Ugh, nice observation.  Yes, first things first, I agree we should clean up `getNodeNames` and `expandNodeNames`.  (I'm surprised we've never noticed this before).  Then, yes, once we decide on the "proper" ordering, we can make everything else operate similarly.

So.... what's the right ordering?  `x[1,1,1]` followed by `x[1,1,2]`, or by `x[2,1,1]`?  In terms of memory access, would one be preferable?

Intuitively, I feel like `x[1,1,2]` is more natural to come second.

Suggesting @perrydv also weigh in here.

danielturek:
@paciorek This could change ordering of nodes in the sampler specs, and hence potentially the ordering of MCMC samplers, and hence expected MCMC testing output.  I'm a little afraid at what the testing might "reveal"

paciorek:
Looks like it's just a lingering bnp test. I'm happy to resolve, or you can. 

I am surprised that reordering samplers doesn't change results for any other MCMC tests. I guess your change doesn't actually reorder the samplers in any of our tests?

paciorek:
Ok, there is a bit more to this rat's nest.

In the BNP example in lines 3392-3424, if one runs:

```
conf$getSamplers()
conf$getSamplers('thetaTilde')
```

the ordering of the thetaTilde samplers that is returned changes depending on the invocation. That is going to seem weird to a user. I may be about to contradict something I've said earlier, but wouldn't it make sense if `getSamplers` (and `printSamplers`) always gave back the result in the order that is being used in the MCMC sampling?

danielturek:
@paciorek Nice eye.  I agree.  I just tried to make that change, and we'll see what happens with testing, now...

perrydv:
For the record, we discussed the issue that getNodeNames and expandNodeNames can potentially return nodes in different orders.  This occurs because getNodeNames always returns nodes in topologically sorted order, while expandNodeNames has a sort argument that defaults to FALSE.  If sort=TRUE, expandNodeNames should give the same ordering as getNodeNames, which is whatever the canonical sorted order for the model was set up as.  That order can depend on the order of nesting indices, among other things.

danielturek:
@paciorek @perrydv I think this is sorted out now - hopefully(?).  I think I'm happy with the state of the MCMC configuration methods, and it's now passing testing.

I welcome any further review from @paciorek

PR #1181, opened 2021-11-21
ADoak_without_HMC

danielturek:
@perrydv @paciorek

This new ADoak_without_HMC:

- Removes the HMC and langevin samplers from branch ADoak.
- Removes all testing of the HMC sampler from ADoak.
- Changes the buildMCMC nimbleFunction, so instead of "special treatment" of HMC samplers, there's a new (enhanced) sampler type ("sampler_BASE2"), which in addition to the usual reset() method, also has before_chain() and after_chain() methods.  (I had wanted to call these new methods as initialize() and finalize(), but that caused a conflict with the internal reference class initialize() method, which I could not seem to work around).
- Adds a new field to the samplerConf reference class (the reference class which represents configuration information for a single sampler) called "baseClassName", which contains the base class name of the sampler - always either sampler_BASE or sampler_BASE2.  This information is used in buildMCMC.


danielturek:
@perrydv @paciorek , I wanted to update you both about progress with HMC.

There's a new branch (branched off of ADoak) called ADoak_without_HMC.  I'm going to create a PR, for merging this into ADoak.  But we should discuss these changes first.

This new ADoak_without_HMC:

- Removes the HMC and langevin samplers from branch ADoak.
- Removes all testing of the HMC sampler from ADoak.
- Changes the buildMCMC nimbleFunction, so instead of "special treatment" of HMC samplers, there's a new (enhanced) sampler type ("sampler_BASE2"), which in addition to the usual reset() method, also has before_chain() and after_chain() methods.  (I had wanted to call these new methods as initialize() and finalize(), but that caused a conflict with the internal reference class initialize() method, which I could not seem to work around).
- Adds a new field to the samplerConf reference class (the reference class which represents configuration information for a single sampler) called "baseClassName", which contains the base class name of the sampler - always either sampler_BASE or sampler_BASE2.  This information is used in buildMCMC.

Furthermore, there's a new package, nimble-dev/nimbleHMC, which:

- Works in conjunction with the above branch, ADoak_without_HMC
- Contains the HMC and langevin samplers
- Contains the testing.
- Everything was passing all testing, until I tried getting Github Actions to install nimble from the branch "ADoak_without_HMC", and I can't get that to work.  But otherwise, the packages (really, the ADoak_without_HMC branch of nimble, and the nimbleHMC package) work properly together.

This is for discussion, and I welcome code review on everything.  But from my perspective, it seems to work as desired.

paciorek:
(many moons pass....)

All this sounds reasonable to me. A few reactions for now:

1) Can we come up with anything more evocative than BASE2, such as BASE_wrapped? I guess that's not any more self-explanatory.

1a) Not advocating for this, just a thought experiment, but perhaps we should just add before_chain and after_chain to all our samplers and make sampler_BASE equal to your sampler_BASE2. But I guess that would affect users who have written user-defined samplers, so maybe that strikes that idea.

2) How about "init" and "finalize" or "prolog" and "epilog" or simply "before" and "after" instead of "before_chain" and "after_chain"

3) Regarding printing warnings about tree depth and divergences, I'm not sure we should equate that with nimble's general verbosity. Stan always reports this as a warning message, I believe. So I think we should always report it (or perhaps have a control argument that turns reporting off but defaults to on).

3a) I think the warnings need to say what the target params are because there could be multiple HMC samplers as part of a single MCMC.
3b) Also, could your print formatting match my new [WARNING], [NOTE] syntax, please?
3c) Finally 'numDivergences' and 'numTimesMaxTreeDepth' will be reasonably easily available to users as cMCMC$samplerFunctions[[i]]$numDivergences, right? I think it's important that the user can easily extract them in addition to being printed out as messages.

danielturek:
@paciorek @perrydv 

Thanks for the feedback.  Some responses below:

1. The original sampler base name is `sampler_BASE`.  I think of this as adding a "next generation" of MCMC sampler type, which I wish I'd started out with, but so it goes.  These "version 2" samplers include a method to execute before each MCMC chain (regardless of the `reset` argument), and at the conclusion of each MCMC chain.  The best name I came up with for this "second generation" sampler base name is `sampler_BASE2`.  Respectfully, I don't understand the implication of `sampler_ BASE_wrapped`.  I'm open to other suggestions for this new base name, but I have nothing better, myself.

1a. I'm quite against adding `before_chain` and `after_chain` methods to all existing samplers.  I have no issue with doing it for all of NIMBLE's samplers in the codebase, but the huge problem is breaking all other existing sampler code that's out there.  It would just upset too many people.  So I'm really not ok with modifying the base MCMC algorithm to **require** every sampler to have these additional methods.  That was a precondition going into the implementation of this.

2. For the methods names: As noted in my first comment, I originally wanted `initialize` and `finalize`, but I cannot get that to work in combination with the standard `initialize` method of all reference classes.  I'm open for other suggestions, but I thought `before_chain` and `after_chain` are the most clear.  Even `before` and `after` don't seem to convey the same meaning to me (before and after what?)  That said, again, this is the best I've come up with, and I'm open to discussion.  `init` and `finalize` are viable, although I'll suggest that they *still* aren't quite as clear as `before_chain` and `after_chain`.  I welcome other thoughts here.

3.  Good points @paciorek.  I'm happy to make some changes about the errors / warning / informative messages.  In particular:
- Make it a property of the HMC sampler (control argument), as you suggested, maybe with default value as the nimbleOption `verbose` ?  What do you think about that?
- I can add the target parameters to the errors.  Or, a truncation of them.  Good idea.
- Can follow the [WARNING] and [NOTE] syntax.  Good idea.

danielturek:
@paciorek Noting that printing messages that include target node **names** I don't believe would get through compilation.

As an alternative, I could say "HMC sampler #x", where x is the number of the HMC sampler in the MCMC configuration.  But otherwise, I don't think I could have a character string of node names pass through compilation.

Am I overlooking something?

danielturek:
@paciorek One final comment on this - about HMC sampler messages also reporting the target node names, or the sampler number.  This would require modifications of the core `buildMCMC`, to pass the sampler # information along to each `buildSampler` method, so it's a somewhat deeply rooted change.  I would be ok with this, but I feel this maybe should take place after more discussion, and perhaps *after* AD, HMC, etc, is actually part of NIMBLE - before we start modifying the core MCMC even more.  I welcome other thoughts or opinions, however.

perrydv:
Hi @danielturek @paciorek .  

- Note that it is possible to print node names from within compiled code.  The following works:
```
test <- nimbleFunction(setup = function(nodes) {},
                   run = function(i = integer()) {print("hey ", nodes[i])})
t1 <- test(letters[1:4])
ct1 <- compileNimble(t1)
ct1$run(1)
ct1$run(2)
```

- Note that `cMCMC$samplerFunctions[[i]]$numDivergences` will not work unless a user turns on a nimbleOption.  By default, interfaces are built only one level in.  So `cMCMC$samplerFunctions[[i]]` will do something, but the next `$` will not.  Turning on the option to build interfaces for at all levels of nimbleFunction nesting weighs down performance substantially.  

- I was also thinking about if there is a good way to keep everything as `sampler_BASE` and make it just work.  I agree we shouldn't do that if it breaks user code.  I think we could do it with a small change in compiler behavior possibly coupled with a small extension to `nimbleFunctionVirtual`.  Say we add `before_chains` and `after_chains` to `sampler_BASE`.  Existing samplers would fail to compile because those methods are inserted in C++ as abstract virtual functions, meaning they are required to be defined by derived classes (samplers).  If we arrange for them to be defined as non-abstract virtual functions, then derived classes will inherit default (do-nothing) behavior as a default, and compilation should "just work" with backward compatibility.  So that could be an option.

- I also thought about the name `sampler_BASE2`.  I'm not thinking of a non-clunky way to do better.

- I also thought about the method names.  How about `before_sampling` and `after_sampling` or `setup` and `cleanup`?

danielturek:
@perrydv @paciorek Thanks to both of you for the feedback.

@perrydv thanks for pointing out about character strings working through compilation.  This is great, I'm sorry I totally missed that enhancement.  I've updated all the HMC messages, to include the nodes names (truncated to 100 characters, maximum).  An example of the  output is:
```
running chain 1...
  [Note] HMC sampler (nodes: b0, b1, sigma) is using 1000 warmup iterations
|-------------|-------------|-------------|-------------|
|-------------------------------------------------------|
  [Note] HMC sampler (nodes: b0, b1, sigma) encountered 6106 divergent paths
  [Note] HMC sampler (nodes: b0, b1, sigma) reached the maximum search tree depth 53 times
```

@perrydv If this is a change we could make to the system of "virtual nimble function inheritance", whereby **default** values of member function could be provided, that would be fantastic.  If it were possible, "virtual" nimble functions could include specifications for **mandatory** methods, where no default is supplied, and also for **optional** methods, where the virtual definition would also provide a default implementation of the method.  I would prefer this approach *much more* than the current addition of `sampler_BASE2`.

Maybe (using `sampler_BASE` as an example) something like:
```
sampler_BASE <- nimbleFunctionVirtual(
    name = 'sampler_BASE',
    methods = list(
        reset = function() { }
    ),
    methods_with_default = list(
        before_chain = function() {},     ## these defaults happen to do nothing
        after_chain = function() {}       ## these defaults happen to do nothing
    )
)
```
@perrydv If this is an enhancement you'd be willing to make to the virtual nimble function system, then I would happily modify the MCMC and HMC code to work with it.

As for names of these methods.  I'm fine with any of {`before_chain`, `after_chain`}, or {`setup`, `cleanup`}, or {`before_sampling`, `after_sampling`}, although I do feel the first option is the most clear.

danielturek:
@perrydv I updated this PR to remove the `sampler_BASE2`, and instead to use the new:

`methodControl = list( METHODNAME = list(abstract = FALSE) )`

way of specifying `nimbleFunctionVirtual` methods with defaults.

Noting also that this branch (`ADoak_without_HMC`) doesn't build at present, until we merge in the update to `nimbleFunctionVirtual` which introduces the `methodControl` list for virtual nimble functions, and non-abstract nimbleFunction methods.

danielturek:
@perrydv Could I merge this into ADoak?  I'm keen to keep the HMC moving forward.

perrydv:
OK @danielturek LGTM, let's move ahead.

danielturek:
@perrydv Thank you.  Merging.

PR #1193, opened 2022-01-21
RJMCMC errors out for non-constant hyperparameters

danielturek:
`configureRJ` uses `model$getParents()` to find non-data stochastic parents of RJ target nodes.  In the event of finding any non-constant hyperparameters, `configureRJ` bails out, with the message:

"Reversible jump target node [node name] appears to have a non-constant hyper-parameter, which is not currently supported for reversible jump MCMC.  Please contact the nimble development team at nimble.stats@gmail.com if support for this case would be helpful."

Also added tests for these cases to `test-mcmcjr.r`.

### DIFF
--- a/packages/nimble/R/MCMC_RJ.R
+++ b/packages/nimble/R/MCMC_RJ.R
@@ -299,6 +299,8 @@ configureRJ <- function(conf, targetNodes, indicatorNodes = NULL, priorProb = NU
             nodeControl <- list(priorProb = priorProb[i], mean = mean[i],
                                 scale = scale[i], fixedValue = fixedValue[i])
             for(j in 1:length(nodeAsScalar)) {
+                ## RJ system does not support non-constant hyperparameters of target RJ nodes:
+                if(length(conf$model$getParents(nodeAsScalar[j], stochOnly = TRUE, includeData = FALSE)) > 0) stop('Reversible jump target node \"', nodeAsScalar[j], '\" appears to have a non-constant hyper-parameter, which is not currently supported for reversible jump MCMC.  Please contact the nimble development team at <nimble.stats@gmail.com> if support for this case would be helpful.')
                 currentConf <- conf$getSamplers(nodeAsScalar[j])
                 ## check on node configuration
                 if(length(currentConf) == 0) {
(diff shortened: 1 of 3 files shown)


perrydv:
Hi @danielturek Thanks for drafting this. 

- Can we do all checking before we do any modifying of the configuration?  That way, if the check fails for the non-first node, we can bail out with the configuration completely unmodified.
- Can we do the call to getParents for either all of `targetNodes` or (within the first for loop) all of `targetNodes[i]` instead of (within the second for loop) each `nodeAsScalar[j]`? You'd have to confirm that it works, but if so fewer calls would be more efficient.
- Can we put the checking itself behind a nimbleOption?  That way, if say `getParents` fails for some unrelated reason but the model code is actually ok, we can give a user a way to proceed (by turning off the nimbleOption in order to disable the check).
- I would lean towards omitting the statement about emailing the development team if the case would be helpful.  That seems true for many things.

danielturek:
@perrydv Great comments.  I incorporated them all.

Only thing to mention - in the now **single** call to `getParents`, if any problematic targetNodes are found, then it enters into a loop, which again uses `getParents` for each individual scalar element of targetNodes, to find the first offending node (for the error message).  This internal loop would be "slow" again, but in the case of "no problems", then only 1 call to `getParents` would take place.  I think this is a nice compromise between efficiency, and helpful error messages.  Let me know if it sounds reasonable.

paciorek:
LGTM. I will merge in soon.

perrydv:
Thanks @danielturek These look good to me.  We could potentially emit a message before entering the slower loop, announcing that a problem has been found and is not being investigated.  But perhaps that is overkill.

paciorek:
That sounds like overkill to me, FWIW.

paciorek:
A few minor things:

1) In testing,  the assignment of RJ to `beta1` should not throw an error. I'm not sure what the `NA` business is there, but shouldn't we just not have an `expect_error` in those cases?
2) Can we simplify from `conf$model$getParents` to `model$getParents`?
3) Not clear to me why we expand node names before passing to `getParents` (since `getParents` should do its own expansion). Is this just because if we find an issue, we'll need to expand anyway?
4) I think `if(length(x))` is cleaner than `if(length(x) > 0)`, but fine with leaving as is.

danielturek:
@paciorek 

1. Agreed about not throwing an error for RJ assignment to `beta1`.  Take a look at the documentation for `expect_error`, in particular the 2nd argument which is called `regexp` ("A character vector giving a regular expression that must match the error message.  If NULLL, the default, asserts that there should be a error, but doesn't test for a specific value.  If NA, asserts that there should be no errors.")
2. Agreed, that `conf$model` can change simply to `model`.  I just made that change.
3. Fair point about exactly where the `expandNodeNames` is done.  I moved it, so it only takes place if there's a problem.

Thanks for the careful code review.

paciorek:
The `NA` argument seems weird to me - if we have code in a `test_that`, then if there is any error not found within `expect_error` we'll get an error, so my approach is just to not use `expect_error`, often just including the code without any `expect_xyz`. It seems to me harder to read / obscure to use `expect_error` with an NA.

danielturek:
@paciorek Noting this `expect_error(..., NA)` approach is already used in `test-mcmcrj.R` lines 149, 152, 155, 475 and 478, as well as in other MCMC tests, at least.

I'm not sure what's best, anyway...

paciorek:
Huh, it looks like I did that. Weird. Well, I guess makes sense to leave that approach for now.

I'll merge this in.

Issue #1176, opened 2021-10-20
nimSeq is broken


perrydv:
`nimSeq(from = 1, to = 10, by = 1)` fails to compile.

paciorek:
This is fine now. Not sure if it was affected by #1199 .

PR #1199, opened 2022-01-29
fix handling of default by = 1 when not given in nimSeq

perrydv:
This PR fixes a bug in default handling of a basic nimSeq case for compilation.

```
# In a nimbleFunction:
1:10 # works (and likely a reason this bug has not been revealed before)
nimSeq(1, 10, by = 1) # works
nimSeq(1, 10) # broke when inserting what should be default of 1.  fixed by this PR
```


paciorek:
Check why issue #1176 report is contradictory of the notes above.

paciorek:
All three cases look fine now. Not sure why #1176 reported that `nimSeq(1, 10, by = 1)` didn't work.

Issue #1194, opened 2022-01-26
3D Array does not work for conjugate sampler.


ethan-alt:
I wanted to conduct a mixture of multivariate normals problem. The code I have is as follows:
```{r}
library(nimble)
## NIMBLE CODE--PRODUCES AN ERROR
code.multi <- nimbleCode({
  gamma ~ dbeta(gamma_shape1, gamma_shape2)
  for ( i in 1:2 ) {
    mu[1:J,i] ~ dmnorm(mu_mean[1:J], cov = mu_cov[1:J,1:J])
    Sigma[1:J,1:J,i] ~ dinvwish(S = S[1:J, 1:J], df = df)
  }
  for ( i in 1:n ) {
    c[i] ~ dbinom(size = 1, prob = gamma)
    d[i] <- c[i] + 1
    y[1:J,i] ~ dmnorm(mu[1:J, d[i]], cov = Sigma[1:J,1:J,d[i]] )
  }
})

## TEST THE CODE
n = 100
constants <- list(
    'n' = n
  , 'J' = 2
  , 'gamma_shape1' = 2
  , 'gamma_shape2' = 2
  , 'mu_mean'      = rep(0, 2)
  , 'mu_cov'       = diag(100, 2)
  , 'S'            = (2 + 10) * diag(2)
  , 'df'           = 2 + 10
)
inits <- list(
    'mu'    = cbind(c(-1, -1), c(1, 1))
  , 'c'     = rbinom(n = constants$n, size = 1, prob = 0.5)
  , 'gamma' = 0.5
  , 'Sigma' = abind( diag(2, 2), diag(1, 2), along = 3 )
)
data <- list('y' = t(mvtnorm::rmvnorm(constants$n, sigma = diag(2))) )
sample <- nimbleMCMC(
  code = code.multi, constants = constants, data = data, inits = inits, thin = 1, niter = 2000, nburnin = 1000
)
```

The sampler threw the following error:
```{r}
Defining model
Building model
Setting data and initial values
Running calculate on model
  [Note] Any error reports that follow may simply reflect missing values in model variables.
Checking model sizes and dimensions
Checking model calculations
Compiling
  [Note] This may take a minute.
  [Note] Use 'showCompilerOutput = TRUE' to see C++ compilation details.
Error: Error, wrong number of indices provided for dep_dmnorm_identity_coeff[iDep].
 This occurred for: dep_dmnorm_identity_coeff[iDep]
[...]

Issue #1175, opened 2021-10-15
incorrect size handling with conjugate Wishart/invWishart and dynamic indexing


paciorek:
A user reported a compilation error in a Wishart conjugate sampler that occurs with our new efficient conjugacy calculations.

The problem is that in lines 582-583 of `MCMC_conjugacy.R`, when we have dynamic indexing the `currentLInk` ends up being `identity` (poorly named -- `identity` ends up being the default name when there is dynamic indexing regardless of the actual link), and so we incorrectly omit setting `targetCoeffNdim <- 0`. 

I need to work through the logic of how best to set this in this case. It might be:

```
if(currentLink == 'multiplicativeScalar' || (link == 'multiplicativeScalar' && nimbleOptions()$allowDynamicIndexing && doDependentScreen)
                        targetCoeffNdim <- 0
```

Probably also worth thinking more about the fact that we use `identity` for the link name with we have dynamic indexing regardless of the actual link. I don't think this causes any problems, but it is confusing labelling.

paciorek:
Verifying that PR #1200 fixes this issue.

PR #1200, opened 2022-01-29
WIP: fix size of `dep_dmnorm_identity_coeff` for dwish/dinvwish conjugacy with dynamic indexing

paciorek:
We weren't detecting that `dep_dmnorm_identity_coeff` is a vector of scalars (one per dependent) when using dynamic indexing with `dwish` and `dinvwish` in the identity relationship case.

This fixes that and adds a test, including testing that `dep_dmnorm_identity_coeff` is correct in both the `dmnorm` and `dwish` cases, correctly zeroing out potential dependencies that are not actual dependencies because of the current dynamic index values.

One thing to check (consult with Daniel) is why in lines 574 and 960 of `MCMC_conjugacy.R`, we are not simply setting `2`=0. I'd like a second opinion to try to jog my memory of why I left `2`=2 when I don't think we ever allow a 2-dim coeff in the case where we have a 2-dim target. We used to mistakenly allow this.

### DIFF
--- a/packages/nimble/R/MCMC_conjugacy.R
+++ b/packages/nimble/R/MCMC_conjugacy.R
@@ -572,17 +572,15 @@ conjugacyClass <- setRefClass(
             ## July 2017
             targetNdim <- getDimension(prior)
             targetCoeffNdim <- switch(as.character(targetNdim), `0`=0, `1`=2, `2`=2, stop())
-            targetCoeffNdimSave <- targetCoeffNdim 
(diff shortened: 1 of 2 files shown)


paciorek:
See also issue #1175 .

paciorek:
@danielturek could you take a look at my question in this thread and see if you have any memory of this?

paciorek:
This fixes issue #1194

PR #1202, opened 2022-02-02
WIP: fix bug giving incorrect BNP sampling when using CRP with unknown base measure parameters

paciorek:
Previously we were assigning a default sampler for base measure parameters. But because we assign a wrapped sampler to the cluster parameters that avoids sampling them when a cluster is not occupied, the updating of the base measure parameters is incorrect. 

This PR assigns a special slice sampler to the base measure parameters that determines which dependent cluster parameters are for occupied clusters and uses only those in updating. Also adds testing and adds a paragraph to the BNP chapter of the manual.


perrydv:
@paciorek LGTM, but my knowledge of this code is somewhat limited.

danielturek:
@paciorek Maybe make the "name" argument of `sampler_slice_CRP_base_param`, match the function name?  At present it's different (I presume leftover from bing copied) as: `name = 'sampler_slice_CRP_cluster_params'`

Maybe move the line:
```
usedClusters <<- nimLogical(numPossibleClusters, FALSE)
```

to instead be just *before* the for loop, setting some elements to TRUE.  It would function the same, but seems logically cleaner to have:
```
        usedClusters <<- nimLogical(numPossibleClusters, FALSE)
        for(i in 1:numPossibleClusters) 
            usedClusters[model[[dcrpNode]][i]] <<- TRUE
        usedDeps <<- usedClusters[clusterIDs]
```

I'll send other comments if I have more time to look over this before tomorrow.

paciorek:
I've addressed the suggestions above.

PR #1204, opened 2022-02-08
Only partially evaluate known-deterministic functions in models

### LINKED ISSUE #1188: nimbleExternalCall breaks model building with constant expression
Reproducible example from a user on nimble-users

```
sink('add1.h')
cat('
 
 extern "C" { double ode(double x0,double k, double tend);}

 #include <vector>
 typedef std::vector< double > state_type;

class f_ode {

    double m_k;

public:
    f_ode( double k ) : m_k(k) { }

    void operator() ( const state_type &x , state_type &dxdt , const double /* t */ )
    {
        dxdt[0] = - m_k*x[0];
    }
};

   
    ')
sink()
sink('add1.cpp')
cat('

#include "add1.h"
// #include <boost/numeric/odeint.hpp>  
 #include <iostream>

using namespace std;
    using namespace boost::numeric::odeint;
   
 double ode(double x0,double k, double tend) {
 
   
    state_type x(1);
    x[0] = x0; // start at x=1.0, p=0.0
   
   
    f_ode myf(k);
    size_t steps = integrate( myf ,
            x , 0.0 , tend , 0.1 );
   
    return(x[0]);
 
}
')
sink()
system('g++ add1.cpp -c -o add1.o -std=c++11')
Radd1 <- nimbleExternalCall(function(x0 = double(0), k = double(0),tend = double(0)){}, Cfun =  'ode',
headerFile = file.path(getwd(), 'add1.h'), returnType=double(0),
oFile = file.path(getwd(), 'add1.o'))
## If you need to use a function with non-scalar return object in model code,
## you can wrap it  in another nimbleFunction like this:
model_add1 <- nimbleFunction(
     run = function(x = double(0)) {
     #   ans <- numeric(length(x))
         nimPrint(Radd1(21, 0.5, 10.0)) # use x inst
         return(1.0)
         returnType(double(0))
     })
demoCode <- nimbleCode({
    for(i in 1:4) {x[i] ~ dnorm(0,1)} ## just to get a vector
   #  z<-model_add1(1) # fails
    y <- 1 # works with next line
     z<-model_add1(y)
})


demoModel <- nimbleModel(demoCode, inits = list(x = rnorm(4)),
check = FALSE, calculate = FALSE)
CdemoModel <- compileNimble(demoModel, showCompilerOutput = TRUE)

```

perrydv:
fixes #1188 

The problem was model code such as

```
  x <- nec(1)
```

where `nec` is a `nimbleExternalCall`.  The partial evaluation system (in `replaceAllConstants`) attempted to evaluate that the same way it would `c+2` in `y[ c + 2]`, if `c` is a constant.  But a nimbleExternalCall can't be evaluated in R, so it errors out.

Actually, the partial evaluation system would also be incorrect if for some reason a user wanted a declaration with a non-deterministic function (that would be weird, but perhaps allowed?). It would get evaluated once as if deterministic. 

My fix here is to split `nimbleOrRfunctionNames` in two, with one called `nimblePreevaluationFunctionNames`, which are basically simple scalar deterministic functions.  (We never partially evaluate non-scalar functions anyway, IIRC).  Then I use the latter as the only functions to be partially evaluated in the `replaceAllConstants` step.  This avoids trying to call the nimbleExternalCall or any other function not  in that vector of safe (known, scalar, deterministic) function names.

Note that this is a different step than the "replacements" work in BUGSdecl.

@paciorek @danielturek curious what you think of this solution.


danielturek:
@perrydv if `nec(1)` is in fact a deterministic function, then I don't suppose it's worthwhile to actually evaluate it (somehow), and replace-it-away?  But maybe that's far less feasible than I can imagine.

Also, this raises my own curiosity - say a user writes a nimbleFunction, where internally it relies on random draws, and hence isn't deterministic.  Then, I note the following happens when such a nimbleFunction is used (in a deterministic declaration) in a model.
  - When the function takes no arguments, it gets replaced-away by the first (random evaluation).  The deterministic declaration is hard-coded to however that function evaluated the first time.
  - When the function takes non-constant arguments, even if it doesn't use them, then it doesn't get replaced-away, and will indeed result in a deterministic model declaration, with a random outcome.

This is somewhat inconsistent, and maybe is something worth thinking about - some day.  If you want, I can file a NCT issue, with the quick example I just tried.

Also, unrelated, I also vaguely recall (once) seeing a user try a model including code such as:
```
x <- rnorm(...)   ## non-deterministic RHS
```
as you mentioned, and recall somehow this ran into a problem - didn't work, or at least was problematic enough to be noticed.

perrydv:
@danielturek 

It is not possible to have uncompiled evaluation of a `nimbleExternalCall`.  Whatever new feature would be involved in supporting that doesn't seem like it will be a higher priority than other things right now.  (Hence a model using a `nimbleExternalCall` must be built with `calculate =FALSE`.)

This PR disables partial evaluation of anything not on the new `nimblePreevaluationFunctionNames` list.  That means it disables partial evaluation of `r` distribution functions.  I'm not sure if that addresses your point.

paciorek:
@perrydv Are we ok to merge this in?  I have no particular reactions but also haven't thought deeply about this.

perrydv:
@paciorek I'm ok with merging it.

PR #1205, opened 2022-02-09
fix coeffSetter if the target is bool

### LINKED ISSUE #1201: incorrect LHS subset assignment into logical vector in compiled code
Using an indexing vector (with possibly duplicated entries) to assign into elements of a logical vector gives an incorrect result in compiled code. Assigning into a numeric vector is fine.

I haven't looked into the cause at all. 

```
myf <- nimbleFunction(
    run = function(clusters = double(1), cluster_in_use = logical(1)) {
        returnType(logical(1))
        cluster_in_use[clusters] <- TRUE
        return(cluster_in_use)
    })
cmyf <- compileNimble(myf)
cluster_in_use <- rep(FALSE, 10)
clusters <- c(2, 3, 2, 5)
myf(clusters, cluster_in_use)   # correct
##  [1] FALSE  TRUE  TRUE FALSE  TRUE FALSE FALSE FALSE FALSE FALSE
cmyf(clusters, cluster_in_use)  # incorrect
##  [1]  TRUE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
cmyf(2, cluster_in_use)  # incorrect
##  [1]  TRUE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
```

This is fine:

```
myf <- nimbleFunction(
    run = function(clusters = double(1), cluster_in_use = double(1)) {
        returnType(double(1))
        cluster_in_use[clusters] <- 99
        return(cluster_in_use)
    })
cmyf <- compileNimble(myf)
cluster_in_use <- as.numeric(1:10)
clusters <- c(2, 3, 2, 5)
myf(clusters, cluster_in_use)   # correct
## [1]  1 99 99  4 99  6  7  8  9 10
> cmyf(clusters, cluster_in_use)  # incorrect
## [1]  1 99 99  4 99  6  7  8  9 10
```

perrydv:
fixes #1201

### DIFF
--- a/packages/nimble/inst/include/nimble/nimbleEigen.h
+++ b/packages/nimble/inst/include/nimble/nimbleEigen.h
@@ -49,26 +49,31 @@ template<typename totLenT, typename knownDimT>
 template<typename T>
 struct nimble_eigen_traits {
   enum {nimbleUseLinearAccess = int(Eigen::internal::traits<T>::Flags & LinearAccessBit)};
+  typedef typename Eigen::internal::traits<T>::Scalar Scalar;
 };
 
 template<>
 struct nimble_eigen_traits<double> {
   enum {nimbleUseLinearAccess = int(1)};
+  typedef double Scalar;
 };
 
 template<>
 struct nimble_eigen_traits<int> {
   enum {nimbleUseLinearAccess = int(1)};
+  typedef int Scalar;
 };
 
 template<>
 struct nimble_eigen_traits<bool> {
   enum {nimbleUseLinearAccess = int(1)};
+  typedef bool Scalar;
 };
 
 template<typename V>
 struct nimble_eigen_traits<std::vector<V> > {
   enum {nimbleUseLinearAccess = int(1)};
+  typedef V Scalar;
 };
 
 template<typename T>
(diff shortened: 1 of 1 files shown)


paciorek:
@perrydv in line 720, you have:

```
 std::cout<<"Constructing coeffSetting with "<<dim1<<" "<<dim2<<std::endl;
```

I assume this was just for debugging and I can remove it. I'm seeing it produce output messages when using `getSamplesDPmeasure`.

paciorek:
@perrydv Also, I'm now seeing this after running an MCMC as I quite R in (some?) situations. Could it be related?

```
munmap_chunk(): invalid pointer
```

I can work on a reprex if helpful.

perrydv:
@paciorek  Oops that message was indeed for debugging.

The `munmap_chunk` message is not familiar.  The internet says it comes from memory management errors that are not seg faults.  I don't know if it's related to this bug fix.  If it's easy to reproduce I could take a look.  If it is related to this we could revert this.

paciorek:
I think the memory error was just a bug in my nimbleFunction coding and not anything related to the PR/nimble itself.

PR #1206, opened 2022-02-09
runCrossVal expand node names before extracting from samples matrix

danielturek:
`runCrossVal` formerly didn't expand node names, before using the node names to index the columns of a matrix of posterior samples.  This caused an error when extracting multivariate nodes, for example `x[1:2]`, when indexing the columns as:

```
samples[, "x[1:2]" ]     ## error
```


danielturek:
@paciorek Pending testing and such, maybe this could also make it into the upcoming release.

PR #1213, opened 2022-03-02
make nimbleList interface of nested nimbleList for predefined use sessionSpecificDll

perrydv:
When one nimbleList contains another, and the contained nimbleList is a predefined, the dll in which we look for SEXP-functions for accessing the C++ objects is the wrong dll.  It is the on-the-fly dll instead of the sessionSpecificDLL.  This problem is OS-specific and occurs on Mac but not Linux.  It has not been checked on Windows yet.  I added a fix that checks for the predefined status when generating the active binding code for the (nested) nimbleList interface class.  I also added some of the predefined entries to dynamicRegistrations.h, which was a step missing in the instructions for updating predefined nimbleLists.  Let's see if this passes tests, but note that due to which tests were run on which OS, this was not actually failing tests in the first place.

### DIFF
--- a/packages/generateStaticCode.R
+++ b/packages/generateStaticCode.R
@@ -15,6 +15,10 @@ directions <- 'Directions:
 4. Review changes this script has made to the predefinedNimbleLists.* files.
 5. Revert the temporary change to GENERATE_STATIC_CODE from step 2,
    and reinstall nimble.
+6. If desired (default = yes), manually update
+   inst/include/nimble/dynamicRegistrations.h to have entries for the three
+   new SEXP-returning functions for each predefined nimbleList.
+   See examples in that file.
 '
 
 how_to_add_a_new_predefined_nimbleList <- 'How to:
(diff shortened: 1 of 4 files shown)


paciorek:
@perrydv can we merge this in? I'm not sure of what your intention was w.r.t. checking on Windows.

perrydv:
@paciorek I am running some selected tests on Windows.  So far both test-nimbleList and test-waic have passed.  ( I had a note that test-waic might also be affected by this issue. )  I am running test-mcmc right now.  I think those should provide enough coverage to decide this change is ok.

For the record, the problem that this PR fixes does not seem to have ever emerged as a bug in an actual use case.  We discovered that test-nimbleList was not passing on OS X due to this OS-specific issue.  Our test suite does not run every test on each major OS, so the test failure had gone undetected because test-nimbleList was being run on linux only.  So if my selected tests pass on Windows too, I think we're good.  I'll update you later.

paciorek:
Ok, this is the last PR to merge in for release apart from the post_pred docs stuff.
Just adding this comment so you have a flag in your email to either merge this or let me know to do so.

perrydv:
@paciorek Please go ahead and do the merge.

PR #1215, opened 2022-03-02
changed addSampler scalarComponents argument to expandComponents

danielturek:
Do not merge.  Opening PR for testing.

### DIFF
--- a/UserManual/src/chapter_MCMC.Rmd
+++ b/UserManual/src/chapter_MCMC.Rmd
@@ -250,21 +250,23 @@ mcmcConf <- configureMCMC(Rmodel, control = list(scale = 3, adaptive = FALSE))
 
 When adding samplers to a configuration using `addSampler`, the default control list can also be over-ridden.
 
-#### Adding samplers to the configuration: *addSampler*
+#### Adding samplers to the configuration: *addDefaultSampler* and *addSampler*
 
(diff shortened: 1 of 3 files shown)


danielturek:
This PR makes some changes to the MCMC configuration class.  It's basically a refactoring of code to add additional entry points to configuration methods to increase options that are available.  Notably, it should not change the behaviour of any existing code.

Most significant, it adds a new option to the `addSamplers` method, using the newly added `default` argument: `conf$addSamplers(..., default = TRUE)`.  Using this new option, the *default* set of samplers will be added for the specified nodes.  Importantly, provides access to the "logic" of default sampler determination and assignment (conjugate, or otherwise), *without* having to create an entirely new MCMC configuration object.  This was sorely missing functionality (and an oversight) in the original implementation.

Also, it changes the name of the `scalarComponents` argument of `addSamplers` to instead `expandTarget`, which is not only more accurate, but in fact actually correct.  The old name (using the word "scalar") was incorrect in the case of multivariate nodes / variables.  The old argument name is still supported, but with a warning message saying it will be deprecated sometime.

Also, addition of a new `nodes` argument to `addSampler`, which functions similar to the still existing `target` argument, except `nodes` *always* undergoes node expansion.  This mimics the functionality of the original entry point for default sampler assignment, that being the `nodes` argument of the `initialize` method of an MCMC configuration object.

Noting that the following two lines are equivalent:
```
conf$addSampler(target = X, expandTarget = TRUE)
conf$addSampler(nodes = X)
```
Either the `target` or the `nodes` argument can be used in combination with `default = FALSE` or `default = TRUE`, but I expect the most commonly used (most convenient) use cases will using the `nodes` argument when `default = TRUE`, and using the `target` argument when `default = FALSE` (the default).

PR #1222, opened 2022-05-22
added replaceSamplers method to MCMC configuration

danielturek:


danielturek:

danielturek:
@perrydv @paciorek I'd like to merge this in to devel.

Anyone, please take a look at `replaceSamplers` method.  I did a little juggling so as to preserve the current sampler configuration, in the event that the call to `addSampler` errors out.

@paciorek Note the change in precision (number of digits) in the test-mcem gold file.  For reasons beyond me, the current gold file was differing in the 7th decimal place on the CI testing (but not on my computer, which agreed with the old gold file).  So @paciorek please take a look at the changes to `test-mcem.R`.

paciorek:
Yes, tolerances can be hard to track down. I'm not worried about the 6 vs. 7 digits. My suggestion would be to just change the gold file and not monkey with R's digits option. But I guess the issue is that you're not able to change the gold file such that testing passes both locally and in CI?

danielturek:
@paciorek That's right - two computers I used both agreed with the **old** gold file.  The CI testing was giving a different result (different than it used to give, which is a mystery) in the 7th decimal place.  So I could not get it to pass both locally and with the automated CI tests.

PR #1223, opened 2022-05-24
refactor nimbleMCMC to pull out model creation

danielturek:
This will refactoring be helpful for future `nimbleHMC` function.


danielturek:
@paciorek Here is one example of the mcem testing failure (which is the first commit in this PR):

[https://github.com/nimble-dev/nimble/runs/6576403030?check_suite_focus=true](https://github.com/nimble-dev/nimble/runs/6576403030?check_suite_focus=true)

I just merged the "patch" (PR #1224 ) into this PR, so testing should pass on this PR now.

PR #1228, opened 2022-08-01
prevent infinite recursion in checkConjugacy

danielturek:
Title says it all.

Need to see what testing says.


danielturek:
@perrydv @paciorek I'm ready to merge this, in, but welcome any code review.

The code prevents a recursive case, which can arise due to the `strutureExpression` mechanism in conjugacy checking.  Given how `strutureExpression` is handled, I don't think there can be any cases where this patch changes the results of a conjugacy check.

paciorek:
Just getting to this. No comments here -- I'm not seeing the case where it happens but seems like a safe enough strategy.

danielturek:
Thanks for the comments @paciorek

PR #1239, opened 2022-10-29
fix issue 1219 by mangling arg names for pointAtAll

paciorek:


perrydv:
@paciorek Thanks this looks good to me.  Sometimes people put two underscores, like `i__` to make it even less likely to clash.  I have no strong opinion.

PR #1241, opened 2022-11-01
trap non-existent run/methods in nimbleFunction()

paciorek:
This addresses NCT issue 374, where a user mistakenly left out the run function by doing

```
a <- nimbleFunction(run <- function(){})
```

@danielturek @perrydv let me know if you can think of any cases where creating a nimbleFunction with only setup code is something we want to allow.

### DIFF
--- a/packages/nimble/R/MCEM_build.R
+++ b/packages/nimble/R/MCEM_build.R
@@ -87,18 +87,17 @@ calc_E_llk_gen = nimbleFunction(
 
 
 ## helper function to extract ranges of nodes to be maximized
-getMCEMRanges <- nimbleFunction(name = 'getMCEMRanges',
- setup = function(model, maxNodes, buffer){
+getMCEMRanges <- function(model, maxNodes, buffer) {
     low_limits = rep(-Inf, length(maxNodes) ) 
     hi_limits  = rep(Inf,  length(maxNodes) )
     nodes <- model$expandNodeNames(maxNodes)
     for(i in seq_along(nodes)) {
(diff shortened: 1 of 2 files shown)


paciorek:
Tests pass except that MacOS run failed for unrelated reasons.

perrydv:
@paciorek This does preclude using a nimbleFunction purely as a data structure with no methods, which can currently be done like this:

```
foo <- nimbleFunction( setup = 
   function(a) {
      b <- numeric(5)
      setupOutputs(a, b) # Mark a and b as member data
})
foo1 <- foo(4)
cfoo1 <- compileNimble(foo1)
cfoo1$a
```

This isn't a common usage, if anyone is doing it at all, but it does work.

I don't have a different great idea for trapping this specific user mistake (in use of R, not of nimble per se).  We could target it very specifically like this

```
if(isTRUE(nimbleOptions("check_setup_for_accidental_run_assignment")) {
sub_setup <- substitute(setup)
if(is.call(sub_setup))
  if(length(sub_setup) > 1)
    if(sub_setup[[1]] == "<-")
      if(sub_setup[[2]] == "run")
        stop("Did you use 'run <-' instead of 'run = ' when calling nimbleFunction?")
}
```

If you want to introduce a new requirement that run or a method must be provided, it doesn't seem likely to break anyone's code, but it is a subtle change.

paciorek:
@perrydv Yeah, good point, I don't think we want as blanket a trap as I have here. And for now, trapping this particular user error is probably more than we want to do.

PR #1242, opened 2022-11-03
add check for unused constants back before if-then-else processing

paciorek:
(Fixes NCT 319)


paciorek:
The failure is in `test-nimbleLists.R`, which should seem to involve inits. Also, it passes when run locally.

PR #1245, opened 2022-11-06
revise docs wrt revamp of posterior predictive handling

paciorek:
@danielturek @perrydv I've drafted modifications to `onAttach` and sections of the manual related to the posterior predictive revamp. Please take a look at the 'files changed' for this PR and either make changes to the branch or post thoughts here and I'm happy to make edits based on our consensus. I tried to be reasonably careful here, but I definitely want both your eyes on how we phrase things and potentially also on where in the manual we talk about this. Right now there is a new subsection at the end of the user-defined samplers section, and a new section just below where we talk about default samplers in the MCMC chapter.

Also, NEWS now has most of what we need for the release (but doesn't include bullets for stuff still pending in unmerged PRs).
 - Note that it is now NEWS.md so it can render correctly in browsers.
 - @danielturek I think somehow in converting to NEWS.md, we (me?) lost your NEWS items from the posterior predictive revamp. I copied them in manually but then did some rearranging and rewording of the bullets related to the revamp. Please take a look.


danielturek:
@paciorek I looked over all the changes.  Some comments and small suggestions are below.

In the MCMC chapter, line 173 doesn't seem accurate to me, specifically how it says "If the node has no stochastic dependents...."  The PP sampler no longer has to do with the presence of stochastic dependents, but rather any downstream data nodes.  Maybe changing line 173 to:
"1. If the node has no data nodes in its entire downstream dependency network, a `posterior_predictive` sampler is assigned.  This sampler updates the node in question *and all downstream stochastic nodes* by simulating new values from each node's prior distribution. As of version 0.13.0, the operation of the `posterior_predictive` sampler has changed in order to improve MCMC mixing. See Section \@ref(sec:post-pred-sampling)."

I think the new subsection "Sampling posterior predictive nodes " is well explained.  In this section, line 194, maybe change the text:
replace: "better than an MCMC that conditions on those nodes, ..."
with: "better than an MCMC that conditions on the values of posterior predictive nodes, ..."

Also in line 194, maybe replace the final sentence with:
"At the end of each MCMC iteration, the posterior predictive nodes are sampled by the `posterior_predictive` sampler based on their conditional distribution."

In chapter_WritingNimbleFunctions.Rmd, new section "User-defined samplers and posterior predictive nodes", line 620, I think you should explain a little more explicitly:
replace: "This behavior will also pertain to user-defined samplers, without problems in most cases."
with: "This behavior will also pertain to user-defined samplers.  There will be no problems when sampler implementation follows the (preferred) model-generic programming style of using `getDependencies` in the `setup` function to determine dependent node sets.  However..."

also:
replace: "create an invalid MCMC if the handling of dependencies in the user-defined sampler ..."
with: "create an invalid MCMC if the handling of dependencies (specifically w.r.t. inclusion of posterior predictive dependencies) in the user-defined sampler ..."

In the package startup message.  It currently says "changed in a way that could affect user-defined samplers in some situations"
Do we want to be more accurate, and say:
"changed in a way that could make some user-defined samplers generate incorrect results."
This re-wording, while more alarmist, does clearly convey to users that their samplers could be affected in an important and adverse manner.  **I'm on the fence about this change.**  I defer to @paciorek and @perrydv opinions.

I'm not sure what happened with the NEWS file, but thanks @paciorek for catching that.

Typo in the first NEWS item: "Thoroughly revampled"

Typo in NEWS: "This reodering can be disabled"

Typo "The MCMC confiuration object "

Also thanks for using proper Imperial spelling in the MCMC-related NEWS items.

paciorek:
Thanks, @danielturek , very helpful. I'll make the changes along those lines. A couple comments:

1. Good catch on line 173.That text is about default sampler assignment was not something I changed, so it reflects the old, out-of-date language.
2. Regarding Imperial spelling, I'm sorry to inform you that that was not my intention.

danielturek:
Re-reading line 173 of MCMC chapter, I think I prefer your use of "conditional distribution" rather than my suggestion of "prior distribution".  So, my updated suggestion for that line 173 (only changing "prior" to "conditional") is:

"1. If the node has no data nodes in its entire downstream dependency network, a `posterior_predictive` sampler is assigned. This sampler updates the node in question *and all downstream stochastic nodes* by simulating new values from each node's conditional distribution. As of version 0.13.0, the operation of the `posterior_predictive` sampler has changed in order to improve MCMC mixing. See Section @ref(sec:post-pred-sampling)."

danielturek (approved):

perrydv:
Thanks @paciorek for drafting this and @danielturek for revisions.

In the MCMC chapter, should we define posterior predictive nodes in \#sec:post-pred-sampling ?  Otherwise this sections looks good to me.

In the writing nimbleFunctions chapter (15), I think it would help to be more explicit.  Suggested text below.  It could alternatively or in addition help to give a short example, even if just in text without code.

The updated Rd entry for posterior predictive sampler looks good.  However I did think that the two sentences "Nimble allows... they are not needed when using nimble" may be confusing because they are a digression to a situation about not using posterior predictive nodes rather than directly describing posterior predictive samplers, but that is not exceedingly clear.  Maybe move them to the end and put in parentheses? (I know, they are not new to this PR.)

I think the less alarmist version of the new startup message is fine.  Very few users will have written a sampler, and anyone who has should realize what the stakes are if we are directing them to look a section of the user manual for updates on the MCMC system.

Should we revise the Rd entry for `getDependencies` to explain the nimbleOption it now checks?  It looks like that is currently untouched in the PR.

Suggested text for Chap 15:

As of version 0.13.0, NIMBLE's handling of posterior predictive nodes in MCMC sampling has changed in order to improve MCMC mixing. Samplers for nodes that are not posterior predictive nodes no longer condition on the values of the posterior predictive nodes. This is done by two changes: 

- turning off posterior predictive nodes as dependencies in the use of `getDependencies` when building an MCMC.
- moving all posterior predictive samplers to be last in the order of samplers used in an MCMC iteration.

The first change calls for careful consideration when writing new samplers.  It is done by making `buildMCMC` set a nimbleOption (described below) that tells `getDependencies` to ignore posterior predictive nodes (defined as nodes with no data nodes in their dependencies, dependencies of their dependencies, and so on) before it builds the samplers (i.e. runs the setup code of each sampler).  That way, the setup code of all samplers that use `getDependencies` will automatically comply with the new system.

User-defined samplers that determine dependencies using `getDependencies` should automatically work in the new system (although it is worth checking).  However if a user-defined sampler manually specifies dependencies, and if those dependencies include posterior predictive nodes, the MCMC results could be incorrect.  Whether incorrect results occur could depend on other parts of the model structure connected to node(s) sampled by the user-defined sampler (i.e. the target node(s)) and/or posterior predictive nodes.  Therefore, it is strongly recommended that all user-defined samplers rely on `getDependencies` to determine which nodes depend on the target node(s).

There are several new nimbleOptions available to take control of the behavior just described.

- `getDependenciesIncludesPredictiveNodes` determines whether posterior predictive nodes are included in results of `getDependencies`.  This defaults to `TRUE`, so that outside of building samplers, the behavior of `getDependencies` should be identical to previous versions of nimble.
- `MCMCusePredictiveDependenciesInCalculations` give the value to which `getDependenciesIncludesPredictiveNodes` will be set while building samplers.  This defaults to `FALSE`.
- `MCMCorderPosteriorPredictiveSamplersLast` determines whether posterior predictive samplers are moved to the end of the sampler list.    This default to `TRUE`.  Behavior prior to version 0.13.0 corresponds to `FALSE`.

Here are some examples of how to use these options:

- If you want to experiment with the behavior of `getDependencies` when ignoring posterior predictive nodes, do `nimbleOptions(getDependenciesIncludesPredictiveNodes = FALSE)`.
- If you want to have MCMC samplers condition on posterior predictive nodes, do `nimbleOptions(MCMCusePredictiveDependenciesInCalculations = TRUE)`.
- If you want to get identical MCMC samplers, including their ordering, as in versions prior to 0.13.0, do `nimbleOptions(MCMCusePredictiveDependenciesInCalculations = TRUE)` and `nimbleOptions(MCMCorderPosteriorPredictiveSamplersLast = FALSE)`.

paciorek:
@danielturek @perrydv thanks, both. I'm editing to include all your suggestions.

Also @perrydv we do have text about checking the nimbleOption in the raw help text for `includePredictive` arg in help for `getDependencies`. Without looking more deeply, I'm a bit confused about when that text was added/why it's not showing up in the Git PR diff. But it looks fine to me now.

PR #1248, opened 2022-12-07
misuse of nodeFunctionID argument in nodeName2GraphIDs method

danielturek:


paciorek:
@danielturek @perrydv should I start working on the release? Let  me know your thoughts on the status of this PR.

danielturek:
@paciorek @perrydv I'm ok with this PR, but lord knows I welcome any degree of code review.

Note that the decision was made to keep this check at the time of `buildMCMC` (rather than at the time of `addSampler`) because a user could modify nimble system options between the times of MCMC configuration and building.

paciorek:
A couple comments/thoughts:

1. I do think we should add a few tests with models where we check that our determination of predictive nodes is correct, in particular checking dependencies in non-predictive node samplers and in predictive node samplers. I think we should include scalar and mv nodes. @danielturek if you'd like me to work on that, please let me know.
2. I think it would be clearer to use `expect_{silent,message}` rather than using `NA` with `expect_error`, the latter of which seems hard to read/obscure to me.

perrydv:
@paciorek @danielturek I agree that having more new tests, beyond the test of correct error-trapping of a joint sampler on a non-PP + PP node, would be important.  @paciorek I'm guessing it would make sense for you to jump in on that if you can.  We could also test whether, after `buildMCMC`, each sampler has the correct dependencies determined, i.e. whether the bug was fixed.

I think 
`samplingPredictiveNode <- if(any(targetScalarComponentsIsPP)) TRUE else FALSE`
could be simply
`samplingPredictiveNode <- any(targetScalarComponentsIsPP)`
or slightly more robustly
`samplingPredictiveNode <- isTRUE(any(targetScalarComponentsIsPP))`

If that is shifted up a line, then
`if(any(targetScalarComponentsIsPP) && ...`
could become
`if(samplingPredictiveNode && ...`

but these are minutia compared to setting up tests.

paciorek:
Ok, I added some test cases. Please take a look and see if you have thoughts on other model configurations to test. I'll also think more about it on Monday morning; this was fairly quick.

The tests are *all* related to correct sampler dependencies. @perrydv  you referred to that as *also* so you might have had something else in mind -- perhaps checking specifically if a node is flagged as posterior predictive?

danielturek:
Thanks for keeping this moving forward.  Yes, please add whatever additional testing is necessary.

paciorek:
I tweaked my tests so that the mv one would have failed if applied to 0.13.0. (Whether the PP node determination is incorrect is pretty happenstantial, depending on exactly what happens with the incorrect IDs.)

I only added a couple model cases, but nothing else particular is occurring to me. @perrydv @danielturek if you have thoughts of other cases, please weigh in. Otherwise, it probably makes sense for me to push to CRAN by the end of the day.

perrydv:
@paciorek @danielturek I think it's ok.  One thing I noticed is that in buildMCMC we could skip all the new checking entirely in the case that there are no stochastic predictive nodes (i.e.  conf$model$getPredictiveNodeIDs() has length 0).  That would have the appeal of avoiding this entire area when it's not relevant.

danielturek:
@paciorek @perrydv Thank you for the opportunity for final code review before proceeding.

I pushed one change (one commit with two changes).  The changes are stylistic only.  Although the changes I made are minor, I did undertake a careful review the code changes.  LGTM.

I did not review the testing (changes in test-mcmc.R), I am trusting the testing is good and what we want.

Proceeding with mini-release of this change thumbs up from me.

Thank you both for your help on this.

paciorek:
I merged this manually. (In retrospect I shouldn't have, as commits didn't get squashed.)

PR #1250, opened 2022-12-08
added checks for MCMC thin being negative or non-integer

danielturek:


danielturek:
@paciorek Can this go in the mini-release, also?

paciorek:
It could. I guess we might as well,  though there is something to be said for having the single focal change, perhaps.

danielturek:
@paciorek Good point.  I'm fine either way - specifically, if this does not go into version 13.1.

paciorek:
I'm leaning towards having 0.13.1 just be a single clean bug fix.

Issue #1252, opened 2022-12-10
improve error message for disallowed `model[[nodes]]` and `model[[nodes[i]]]` usage


paciorek:
Currently, we (unhelpfully) output:

```
> cmcmc <- compileNimble(mcmc, project = model, showCompilerOutput = TRUE) Compiling 
    [Note] This may take a minute. 
    [Note] On some systems there may be some compiler warnings that can be safely ignored. 
Error in exists(as.character(varOrNodeExpr), x, inherits = FALSE) : 
      first argument has length > 1
```

User message from 2022-12-07 has a reprex, but it would be easier to just make a minimal reprex when we pursue this.

paciorek:
Actually `model[[nodes]]` is well-trapped: `Error: Length of nodes requested from model using '[[' is 2. It must be 1.`

paciorek:
I've added error trapping for `model[[nodes[i]]` in `determineNdimsFromNfproc`. See PR #1259.

Issue #1237, opened 2022-10-26
Better error-trapping of incorrect use of `[[` in compilable code


perrydv:
We had a user try to use `[[` to index numeric objects in a `nimbleFunction`.  The error from compileNimble was untrapped and inscrutable.  We could make the keyword processor for `[[` do better.

paciorek:
Actually there is a bug in the `[[` keyword processor. `callerCode` is used but not defined. It appears that the `[[` handling of certain cases has never been exercised (e.g., handling of `symbolNimbleList`). 

I am working through this, in conjunction with issue #1252, which also involves this keyword processor.

paciorek:
Ok, once `callerCode` is defined, the error is now:

```
Error: In sizeNFvar: First argument is not a nimbleFunction or a nimbleList
 This occurred for: nfVar(ARG1_x_,"2")
 This was part of the call:  y = nfVar(ARG1_x_,"2")
```
This is ok, but I am wondering if we should add text that says something like: "List-like subsetting with `[[` and `$` in nimbleFunctions can only be done with a nimbleFunction or nimbleList".

@perrydv any opinion?

perrydv:
Yes a better message makes sense to me.

PR #1259, opened 2023-01-21
trap list-like syntax errors in nimbleFunction run code

paciorek:
This address issues #1237  and #1252 , by

- fixing a bug in handling `nimbleList[['foo']]` syntax
- adding a test for the former case
- error trapping `model[[nodes[i]]]`

Do not merge until deciding on the question of whether to enhance the `sizeNFvar` error message to be discussed in issue #1237.


perrydv:
LGTM. We can see room for even more error-trapping, but this is a bug fix and improved error-trapping for now.

PR #1266, opened 2023-02-15
MCMC unsampled nodes determination handles RW_PF sampler

### LINKED ISSUE #1265: Warning about unsampled nodes is unnecessary for pMCMC
When using pMCMC (sampler RW_PF) from nimbleSMC, the warning
`[Warning] No samplers assigned for 25 nodes, use conf$getUnsampledNodes() for node names`
appears from the compileNimble step because there are, indeed, no samplers on the latent-states of the state-space model.
Can we provide a way to manage that?  Or when the pMCMC sampler is added emit a message saying it is ok to ignore the warning later?

danielturek:
Fixes #1265


paciorek:
@danielturek ok to merge?

danielturek:
@paciorek Yes.

Issue #1263, opened 2023-01-31
handling of deterministic data nodes


paciorek:
A user post highlights that we allow a user to set a deterministic node as data but then we ignore that information in determining predictive nodes and assigning samplers, and we overwrite the 'data' node values without any messages AFAIK.

Some options:
1. Modify `setData` to error out or warn the user and set the data flag to `FALSE`.
2. Modify our handling of such cases more generally to have a node not be predictive if there is a downstream deterministic data node. This would (I think) essentially allow one to introduce constraints by setting a downstream deterministic node as data.

paciorek:
`help(setData)` indicates that designating a  deterministic model node as 'data' will result in an error.
But no error is actually thrown.

paciorek:
Never mind about item 2. above. It's not consistent with how `calculate` works, in that `calculate` would need to return a boolean about whether the current value is equal to what is being calculated.

danielturek:
As best I recall, we've never considered (or really programmed for) deterministic nodes to be set as `data`.  Nothing is preventing this, setting a deterministic node as `data` will "work" just fine, with data flags being created and set (for deterministic variables) in the model's internal `isDataEnv` structure.  These data flags would never be used by the MCMC (since it only accesses these flags for stochastic nodes), but the querying of these flags and such (e.g. `model$isData`) would work fine.

It's a good question to consider, where to put in checks / warnings for this case.  I guess (offhand) probably in `model$setData`, where we could issue an error, if one sets a data value for a deterministic node.

Setting a deterministic node as `data` certainly would not ensure the values persist.  They would get overwritten by the first `model$calculate()` call of the relevant nodes, which would happen at some point in model initialization, or the MCMC, or otherwise.

paciorek:
Ok, I will work on this, in particular:

- in `setData` return an error when trying to set deterministic nodes as data
- noting that this is not possible in the manual

We could have the error actually be a warning and tell them we are not setting the deterministic nodes as data, but my thought at this point is that we want to force the user to reconsider what they are doing.

paciorek:
Some headaches:

- Need to exclude `RHSonly` nodes, but currently these are sometimes flagged as data.
- Will probably want to return an error if try to set deterministic nodes via `constants`, by similar logic to erroring if set deterministic nodes via `data`.

PR #1269, opened 2023-02-23
error trap deterministic data nodes

paciorek:
This addresses issue #1263 by adding an error trap in `setData`.


danielturek:
@paciorek  I see this addresses the case of a variable with both stochastic and deterministic entries... and requiring that NA's be provided for the deterministic elements.  That was my only concern.

paciorek:
`RHSonly` nodes were causing problems. I've redefined `isDeterm` to not be `!isStoch()` but rather check for "determ" and `setData` now will set values for RHSonly nodes but not flag them as 'data' nor error out if provided via `data`.

Will see what happens with testing...

Issue #1218, opened 2022-03-24
inconsistent use of offset in internal conversion functions


perrydv:
I noticed that `SEXP_2_int` in the C++ code uses the `offset` argument inconsistently for double vs integer or logical input.  I want to check if this is correct.

perrydv:
It looks like the offset argument to SEXP_2_int is non-zero from only one place, which is via C_getDependencyPathCountOneNode, called ultimately only from MCMC conjugacy determination, with offset = -1 to convert from an R index to a C index.  The SEXP input is a graphID.  I will make a pull request removing offset from SEXP_2_int and doing the -1 directly in the once place it is needed. 

In summary, the offset argument looks like it was designed in general for potentially other use cases, but in practice now there is only one use case so for future clarity I will isolate the offset step to that place.

PR #1277, opened 2023-03-11
Clean up use of offset argument in SEXP_2_int

perrydv:
Issue #1218 raised the issue that the offset argument was applied only for one type of input.

It turns out that offset is only non-zero (-1) when called from one pathway for graphs.  This PR removes offset as a general argument to SEXP_2_int and instead does the -1 step in the one place where it is used.


paciorek:
Test failure is unrelated. @perrydv should we merge?

Issue #1268, opened 2023-02-22
bug in modelDef splitVertices method ?


danielturek:
@perrydv , I think the current incarnation of the `modelDef` class (and associated processing) is largely your work.

This has to do with the `splitVertices` function, which exists in `BUGS_modelDef.R` on line 1637.

I've reduced this to a minimally reproducible example.  I think something is amiss here: model building errors out, when I think it shouldn't.  I began a deep dive into this, but I think it would be quicker and easier for you.

Example below:

```
library(nimble)

dDist <- nimbleFunction(
    run = function(x = double(0),
                   theta1 = double(1),
                   theta2 = double(1),
                   log = integer(0, default = 0)) {
        returnType(double())
        return(0)
    }
)

code <- nimbleCode({
    for (i in 1:2) {
        for (j in 1:2){
            b[i,j] <- X[i, index[j]]
        }
    }
    for (i in 1:2) {
        y[i] ~ dDist(
            theta1 = a[1:2, index[i]],
            theta2 = b[1:2, 1]
        )
    }
})

Rmodel <- nimbleModel(code, constants = list(index=c(1,1)))

## Error in if (is.na(uniqueCurrentVertexIDs[1]) | currentVertexCounts[uniqueCurrentVertexIDs[1]] !=  : 
##  missing value where TRUE/FALSE needed
```

paciorek:
A more minimal example; the user-def dist is not needed.

```
code <- nimbleCode({
    for (i in 1:2) {
        for (j in 1:2){
            b[i,j] <- X[i, index[j]]
        }
    }
    for (i in 1:2) {
        y[i] <- sum(a[1:2, index[i]])
        
    }
})

Rmodel <- nimbleModel(code, constants = list(index=c(1,1)))
```

Also it only seems to happen if `index` is `c(1,1)`. Perhaps a scalar vs. vector issue.

paciorek:
Fixed by #1279

PR #1279, opened 2023-03-13
Fix splitVertices to address issue #1268

perrydv:
Issue #1268 gives a corner case where `splitVertices` breaks, resulting in failure to build a very simple model.

The cases covered by `splitVertices` are quite varied, and the code is not simple, and it has not had problems in a long time, and we have a longer-term development effort to renovate how we handle graphical model relationships.  For these reasons, the intention here is as simple a fix as possible.

The details are as follows.  There is a `maxVertexID` determined early on in model building that attempts to be an upper bound on the possible number of graph vertices (i.e. nodes) needed for a model.  A vector of that length is then used to track how many scalar elements correspond to each vertex.  It turns out sometimes `splitVertices` results in skipping vertex IDs, with the result that a vertexID could end up larger than the `maxVertexID`, but it would be rare for it to happen (as in Issue #1268).  The `maxVertexID` logic wasn't wrong, but to strictly ensure that no vertexID is larger than it, we would have to go in and handle any cases where skipping occurs so that it would never occur.  This would be a pain and involve careful rethinking that would not be productive for code that is on the way out anyway.  Instead I have made a proposed fix that simply checks if an indexID gets larger than what is already being tracked and extends the tracking vector if necessary.  I think this is a minimal fix and we will see what testing turns up. 

Note that in any case, the vertexIDs at issue are only temporary, as they are before topological sorting, so any skipped indices do get cleaned up at a later stage of model building.

I also have a slightly more involved additional fix that could reduce some of the index-skipping behavior, and I might like to see if it passes tests, but even if it does there might be a case for omitting it to stick to the absolute simplest fix right now, even inelegant as it is.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -1679,7 +1679,8 @@ splitVertices <- function(var2vertexID, unrolledBUGSindices, indexExprs = NULL, 
     if(all(is.na(var2vertexID)))
         currentVertexCounts <- rep(0, maxVertexID)
     else
-        currentVertexCounts <- tabulate(var2vertexID, max(max(var2vertexID, na.rm = TRUE), maxVertexID))
+      currentVertexCounts <- tabulate(var2vertexID, max(max(var2vertexID, na.rm = TRUE), maxVertexID))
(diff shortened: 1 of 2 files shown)


paciorek:
@perrydv test failure is unrelated. Shall we merge?

Issue #1261, opened 2023-01-25
`max` and `min` of a scalar are not error-trapped


paciorek:
These fail compilation at the C++ compilation stage.

```
a <- nimbleFunction(
    run = function(x = double(0)) {
        returnType(double())
        y = max(x)
        return(y)
    }
)

ca <- compileNimble(a, showCompilerOutput = TRUE)

code <- nimbleCode({
    y <- max(a)
})
m <- nimbleModel(code, inits = list (a=7))
cm <- compileNimble(m)
```

We could error trap them or ensure they work like most of our other vectorized functions.
For simplicity, I may just look into error trapping.

PR #1281, opened 2023-03-13
error trap reduction ops on scalar inputs

paciorek:
This addresses issue #1261


paciorek:
@perrydv just flagging you in case you want to take a look at this change to size processing in the compiler. It's straightforward, so up to you whether to take a look.

perrydv:
@paciorek LGTM.  In the second part of the new test, you could use `expect_no_error`, but it might not matter.

Issue #1285, opened 2023-03-22
Unhelpful error message from indexing typo


perrydv:
We had a nimble-users report on this case:

In model code:
```
for (lv in 1:3){
<...>
    w_z[i:m,lv]~dmnorm(w_zmus[1:m], cholesky=cov_wz[1:m, 1:m], prec_param=0) 
}
```
where `m` is a constant and `i:m` was a typo for `1:m`.  The error given was "Error in i[iAns]:75 : NA/NaN argument".  This arose from `expandContextAndReplacements`, I think.  `i` is used as the index of other for loops in the model.  Can we trap this more helpfully, or possibly just use a try-catch around processing the whole line so we can at least point to where an error arose?

paciorek:
This simpler case seems fine. 

```
code <- nimbleCode({
    for(i in 1:3)
        y[i] ~ dnorm(0,1)
    for (lv in 1:3){
        w_z[i:m,lv]~dmnorm(w_zmus[1:m], cholesky=cov_wz[1:m, 1:m], prec_param=0) 
    }
})

m <- nimbleModel(code, constants = list(m=4))
Error in genVarInfo3() : 
  genVarInfo3: indexing error found for model variable(s): w_z ; please check that variables used for indexing are properly defined in the relevant for loop(s)
```

paciorek:
I am able to reproduce with the user's (slightly fixed up) code. 
I will track the issue down...

PR #1289, opened 2023-04-03
fix issue 1285

paciorek:
This catches a case (see issue #1285)  where in `genExpandedNodeAndParentNames3` the for loop code has variables not present in `BUGSdecl$replacementsEnv`.


paciorek:
The PR also has a couple very minor things that I had meant to commit directly to devel.

PR #1264, opened 2023-02-10
Fix MCMC conjugacy detection from differing slices of multivariate nodes

danielturek:
Only opening PR to see if testing currently passes.

PR #1290, opened 2023-04-04
conjugacy checking handles structureExprs with other nodes

danielturek:
Do not merge.

Opening PR to trigger tests, and see what cases this causes to fail.


danielturek:
@paciorek Requesting your feedback on this.

Basically: this tightens up the "conjugacy checking" to a higher level, to prevent false detections of "conjugacy" when it's not present.  This is done in a number of ways, basically being more precise for the logic handling nodes-contained-within-other-nodes, but two bottom line points:

First, the following situations (that you raised) no longer suggest conjugate relationships:
```
    code <- nimbleCode({
        mu[1:2] ~ dmnorm(mu0[1:2], Q[1:2,1:2])
        y[1:3] ~ dmnorm(mu[1:3], Q[1:3,1:3])
    })
```
(See the new tests added, "Conjugacy checking does not return conjugate for subsets (or supersets) of multivariate nodes" added in `test-mcmc.R` for the full suite of these situations)

Second, as a result of this tightening up of the checking, the system no longer suggests conjugacy for the case of "multivariate normal slices along different dimensions of an array", as in:
```
    code = nimbleCode({
        y[1:3] ~ dmnorm(mu[k, 1:3], pr[1:3,1:3])
        pr[1:3,1:3] ~ dwish(pr0[1:3,1:3], 5)
        for(i in 1:3)
            mu[1:3, i] ~ dmnorm(z[1:3], pr0[1:3, 1:3])
        k ~ dcat(p[1:3])
    })
```
The test for this ("Testing multivariate normal-Wishart dependency conjugacy detection with dynamic indexing" in `test-dynamicIndexing.R`) has been updated to no longer expect conjugacy for this case.  Importantly, note the fact that conjugacy *formerly* was detected was *not by design*, but only an artifact of the less-than-tight conjugacy checking logic.  So, although we no longer would detect conjugacy in this case, the system is not designed to do so, and **I think it's more important to "not assign conjugate samplers in cases without conjugacy", rather than to "accidentally assign conjugate samplers when it's in fact correct, but the system was not designed to correctly assess that case."**

I welcome your thoughts.

Also note that this PR supersedes PR #1264 ("Fix MCMC conjugacy detection from differing slices of multivariate nodes"), all changes from that PR are contained in this PR, so if/when we merge this PR, then PR #1264 should be closed (without merging).

This PR should, however, also be assessed bearing in mind your comments (from April 2nd) in the discussion of PR #1264.

danielturek:
Closing PR #1264

paciorek:
I will try to look at this in the next few days.

paciorek:
I can't say that I'm fully seeing how all the pieces fit together, but it looks like we're now covering the various cases that I can think of, so I'm happy to see this proceed.

I haven't fully worked through it, but I think that in all of the various cases we've considered here, we could arrange for our numerical calculation of `offset` and `coeff` to do the right thing in the context of setting up a conjugate sampler. This could also handle the case where we have `dnorm` and `dmnorm` for prior/likelihood or likelihood/prior. But that's all a bigger project, so removing conjugacy for these various cases now makes sense.

paciorek:
The error was:

```
> mcmcConf <- configureMCMC(model)
Error in getUnrolledIndicesList(node) : 
  something went wrong with Daniel's understanding of newNimbleModel
```

danielturek:
@paciorek This is addressed in PR #1314

PR #1293, opened 2023-04-07
addSampler now handles expandTarget and scalarComponents separately

danielturek:


danielturek:
Noting the testing failure is (previously witnessed) transient failure in Windows only:

```
Run RCMD INSTALL --install-tests packages\\nimble
* installing to library 'D:/a/_temp/Library'
ERROR: dependency 'igraph' is not available for package 'nimble'
* removing 'D:/a/_temp/Library/nimble'
Error: Process completed with exit code 1.
```

PR #1296, opened 2023-04-13
Initial skeleton for support of nimIntegrate in the DSL to work like …

paciorek:
…R's integrate


paciorek:
That was just a demo with Paul.

PR #1300, opened 2023-04-20
lift nimNewArr if LHS of <- is not just a name

### LINKED ISSUE #1234: `nimNumeric` in a model gives a segfault at compiled execution
It seems that `rep` works fine in a model, but the use of `numeric` or `nimNumeric` results in a segfault, when used in a model, at the time of compiled execution.

Minimal reproducible example:

```
library(nimble)

## this works:
code <- nimbleCode({ a[1:5] <- rep(0, 5) })
Rmodel <- nimbleModel(code)
Rmodel$a
Cmodel$calculate('a')
Cmodel$a

## this fails:
code <- nimbleCode({ a[1:5] <- numeric(5) })
Rmodel <- nimbleModel(code)
Rmodel$a
Cmodel$calculate('a')   ## segfault
```

perrydv:
Fixes #1234 

This fixes cases like

```
a[1:4] <- numeric(2)
```

by lifting the new vector (in general array) before assigning into the block of a. The issue is that the LHS is a block, not a full object. The fix is a small change in a heavily used piece of size processing, so the test suite will be important.

### DIFF
--- a/packages/nimble/R/genCpp_sizeProcessing.R
+++ b/packages/nimble/R/genCpp_sizeProcessing.R
@@ -967,15 +967,32 @@ sizeNimArrayGeneral <- function(code, symTab, typeEnv) {
                 asserts <- c(asserts, sizeInsertIntermediate(code, 1, symTab, typeEnv))
     }
 
-    if(!useNewMatrix) 
-        if(inherits(code$caller, 'exprClass'))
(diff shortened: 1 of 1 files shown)


perrydv:
@paciorek It looks like this is good to go. Let me know any code review.

PR #1301, opened 2023-04-20
check for risky aliasing with nimbleFunction and nimbleList objects

### LINKED ISSUE #1246: aliasing in assignment can occur for nimbleList objects
The assignment

```
some_nimList_obj$x <- some_nimList_obj$x[1:n] 
```

seems to encounter wrong compiled behavior that I suspect is due to an aliasing issue that hasn't been caught in code-generation.

perrydv:
Fixes #1246 

This fix puts a check in assignment operations for aliasing like `nf$a <- nf$a[1:3]`.

Because the changes will touch compiler processing of any assignment operation, the test suite will be important.  Let's see what happens.


paciorek:
@perrydv ok to merge?

perrydv:
@paciorek Yes but I forgot to add regression testing, so I just did so.  They are the last two tests in test-nimbleList.R.  Feel free to merge.

paciorek:
I think the Windows failure is just transient, but not entirely clear. We will run tests on Windows before release so will catch any OS-specific thing then.

PR #1314, opened 2023-05-12
conjugacy patch for single elements of multivariate deterministic nodes

danielturek:


danielturek:
@paciorek This fixes the issue in the `linear_predictors` demo of the `nimble-demos` repo, as described in #1290

PR #1318, opened 2023-05-17
Move compilation steps for nimbleCppADbaseClass

perrydv:
This PR takes nimbleCppADbaseClass out of libnimble.a and instead makes a session-specific .o file from it if needed, which is linked to on-the-fly compilations.  This seems to resolve the crashes we were seeing on Windows from `nimDerivs(model$calculate(nodes), ...)`.

### DIFF
--- a/packages/nimble/R/cppDefs_cppProject.R
+++ b/packages/nimble/R/cppDefs_cppProject.R
@@ -239,6 +239,19 @@ cppProjectClass <- setRefClass('cppProjectClass',
                                        writeDynamicRegistrationsDotCpp(cppName, dllName)
                                        nimbleUserNamespace$sessionSpecificDll <- compileStaticCode(dllName, cppName, showCompilerOutput)
                                    },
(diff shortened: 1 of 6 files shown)


paciorek:
@perrydv I will run some of the longer/omitted AD tests on Linux, unless this doesn't seem like the time for it yet.

perrydv:
@paciorek Sure. Do you know what's going on with the macOS failures? It looks like an CI infrastructure issue.

paciorek:
Ok, I've run basically all of our AD tests with the new configuration. Things generally look good:

- For the `ADfunctions`, I no longer need to turn off `clearCompiled`. Either with or without it turned off, no seg faults. I do still need to use `libnimble.a`.
- For the giant model in `ADmodels-bigmv.R`, which is equivalent to NCT issue 355, I do still get a seg fault.
- Otherwise, I am not seeing any seg faults with libnimble.a. Haven't tested with libnimble.so.

I've been meaning to look at the MacOS issue...

paciorek:
@perrydv I think we might as well merge this in. Sound ok?

Issue #1324, opened 2023-06-02
issue with error trapping deterministic data


paciorek:
Report from Daniel:

Example below works correctly in previous versions.
```
code <- nimbleCode({
    for(i in 1:N) {
        x[i,1] ~ dbern(0.5)
        for(j in 1:M) {
            y[i,j] ~ dnorm(x[i,j], 1)
        }
    }
})

N <- 2
M <- 3
constants <- list(N=N, M=M)
data <- list(
    x = matrix(c(0,  0, 0,
                 NA, 0, 0), byrow = TRUE, nrow = 2),
    y = array(0, c(N,M))
)
inits <- list(
    x = matrix(c(NA, NA, NA,
                 0,  NA, NA), byrow = TRUE, nrow = 2)
)

Rmodel <- nimbleModel(code, constants, data, inits)
Rmodel$calculate()   ## NA
Rmodel$x             ## not set
Rmodel$isDataEnv$x   ## not correct
```

danielturek:
Problem appears to be the result of [commit 46e273cc0034c9b6c4ac7cefe90c429bd8b9c7c4](https://github.com/nimble-dev/nimble/commit/46e273cc0034c9b6c4ac7cefe90c429bd8b9c7c4)

perrydv:
What's happening is that the `NA`s in `inits` are over-writing the data values for the RHSonly nodes, in `model$setData`.  This happens only in the case that the variable has a mix of RHSonly and stochastic nodes.  The problem arises because in that case `model$setData` copies only the non-data elements into the variable, and the RHSonly elements are now non-data (they were previously data), so they are copied from inits, and that results in the `NA`s being used when they shouldn't be.

A workaround is to replace the `NA`s in the `inits` with the same data values for the RHSonly portion of the variable.  In the above example, this change appears to fix it:
```
inits <- list(
  x = matrix(c(NA, 0, 0,
               0,  0, 0), byrow = TRUE, nrow = 2)
)
```

paciorek:
I'll say more in the design document on nimbleCoreTeam, but just a couple notes here that I am seeing things differently than @perrydv :

- non-data elements *are* set in `setData`; it's just that the RHSonly nodes are now being overwritten in `setInits` (because they are no longer flagged as `data`), which is invoked after `setData` in `newModel()`.
[...]

Issue #1326, opened 2023-06-11
bug in `setData` in 1.0.0


paciorek:
In trying to avoid flagging RHSonly and deterministic elements as 'data' in `setData`, I introduced a bug that affects models that have variables that are mix of actual data and either RHSonly or deterministic nodes. I think the most common cases this would affect are most models with state space/autoregressive structure on data, including occupancy and capture/recapture type models. 

The issue is here in `setData`:

```
  expandedNodeNames <- expandNodeNames(varName, returnScalarComponents = TRUE)
  determElements <- .self$isDeterm(expandedNodeNames, nodesAlreadyExpanded = TRUE)
  if(any(determElements))
    if(any(!is.na(varValue[which(determElements)])))
       stop("setData: '", varName, "' contains deterministic nodes. Deterministic nodes cannot be specified as 'data' or 'constants'.")
                                          
  .self[[varName]] <- varValue
  ## Values set as NA are not flagged as data nor are RHSonly elements.
  isDataVarValue <- !is.na(varValue)
  isDataVarValue[!.self$isStoch(expandedNodeNames, nodesAlreadyExpanded = TRUE)] <- FALSE
```

The problem is that `expandedNodeNames` returns node elements in a model, but `varValue` is a rectangular array. And there may be elements of `varValue` that don't correspond to node elements (e.g., `y[1]` if the model code only defines `y[2],...`, with `y[1]` never appearing on either LHS or RHS). In that case the correspondence between the elements of `expandedNodeNames` and `varValue` is not correct. 

I think problems could occur under the following conditions:
1) Not all elements of a variable are part of nodes, **and**
2) There are any RHSonly or deterministic elements in a variable that has data elements (since this would be when `stop()` is invoked or `isDataVarValue` is modified).

I think occupancy and capture-recapture models would be affected in cases where at least one individual is not observed for the entire time period, which of course would be most situations.

Without setting up a whole "accessor" system along the lines of what we do with `values<-`, I'm not seeing an easy way to correctly access values in `varValues` or `isDataVarValue`, so for now my thought is to revert to the pre-1.0.0 behavior and accept that RHSonly and deterministic nodes may be flagged as data by users.

paciorek:
Actually I suppose we could take `expandedNodeNames` which might look like `c('y[3,1]','y[4,1]')` and do some string/code modification to be able to use `eval` to do `isDataVarValue[3,1] <- FALSE`, etc.

PR #1327, opened 2023-06-12
Fix setData

paciorek:
This fixes issues #1324 and #1326 by:
 - only overwriting values set by `setData` with `setInits` if the value from `setData` is NA.
 - warning users if they try to overwrite a value set by `setData` in `setInits` and the `setInits` value is different from the `setData` value and not NA.
 - adding an `includeRHSonly` flag to `isDeterm` with default `FALSE` and then updating some uses of `isDeterm` in our codebase to set `includeRHSonly` to TRUE.
 - fixing #1326 by indexing into `varValue` (when checking for deterministic data) and `isDataVarValue` (when setting data flags) using `eval` with the scalar node names (similar to how `isDataFromGraphID` works). This avoids setting the data flag to TRUE for deterministic and RHSonly elements of a variable that has data elements. Note that positions in `isDataEnv$x` can still be TRUE for positions that do not correspond to any nodes (e.g., if `x[1]` is not defined at all in the model, but the user provides a non-NA value for `x[1]` in `setData`).

Note that I have not yet added various tests, in particular the CR-type example where there are undefined nodes in a variable, such as `y[i, 3:6]` defined but `y[i, 1:2]` undefined. I will also review the changes again carefully.

### DIFF
--- a/packages/nimble/R/BNP_samplers.R
+++ b/packages/nimble/R/BNP_samplers.R
@@ -1369,7 +1369,7 @@ sampler_CRP <- nimbleFunction(
     for(varIdx in seq_along(clusterVarInfo$clusterVars)) {
         if(length(unique(clusterVarInfo$clusterNodes[[varIdx]])) != length(clusterVarInfo$clusterNodes[[varIdx]]))
             stop("sampler_CRP: cluster parameters in different clusters must be part of conditionally independent nodes.")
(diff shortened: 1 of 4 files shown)


paciorek:
Per discussion in NCT issue 457, this approach won't work because it prevents use of `setInits` to generally change values. 
See next PR that avoids the RHSonly as data issue and simply addresses determ nodes as data.

PR #1328, opened 2023-06-13
Fix `setData`

paciorek:
This fixes issues #1324 and #1326 by:
- adding an includeRHSonly flag to isDeterm with default FALSE and then updating some uses of isDeterm in our codebase to set includeRHSonly to TRUE.
- not setting determ nodes as data and warning user of that situation
- fixing issue #1326 by (1) indexing into `varValue` (when checking for deterministic data) using eval with the scalar node names (similar to how `isDataFromGraphID` works) and (2) not trying to avoid RHSonly nodes as data (i.e., reverting to pre-1.0.0 behavior).


paciorek:
Merged manually.

PR #1329, opened 2023-06-25
added maxDimCovHistory control option for RW_block sampler

danielturek:
The title says it all.


danielturek:
I'll leave this to (hopefully) get merged at some point in the future.

paciorek:
In email I suggested we might want to instead have the option be `MCMCsaveHistoryCov`, allowing a user to separately say whether they want the covariance saved in addition to the other history stuff. 

Changing the max dimension seems a bit more round-about to me than just having the user directly specify whether they want the covariance saved? 

That said, as I think about it more, a user could well want to save the covariance only for some of the nodes, so keeping the option as something that relates to dimension makes sense. Otherwise they wouldn't be able to say they want covariance saved only for the lower-dimensional nodes...

danielturek:
@paciorek I just saw your comment here.  Thanks for your thoughts on this.

paciorek:
Sorry to confuse things - there's (1) the question of whether the user should set the max dimension or simply say whether the covariances should be saved and (2) the question of system option vs. sampler control option. I hadn't noticed that the proposal was to use a sampler control option.

On (1), I've come around to the idea of the user controlling the max dimension.

On (2), as far as system option vs. sampler control option, I'm fine with having it as a sampler option.

paciorek:
Hmm, it looks like you changed line 1556 to 

```
scale <- extractControlElement(control, 'scale',               1)
```

but then `originalScale` is not defined. I'm changing that line (on devel) to be

```
scaleOriginal <-  extractControlElement(control, 'scale',               1)
```

but let me know if I'm missing something.

PR #1339, opened 2023-08-12
fixed multiple warnings of MCMC unsampled nodes

danielturek:
This fixes the problem of outputting multiple (2) warnings about unsampled nodes, which would occur when using:
```
buildMCMC(model, nodes = only_some_of_model_nodes)
```


danielturek:
Wow, I'm totally at a loss for how this passed testing.

Issue #1345, opened 2023-10-08
`registerDistributions` type-checking does not allow default values for `x`


danielturek:
When defining the density function of a custom distribution, the type-checking that takes place in `registerDistributions` does not allow a default value for `x`.  During the type-checking, we require an `identical` to pass for the return type defined by  the density function (in the dist <- nimbleFunction(`x = ...`) of the argument list), and the `returnType` statement of the generation function.  If you provide a default value of `x` in the density function, this `identical` check fails.  Example below:

```
library(nimble)

ddist <- nimbleFunction(
    run = function(x = double(0, default = 0), log = integer(default = 1)) {
        returnType(double())
        return(0)
    }
)

rdist <- nimbleFunction(
    run = function(n = integer()) {
        returnType(double())
        return(0)
    }
)

registerDistributions('ddist')

## Error in FUN(X[[i]], ...) : 
##   checkDistributionFunctions: random generation function `rdist` is missing `returnType` or `returnType` 
##   does not match the type of the `x` argument to the corresponding density function.
```

paciorek:
I think I would argue that instead we should not allow default arguments for user-defined distributions (by analogy with standard distributions).

@danielturek what do you think about me adding a check that errors out if there is a default?

paciorek:
Fixed by PR #1371

PR #1346, opened 2023-10-08
registerDistributions type-checking allows default value for x arg

danielturek:
Modified the type-checking in `registerDistributions`, to allow a default value for the `x` argument of user-defined distributions.  For example, the following is now ok:

```
ddist <- nimbleFunction(
    run = function(x = double(default = 0), log = integer(default = 1)) {
        returnType(double())
        return(0)
    }
)
```

This should never affect model calculations, but does result in a user-defined density function (in the R environment) which has a default value for the `x` argument.


danielturek:
@paciorek @perrydv I think this is a pretty reasonable change, and I'm ok with this getting merged, pending any thoughts from either of you.

paciorek:
Closing this without merging given discussion in issue #1345

PR #1349, opened 2023-10-12
inline many core functions

perrydv:
This PR changes the following C++ functions in Utils.h to be inlined. That means they now use the C++ keyword `inline` (previously they called a library function that on-the-fly compilation would link to) and thus the function contents get placed directly where called and are compiled there. This reduces function call overhead and presumably allows compiler optimizations. A user shared that the new `pow_int` in version 1.0.0 introduced a surprisingly large slowdown in performance compared to previous versions, and this was traced to `pow` becoming `pow_int` in user-defined code. 

This PR inlines the following (with notes on performance [before --> after] on a mac for 10 million calls, timed in seconds):

- `pow_int` (0.201 --> 2e-6)
- `ilogit`.     (0.073 --> 2e-6)
- `logit`       (ditto)
- `cloglog`. (0.143 --> 2e-6)
- `icloglog` (0.146 --> 2e-6)
- `iprobit` (little difference because the cost is in pnorm)
- `probit`. (little difference because the cost is in qnorm)
- `nimEquals`
- `nimbleIfElse`
- `lfactorial` (little difference because the cost is in lgamma)
- `factorial`. (little difference because the cost is in gamma)
- `nimRound`
- `pairmax` in overloaded (double, double) case
- `pairmin` (ditto)
- `nimStep`
- `cube`
- `inprod` in overloaded (double, double) case (should be rare)

Cases without notes went from about 0.03-->2e-6, and that 0.03 might be just the baseline cost of packing up function calls.

All of these seem fast before or after the changes (these are cumulative times for *10 million* calls). Yet, the `pow_int` was the source of a serious slowdown, so perhaps the overhead and costs compound in more complex compiled situations using lots of memory etc. The benchmarks I ran were isolated and had nothing else going on.

### DIFF
--- a/packages/nimble/inst/CppCode/Utils.cpp
+++ b/packages/nimble/inst/CppCode/Utils.cpp
@@ -67,29 +67,29 @@ void nimStop() {NIMERROR("");}
 
 bool nimNot(bool x) {return(!x);}
 
-double pow_int(double a, double b) {
-  return pow(a, fround(b, 0.));
-}
+// double pow_int(double a, double b) {
+//   return pow(a, fround(b, 0.));
+// }
 
-double ilogit(double x) {return(1./(1. + exp(-x)));}
+// double ilogit(double x) {return(1./(1. + exp(-x)));}
 
-double icloglog(double x) {return(1.-exp(-exp(x)));}
+// double icloglog(double x) {return(1.-exp(-exp(x)));}
 
-double iprobit(double x) {return(pnorm(x, 0., 1., 1, 0));}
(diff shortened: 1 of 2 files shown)


danielturek:
@perrydv The performance results are remarkable.

Who ever knew the power of the cpp keywork `inline` ?

perrydv:
Thanks @danielturek . Well Eigen for example makes heavy use of inlining. It will be interesting to see what kind of net performance gains these give, whether they are swamped by other costs.

PR #1376, opened 2023-12-21
additional catch for vector inits lists of lists

danielturek:
Error fix for inits provided as a list of lists, for non-scalar inits lists.


danielturek:
This PR adds an additional check, when `inits` is provided as a list-of-initial-values-lists.

With this change, in the `assignDimensions` method, we only use the first list of initial values for determining model dimensions.

Issue #1356, opened 2023-11-21
inconsistent/confusing warning about nimbleFunction not yet created


paciorek:
When using `nimOptim` or the (being drafted) `nimIntegrate`, we get the warning `variable '<some_nimbleFunction>' has not been created yet` when the output of `optim` is assigned to a variable. When instead the output is directly returned, the warning is not given. It looks like this relates to how `typeEnv$.AllowUnknowns` is set in `sizeAssign` vs. `sizeReturn`. In `sizeAssign` we set it to FALSE before calling `asserts <- recurseSetSizes(code, symTab, typeEnv, useArgs = c(FALSE,  TRUE))`, but we don't do that before calling `recurseSetSizes` in `sizeReturn`.

@perrydv do you have any idea why we handle `.AllowUnknowns` differently in these two situations?

Here's an example. 

```
    objectiveFunction_inner <- nimbleFunction(
         run = function(par = double(1), power = double(0), scaling = double(0)) {
             return(sum(par) * exp(-sum(par ^ power) / scaling))
             returnType(double(0))
         }
    )
    objectiveFunction <- nimbleFunction(
        run = function(par = double(1)) {
            return(objectiveFunction_inner(par, 2, 3))
             returnType(double(0))
         }
     )
     optimizer <- nimbleFunction(
         run = function(method = character(0), fnscale = double(0)) {
             control <- optimDefaultControl()
             control$fnscale <- fnscale
             par <- c(0.1, -0.1)
             return(optim(par, objectiveFunction, method = method, control = control))   # NO WARNING
            ## If the following two lines replace the previous line, there is a WARNING.
             ## out <- optim(par, objectiveFunction, method = method, control = control)) 
             ## return(out)
             returnType(optimResultNimbleList())
         }
     )
     cOptimizer <- compileNimble(optimizer)   
```

paciorek:
This duplicates (accidentally) #1295 but has more detail so I've closed 1295.

paciorek:
My thinking at the moment is there are two issues here.
1. We generally **should** warn if a variable (like `b` above) is missing in `return()` as we do with `assign`.
1a. I think there is an argument for erroring out if missing `b` rather than just warning, but I suspect Perry thought this through and decided on warning for good reason.
2. For `nimOptim` (and `nimIntegrate`) we should handle function args differently such that we look to see if they exist rather than standard processing of them as names. I think we may be able to modify `sizeOptim`  to deal with this.

[...]

Issue #1365, opened 2023-12-06
discuss: should missing variables in a nf trigger warning or error?


paciorek:
In a case like this, we warn about missing `b` and C++ compilation fails. It's not clear to me why we don't error trap this in our compilation process. I guess the idea is that we want to allow uncompiled execution with `b` being found by scoping? I'm not sure that feels compelling to me if that is the reason given we generally want uncompiled and compiled execution to match.

Note that if this were `c <- foo(a)` with missing `foo` we do warn about that via a separate mechanism.

As a side note, 

```
test1 <- nimbleFunction(
    run = function() {
        a  <- 7
        c <- a+b  # warning about 'b' not yet created
        return(c)
        returnType(double(0))
    })
compileNimble(test1)

Compiling
  [Note] This may take a minute.
  [Note] Use 'showCompilerOutput = TRUE' to see C++ compilation details.
Error: Failed to create the shared library. Run 'printErrors()' to see the compilation errors.
In addition: Warning message:
variable 'b' has not been created yet. 
```

paciorek:
I'm changing this to an error.

PR #1378, opened 2023-12-21
fix issue 1356 by not warning of missing nf via `sizeOptim`

paciorek:
also adding warning of missing vars in `sizeReturn`.


paciorek:
Fixes issue #1356

paciorek:
This is getting hung up on the test-ADlaplace issue but o.w. seems ok, so I will merge once @perrydv takes a look.

paciorek:
Actually there is also a bit of tangle in terms of tests added for issue #1365 / PR #1379 causing the Linux test failure seen here. 
Once Perry takes a look I will probably just merge this in and deal with any testing fallout directly on devel.

perrydv:
I've tried changing to fix to the second aspect of this, the spurious warning for the `fn` or `gr` argument to `nimOptim` if it is an RCfunction.

PR #1380, opened 2023-12-21
warn if indexing in model code uses vars from user environment rather than `constants`

paciorek:
Fixes NCT issue 468.


paciorek:
Just the ADlaplace issue, so merging.

PR #1382, opened 2023-12-22
detect invalid variable names in nfcode

### LINKED ISSUE #1335: Invalid parameter name 'int'
Dear all,

Similarly to #1219, using `int` as a parameter name (after intercept) leads to a compilation error, which is not straightforward to track back to the source of the problem.

``` r
library(nimble)
#> nimble version 0.13.0 is loaded.
#> For more information on NIMBLE and a User Manual,
#> please visit https://R-nimble.org.
#> 
#> Note for advanced users who have written their own MCMC samplers:
#>   As of version 0.13.0, NIMBLE's protocol for handling posterior
#>   predictive nodes has changed in a way that could affect user-defined
#>   samplers in some situations. Please see Section 15.5.1 of the User Manual.
#> 
#> Attaching package: 'nimble'
#> The following object is masked from 'package:stats':
#> 
#>     simulate
code <- nimbleCode({
  int ~ dexp(1) 
})
rModel <- nimbleModel(code)
#> Defining model
#> Building model
#> Running calculate on model
#>   [Note] Any error reports that follow may simply reflect missing values in model variables.
#> Checking model sizes and dimensions
#>   [Note] This model is not fully initialized. This is not an error.
#>          To see which variables are not initialized, use model$initializeInfo().
#>          For more information on model initialization, see help(modelInitialization).
cModel <- compileNimble(rModel)
#> Compiling
#>   [Note] This may take a minute.
#>   [Note] Use 'showCompilerOutput = TRUE' to see C++ compilation details.
#> Error: Failed to create the shared library. Run 'printErrors()' to see the compilation errors.
printErrors()
#> using C++ compiler: ‘g++ (Ubuntu 11.3.0-1ubuntu1~22.04.1) 11.3.0’
#> using C++11
#> In file included from P_1_code_MID_1.cpp:7:
#> P_1_code_MID_1.h:24:23: error: expected unqualified-id before ‘int’
#>    24 |   NimArr<1, double> * int;
#>       |                       ^~~
#> P_1_code_MID_1.cpp: In constructor ‘code_MID_1::code_MID_1()’:
#> P_1_code_MID_1.cpp:48:22: error: expected primary-expression before ‘int’
#>    48 | namedObjects["int"]=&int;
#>       |                      ^~~
#> P_1_code_MID_1.cpp: In member function ‘void code_MID_1::pointAtAll(code_MID_1_MV_UID_1*, int)’:
#> P_1_code_MID_1.cpp:52:4: error: expected unqualified-id before ‘=’ token
#>    52 | int=&(values__->int_Vec[i__]);
#>       |    ^
#> make: *** [/usr/lib/R/etc/Makeconf:200: P_1_code_MID_1.o] Error 1
```

<sup>Created on 2023-07-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Perhaps some detection of invalid node names with a more informative error message would be useful?

Best.
ƒacu.-

paciorek:
Fixes #1335 , a longstanding issue that certain variable names conflict with C++ keywords but this is not error-trapped and C++ compilation fails.

### DIFF
--- a/packages/nimble/R/BUGS_modelDef.R
+++ b/packages/nimble/R/BUGS_modelDef.R
@@ -20,6 +20,10 @@ graphNode <- setRefClass(
     )
 )
 
+reservedKeywords <- c("Values", "NimArr")  # These prevent compilation. Not clear without further investigation why othe class names in `include/nimble` do not.
+## Add C++ keywords (from https://en.cppreference.com/w/cpp/keyword).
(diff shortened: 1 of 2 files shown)


paciorek:
@perrydv would be good to have you give this a quick look before I merge.

perrydv:
@paciorek I think this is ok. I can't think of a case where it would go haywire and do something unexpected. We could give it a nimbleOption toggle to be safe, but up to you.

Issue #1366, opened 2023-12-07
eigenization of besselK - is `toEigenize` specified correctly?


paciorek:
In `sizeRecyclingRuleBesselK`, we have `code$toEigenize <- TRUE` instead of assigning `"yes"`. Check this as it seems inconsistent at best and possibly a bug.

paciorek:
BesselK calcs seem fine, including with recycling. Seems strange that this would all work when `code$toEigenize <- TRUE`.

I haven't delved more deeply, but I'm changing to be "yes".

PR #1385, opened 2023-12-22
fix besselK to have standard `toEigenize`

paciorek:
@perrydv wanted to look at this, in part to figure out why it hasn't caused a problem already.


paciorek:
This fixes issue #1366

perrydv:
@paciorek LGTM but one note on merging everything together. The changes in #1396 will, I think affect the besselK recycling rule C++ implementation. Specifically I would check what happens if you give a matrix argument. I think it should return a vector using all the elements, but it is worth checking or adding a test if necessary.

PR #1389, opened 2024-01-10
Support stochastic indexing in AD. Allow dists and functions in a model that uses AD to lack AD support.

perrydv:
This branch is not completely finished, but I want to initiate general testing.

This includes two changes:

1. New atomics and associated code generation to support stochastic indexing within AD.
2. Allow distributions and functions (whether user-defined or built-in) that lack AD support to be used and compiled in a model that has `buildDerivs=TRUE`. This includes:

    - At `nimbleModel`, emit a warning message.
    - At `compileNimble`, no message is emitted.
    - At C++ run-time, if the dist or function lacking support is called, it is a (hard stop) run-time error and a lengthy message is emitted. This includes information about how to turn it into just a warning or disable the error-trapping entirely.

Testing for these features is partially written and is (current commented-out) at the bottom of test-ADbasics and test-ADfunctions. At the launch of this PR, I want to see if all other testing passes. (At the moment we have one stable testing problem that fails reliably.)

`dcat` has not been updated but will be, with additional testing.


perrydv:
The test failures were actually because I tweaked the testing infrastructure (`make_op_param`) in a way that accidentally broke other tests. I fixed that and also merged in the "--preclean" (fix-1368). Let's see what happens next with testing.

paciorek:
Ok, here are various comments (mostly minor) about messaging/error-trapping around new AD functionality. I've separated out some issues as separate GH issues when they might involve more complicated discussion.

 - [] See NCT issue 493 for a (new) problem with use of `dcat`. Because of that I'm not able to check more complicated use of `dcat` as prior for a discrete state.
 - See NCT issue 494 about a confusing message about use of `pow` with AD in standard models.
 - We still warn that "model cannot be compiled" in various cases that are now fine, such as use of `dconstraint`, `T()`, etc. in places in model where derivs are not taken.
 - Still have warning that derivs cannot be build for models that have dynamic indexing.
 - In the C++ run-time error messaging about having a model node included in derivatives that don't support them, I'd like to have the messaging better conform to our messaging system (I do recognize that run-time messages are somewhat different). I can do this.
 - I'd like to update some of the messaging to use `messageIfVerbose`. I can do this.
 - Some of the new tests have `expect_` outside of `test_that`. That can cause problems in CI. As suggested in email, I am hoping we can use `expect_message` in these cases inside of `test_that`.
 - I think we should change the syntax related to "does not have support for derivatives" from "[Warning]" to "[Note]". Also there is an extra space in this message. In many cases that a user writes a model like this, there will be nothing wrong.
 - We need `temporarilyAssignInGlobalEnv` for user-defined distributions in the new testing.

@perrydv I'm happy to do much of the updating of the tests and messaging, but want to give you a chance to first see my comments and address things that you are best equipped to handle.

paciorek:
Various testing and updated errors/warnings based on discussion with @perrydv . I will also add some overall testing via tests in nimbleHMC for CAR model and a stochastic index model with dcat.

PR #1390, opened 2024-01-10
enable AD for CAR models

paciorek:
This sets up templated versions of `dcar_normal` and `dcar_proper` for AD and enables `buildDerivs` for various auxiliary nimbleFunctions functions in `CAR.R`. In the latter case, this requires various uses of `ADbreak`, `ignore` and setting up index variables as integers.

Testing now includes tests that AD-enabled CAR models will build and basic (non-HMC) MCMC will run. Substantive HMC testing could be done in `nimbleHMC`.

@perrydv a glance over the AD-enabled code would be helpful.

@danielturek just an FYI; I'm sure you'll be happy to see this.

### DIFF
--- a/packages/nimble/LICENSE
+++ b/packages/nimble/LICENSE
@@ -1,3 +1,3 @@
-YEAR: 2023
+YEAR: 2024
 COPYRIGHT HOLDER: Perry de Valpine, Christopher Paciorek, Daniel Turek, Clifford Anderson-Bergman, Nick Michaud, Fritz Obermeyer, Claudia Wehrhahn Cortes, Abel Rodriguez, Sally Paganin, Wei Zhang, Duncan Temple Lang
 ORGANIZATION: University of California
--- a/packages/nimble/R/CAR.R
+++ b/packages/nimble/R/CAR.R
@@ -23,7 +23,7 @@ CAR_convertWeightMatrix <- function(weightMatrix) {
 #' 
 #' Two alternate representations are handled:
 #'
-#' A single matrix argument will be interpreted as a matrix of symmetric unnormalized weights;
+#' A single matrix argument will be interpreted as a matrix of symmetric unnormalized weights.
 #'
 #' Two lists will be interpreted as (the first) a list of numeric vectors
 #' specifying the adjacency (neighboring) indices of each CAR process component,
--- a/packages/nimble/R/CAR.R
+++ b/packages/nimble/R/CAR.R
@@ -234,8 +234,8 @@ CAR_calcNumIslands <- nimbleFunction(
     run = function(adj_in = double(1), num_in = double(1)) {
         N <- dim(num_in)[1]
         L <- dim(adj_in)[1]
-        adj <- nimInteger(L)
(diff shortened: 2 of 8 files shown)


paciorek:
I ran the disease mapping example from our examples with HMC and results compared favorably (in terms of mixing) and were numerically similar to our default MCMC.

Not sure if we want an actual nimbleHMC test that assesses validity of the HMC results for a CAR model.

danielturek:
@paciorek I'm very happy to see this.  Thank you.

I'm pushing some (very) minor changes to style/comments.

One minor question, should the line (# 293 of `CAR.R`):
```
 M <- rep(-1234, length = N)
```
instead be:
```
 M <- rep(-1234L, length = N)
```
?  It seems that without this, some elements of M (corresponding to zero neighbor islands) might persist as double value -1234, and have derivatives tracked.  But perhaps this wouldn't cause any downstream problems, I'm not sure.

perrydv:
This is minor issue @paciorek but I noticed that in RCfunction_core.R, in `fxnsNotAllowedInAD`, the `dcar_normal` and `dcar_proper` entries do not appear to be correctly entered. They have an extra "d". They were correctly entered in `distsNotAllowedInAD`. This might mean that they were error-trapped in one place but not another. I am not tracking it down at the moment. And it might become moot from this branch.

paciorek:
Ok, I've fixed the issue Perry raised and set `M` in `ignore` in terms of Daniel's comment.

@perrydv given you responded does means you looked over the changes and things seem fine to you such that I can merge this in?

paciorek:
Some tests were failing because of typo in `numNumeric`.

perrydv:
@paciorek Thanks for this. My previous comment was actually something I just noticed while working on something else, so no I had not done a thorough code review. Now I've looked through it and here are my notes:

In `CAR_calcNumIslands`

- Reminder that you can use `nimInteger(..., init=FALSE)` if you are going to populate values in a for loop anyway. Someone somewhere will celebrate those microseconds ;-)
- You have `nimInteger(N)` and `rep(0L, N)` a few lines apart. They should be equivalent. It doesn't matter. If you want a more consistent style you could go with the first (with init=FALSE), or either.
- I think you should be able to do `indToVisit <- nimInteger(nNeighbors, value =  adj[adjStartInd:(adjStartInd+nNeighbors-1)])`, unless the AD processing gets confused about excluding all of that line from anything to do with AD.

In `CAR_calcCmatrix`

- You also have the option of `init=FALSE` in `nimMatrix`, although for `Cmatrix` you might defensively really want the zeros because not every value will be filled by the following for loops.

In the C++

- It looks like instead of `Type(R_NaN)` we are using `CppAD::numeric_limits<Type>::quiet_NaN()`
- I think `if(Type(count) != Type(L))` should be `if(count != L)`. It would take different syntax to have the conditional recorded on the AD tape, but if I follow these are fixed values that can be baked into the tape (and in this case would bake in a NaN and it would all be disaster anyway).

paciorek:
The failure seems to be unrelated to this PR. I will look into it on devel.

Issue #1372, opened 2023-12-18
can't compile Laplace-based code on Apple Silicon machine under devel


paciorek:
I'm getting compilation issues on my new M2 Mac when trying to run `test-ADlaplace.R`. Strangely there are warnings about the bitwise `|` that I thought we dealt with. 
Also, despite everything below seeming to be just warnings, the compilation fails.

Using 1.0.1 things seem fine, as does using devel on an Intel-based Mac. 

Failure occurs on the first test case:

```
 m <- nimbleModel(
    nimbleCode({
      y ~ dnorm(a, sd = 2)
      a ~ dnorm(mu, sd = 3)
      mu ~ dnorm(0, sd = 5)
    }), data = list(y = 4), inits = list(a = -1, mu = 0),
    buildDerivs = TRUE
  )

  mLaplace <- buildLaplace(model = m)
  mLaplaceNoSplit <- buildLaplace(model = m, control = list(split = FALSE))
  cm <- compileNimble(m)
  cL <- compileNimble(mLaplace, mLaplaceNoSplit, project = m)
```

```
 cL <- compileNimble(mLaplace, mLaplaceNoSplit, project = m)
Compiling
  [Note] This may take a minute.
  [Note] On some systems there may be some compiler warnings that can be safely ignored.
Error: Failed to create the shared library.
clang++ -arch arm64 -std=gnu++17 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -DR_NO_REMAP   -DEIGEN_MPL2_ONLY=1 -I"/Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/library/nimble/include" -Wno-misleading-indentation -Wno-ignored-attributes -Wno-deprecated-declarations  -I/opt/R/arm64/include    -fPIC  -falign-functions=64 -Wall -g -O2  -c P_6_AGHQuad.cpp -o P_6_AGHQuad.o
using C++ compiler: ‘Apple clang version 15.0.0 (clang-1500.0.40.1)’
using SDK: ‘MacOSX14.0.sdk’
P_6_AGHQuad.cpp:1588:5: warning: variable 'Interm_586' set but not used [-Wunused-but-set-variable]
int Interm_586;
    ^
P_6_AGHQuad.cpp:1647:5: warning: variable 'Interm_592' set but not used [-Wunused-but-set-variable]
int Interm_592;
    ^
[...]

PR #1392, opened 2024-01-11
change `|` to `||` to avoid error on M2 Mac

paciorek:
This fixes issue #1372


paciorek:
Errors are unrelated.

PR #1396, opened 2024-01-16
Fix the recycling-rule bug in specific cases of pow_int

### LINKED ISSUE #1395: C++ problem with compound calls to recycling-rule code using function arguments
This issue came from a user on nimble-users on 1/11/24.

```
testNimble <- nimbleFunction(
  run = function(r = double(2)) {
    returnType(double(2))
    return( r + r^2 )
  })
cTestNimble <- compileNimble(testNimble)

testR <- function(r) { r + r^2 }

testInput <- matrix(c(0  ,0.2,0.4,0.6,0.8,1  ), ncol=2)

print("expected output:"); print(testR(testInput))
print("uncompile nimble output:"); print(testNimble(testInput))
print("compiled nimble output:"); print(cTestNimble(testInput))
```

The second column of matrix output is wrong. The problem occurs only when using a recycling-rule function (the fairly new `pow_int` for `pow_int(r, 2)` in this case) with a matrix input, compounded with another operation (`r+` in this case), when both operations have as their arguments variables which are function arguments (`r` and `r` in this example).

The following two cases do not display this problem:
```
testNimble <- nimbleFunction(
  run = function(r = double(2)) {
    returnType(double(2))
    a <- 2 # by assigning 2 to a variable, pow instead of pow_int is used on the next line; no bug
    return( r + r^a )
  })
cTestNimble <- compileNimble(testNimble)

testNimble <- nimbleFunction(
  run = function(r = double(2)) {
    returnType(double(2))
    t <- r # by making a local copy, different Eigen handling is invoked; no bug
    ans <- t+ t^2
    return( ans )
  })
cTestNimble <- compileNimble(testNimble)
```

perrydv:
Fixes #1395

### DIFF
--- a/packages/nimble/R/cppDefs_cppProject.R
+++ b/packages/nimble/R/cppDefs_cppProject.R
@@ -204,7 +204,9 @@ cppProjectClass <- setRefClass('cppProjectClass',
                                        ssDllName <- normalizePath(file.path(dirName, paste0(dllName, .Platform$dynlib.ext)), winslash = "\\", mustWork=FALSE)
                                        ssdSHLIBcmd <- normalizePath(file.path(R.home('bin'), 'R'),
                                                                           winslash = "\\", mustWork=FALSE)
-                                       ssdSHLIBargs <- paste('CMD SHLIB', cppName, '-o', basename(ssDllName))
+                                       ssdSHLIBargs <- paste('CMD SHLIB',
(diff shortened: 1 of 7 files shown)


perrydv:
@paciorek I think this is good to merge of have you do any code review or checking of test coverage.

perrydv:
Sorry, typo, "of" --> "or". Basically I'm fine with it but I'm not sure if you've given it a code review or feel that would be useful.

PR #1401, opened 2024-01-19
stop instead of simply printing on run-time size errors.

perrydv:
Do not merge.

This branch contains small changes that will touch lots of tests.

1. Run-time size-checking will use `nimStop` instead of `nimPrint` by default. We may have tests for these messages and if so they may fail and need updating.
2. In cases of "vector + matrix", the size of the result was being set to the size of the matrix. Now it will be set to (size of vector, 1). Same for other binary component-wise operators. The previous change (error on size mismatch) should prevent this from mattering as much anyway. Again, this might break some tests and the easiest way to find out is this PR.


perrydv:
Looks like the test failure is in test-errors, and that makes sense since I changed the error-handling behavior. I will look at that tomorrow.

perrydv:
@paciorek I think this is good to merge. I added a very conservative fallback option on the vec+mat size issue to revert to old behavior if necessary. Tests passed before and they should again...

PR #1407, opened 2024-01-26
Supports `addSampler(..., allowData = TRUE)`

danielturek:
Adds an argument `allowData` to `addSampler` method.  A few details:

`configureMCMC` entry point will *never* assign samplers to data, even if data nodes are included in the `nodes` argument.  There is no support for this; there is no `allowData` argument to `configureMCMC`.

Samplers *can* be added to data nodes, using:
```
conf$addSampler(target, ..., allowData = TRUE)
```

When `allowData = FALSE`, data nodes are filtered out of the `target` nodes.
When `allowData = TRUE`, samplers can be added to data nodes.

Suport is also provided for using default sampler assignment logic to [any set of nodes, which can include] data nodes.  This is accessed using:

```
conf$addSampler(target, default = TRUE, allowData = TRUE)
```

Note that `posterior_predictive` samplers will never be assigned to data nodes, since they are not posterior predictive.  A small case could be made for supporting this, but if a user is intentionally assigning samplers to data nodes, then the presumption is they are advanced enough to consider what sampler they would like to add, and if not, they can accept the (valid) default non-PP sampler assignment

Testing added, NEWS updated.


paciorek:
CAR failure was a hiccup of mine. I've fixed on devel and am rerunning that test set.

danielturek:
@paciorek whenever you push the CAR (testing) fix to `devel`, could you please merge that fix also into this branch (`samplers_to_data`) and push the changes?  I'll be away from a computer starting shortly until much later today.

paciorek:
Having trouble getting it to rerun failed test with the fixes on devel. Just going to merge as things seem fine.

PR #1422, opened 2024-03-04
modelValuesAccessorVector setup template matches modelValues

danielturek:
@perrydv as best I see this fixes a small typo in the `modelValuesAccessorVector_setupCodeTemplate`, which defines the setup code for copy operations to/from modelValues objects.  I believe this was a harmless partial matching, which as-written matches the correct modelValues object because of the same prefix "model".  This PR corrects the partial match typo, and also enforces exact matching by now using `LIST[["VAR"]]` syntax.

That said, @perrydv you should definitely look at this, and confirm this makes sense.


danielturek:
@perrydv @paciorek Pending testing, I think this is ok for review, and merging into devel.

paciorek:
@perrydv just wanted to check with you on this before merging.

perrydv:
Yes, LGTM. Thanks @danielturek for tracking this down.

PR #1430, opened 2024-03-23
protect gradient methods in nimOptim C++ from when the user-provided fn or gr modifies the input vector.

perrydv:
I think I fixed this once before, but somehow evidently it didn't make it into devel. I'm not sure what happened.

The issue is that a nimOptim call provides an objective function and, optionally, a gradient function. In compiled code, these will take the parameter vector by reference, and they may modify it. If a parameter transformation is in use, they might even shorten the parameter vector.  If the calling code assumes the parameter vector was unchanged, a bug can result. Minor changes are made in this PR to protect against that scenario.

Specifically, I believe this was causing a crash in test-ADlaplace, in the test involving a length-changing transformation (from a Dirichlet). There seems to be some kind of other intermittent issue coming up in test-ADlaplace, and I am not sure if they are related.


paciorek:
This also deals with PR #1432

PR #1432, opened 2024-04-02
Check on VECTOR_ELT intermittent errors

perrydv:
Do not merge.

This has some trial pieces to investigate the intermittent "`VECTOR_ELT`" error that appears intermittently and perhaps entirely (?) in testing.

Note two things to polish if this appears to fix the issue:

- The `R_PreserveObject` calls need to be matched by `R_ReleaseObject`, which would take setting up destructor/finalized behavior.
- The testing has been hot-wired to run only ADlaplace and nimbleLists because those are the usual ones where we see the issue.

If tests pass, the plan is to re-run them multiple times due to the intermittency of the issue.


perrydv:
@paciorek I believe this is now ready to be merged. OTOH I have already merged this into the branches for #1430 and #1428  to help them pass tests.

paciorek:
I am going to merge in #1430 rather than this one.

PR #1438, opened 2024-04-30
tweak BFGS call

perrydv:
There was a small weird thing in how we call BFGS from compiled optim. I don't think it is a bug. This PR fixes it. I don't think there will be any change in behavior but I am not 100% sure.


perrydv:
@paciorek This failed in test-ADdists. It is not clear how my tweak in nimOptim's C++ code could have impacted test-ADdists. Do you think this could be something unrelated to my change?

paciorek:
Arggh - seems like another heisenbug - I ran `test-ADdists.R` locally under `fix-BFGS` and things seemed fine.

paciorek:
Merging as we are dealing with the testing issue elsewhere.

PR #1441, opened 2024-05-16
Remove testthat::context() calls

perrydv:
testthat::context() has been deprecated for quite a while, so this PR removes uses of it in tests/testthat.


paciorek:
only failure is the new heisenbug, so merging.

Issue #1433, opened 2024-04-07
add `digits` argument to `nimPrint`?


paciorek:
In the course of looking at a Laplace/AGHQ question it was helpful to be able to print out more than the default number of digits that nimPrint/sstringstream prints.

The following modification to `cppOutputCout` and `cppOutputCoutNoNewline` could be a starting point:

```
    paste0('_nimble_global_output.precision(20); _nimble_global_output << fixed << ', paste0(unlist(lapply(code$args, nimGenerateCpp, symTab, asArg = TRUE) ), collapse = '<<'), '<<\"\\n\"; nimble_print_to_R(_nimble_global_output)')
```

@perrydv @danielturek does this seem like a useful tweak?

paciorek:
Done in #1445

PR #1445, opened 2024-05-21
add control of number of digits printed from C++

paciorek:
This adds a nimble option controlling number of digits printed from C++ via `nimPrint` and `nimCat`.

This addresses issue #1433 .

This carries the danger of upsetting testing via gold files. I set the default digits to 7 to mimic R, but we may need to change this to mimic current behavior by setting it to 6.


perrydv:
@paciorek These look like good simple changes. Just two points:

- Perhaps rephrase the User Manual edit to the shorter "described in detail in [@Michaud_etal_2021]"
- It looks like the control over digits is only done for compiled behavior. Do we want to change `nimPrint` and `nimCat` to be consistent with those? I'm not sure it's worth the trouble but flagging it to consider.

paciorek:
Yeah, I was thinking we just leave uncompiled behavior to be controlled in the usual R fashion.

I'll merge in, with the user manual edit.

PR #1447, opened 2024-05-22
categorical sampler has length control parameter

danielturek:
The title says it all.

Following suggestions in NCT # 505.

### DIFF
--- a/packages/nimble/R/MCMC_samplers.R
+++ b/packages/nimble/R/MCMC_samplers.R
@@ -163,23 +163,25 @@ sampler_categorical <- nimbleFunction(
     name = 'sampler_categorical',
     contains = sampler_BASE,
     setup = function(model, mvSaved, target, control) {
+        ## control list extraction
+        length <- extractControlElement(control, 'length', 'prob')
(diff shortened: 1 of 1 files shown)


danielturek:
Specifically, this addresses using the `categorical` sampler on distributions other than `dcat`.  In such case, the issue is to determine the number of possible outcomes (or categories) `k`, where the distribution outcomes are the values $(1, 2, 3, \ldots, k).$

In support of that, this PR: 

- Adds a `length` control parameter to the `categorical` sampler.
  - If `length` is a character string, it gives the distribution parameter name, for which the length of this parameter defines the value of `k`.
  - It `length` is numeric, it itself defines the value of `k`.
- The check for a `dcat` distribution is relaxed.  A second control parameter `anyDistributionOK` can be set to `TRUE` to subvert this check.

paciorek:
@danielturek I used the control element named `check` to be able to turn off some checking in the (new) PG sampler. 
Maybe we should use that instead of `anyDistributionOK`. I can change it if you're ok with that.

danielturek:
@paciorek Change just made on a branch, with PR, please merge it if it looks good.

PR #1448, opened 2024-05-24
Update `model_macro_builder` docs

kenkellner:
Describes the `modelInfo` and `.env` arguments to `fun` in more detail and warns about a new constant being lost if nimble moves it to data. Also adds a new example of a macro that modifies the constants.


perrydv:
These look good to me. There is a paragraph without line breaks as the others have before "#'". It might be just how it shows on my editor (emacs). At some point when we turn this feature more clearly towards users we may want to update the documentation and also get some into the User Manual. But for now these are good changes to stay up to date and complete. Thanks @kenkellner

kenkellner:
Roxygen issues should be fixed now.

PR #1449, opened 2024-05-24
Make optim methods extensible. Use nlminb by default for Laplace.

perrydv:
We have been having issues from the quality of inner optimization in Laplace approximation.

This PR includes changes to `nimOptim` and to `buildLaplace`.

For `nimOptim`:

- A new system for having compiled `nimOptim` call back to R to run an arbitrary optimizer, whose function, gradient, and/or hessian call back into compiled C++. The system uses `nimOptimMethod(<method name>, <method function>)` to register a method. Support for `nlminb` is provided using this system. Minor changes to documentation include bits in `help(nimOptim)` and also the User Manual. Both describe this new feature only briefly and state that is it subject to change in future versions.

- This does change `nimOptim` so that it accepts an `he` (Hessian calculator) function, making it less similar to R's `optim`.

- Uncompiled `nimOptim` will also allow `method` to be outside of the usual `optim` choices and if found package up appropriate functions to call an optimizer registered via `nimOptimMethod`.

- `test-optim` was updated to include `nlminb` and `my_nlminb` (same but entered as if a user was doing it) in the list of samplers for all tests. A small amount of new testing was added.

For `Laplace`:

- `nlminb` was set as default for both inner and outer optimization.
- `test-ADlaplace` was updated, primarily to tweak tolerances where needed. It was observed that some tolerances could be reduced, but I did not systematically check and do so. Other tolerances needed to be loosened. I primarily interpreted this as noise, because values checked were still close enough to be comfortable and there could be "accidents" of one method or another landing more precisely on an arg max. OTOH a few tolerances were newly uncomfortably large (e.g. 0.05 for values of order 1.0) and these were on standard errors. I determined one case (crossed random effects LME) where a fine-scale walk over the llh max was not as smooth as needed for good finite different Hessians. I tried changing to the "last.best" mode of initialization for inner optimizations, and that helped substantially.

I wonder if we should change default for `innerOptimStart` to "last.best". But I have not modified it because I know @paul-vdb is working separately on this.


paul-vdb:
@perrydv I agree that last.best makes sense as a default. That's fine to do. The new updates I've done won't be impacted by that. I've just given the user more control to choose other options which just is done by some functions that update that default option.

perrydv:
Do not merge. I will attempt to combine this branch with `AGHQ_updated`.

perrydv:
I am closing this PR because this work is being brought in through #1455.

PR #1451, opened 2024-05-29
WIP: remove virtualNodeFunctionDefinitions

paciorek:
This investigates NCT issue 500 in which we're discussing whether the virtual nodeFunction definitions in the user global environment (for user-defined functions) and in the nimble namespace (for nimble's distribution functions) are needed.

This passes `test-user.R` when run manually. I am opening this PR to run full tests.

### DIFF
--- a/packages/nimble/R/distributions_processInputList.R
+++ b/packages/nimble/R/distributions_processInputList.R
@@ -563,9 +563,6 @@ registerDistributions <- function(distributionsInput, userEnv = parent.frame(), 
             nimbleUserNamespace$distributions$add(distributionsInput)
         } else 
             nimbleUserNamespace$distributions <- distributionsClass(distributionsInput)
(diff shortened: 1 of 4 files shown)


paciorek:
@danielturek @perrydv this PR is the 'nuclear' option to address minor NCT issue 500. On the one hand, feels like we might as well remove cruft, OTOH, this is longstanding code, so I am wary of breaking something. But testing passes. It's hard to believe we need these various `node_stoch_dfoo` functions if all the various uses of nimble- and user-defined distributions are fine without them.

danielturek:
@paciorek I'm a little torn, but given the testing, I feel like it has to be an unused relic.

I'm ok if you remove it; on the small chance that something breaks, we'd probably remember this.

paciorek:
I'm going to hide the definition creation behind a nimble option and then remove in next release. Doing on a different branch/PR.

PR #1453, opened 2024-05-31
Support AD when nimbleFunction objects are used in models

perrydv:
We recently added support (for the imminent release) for using methods of nimbleObjects as user-defined distributions and functions in models.

There were shortcomings in support for AD with these new features. This PR attempts to clean up those gaps and also add tests, which were obviously missing or we would have noticed the shortcomings earlier.

I have locally successfully run `test-ADbasics`, where most of the new testing resides. Let's see any other issues with CI testing. 

Incidental to its main purpose, this PR has several other pieces:

- add tests for the recent features of having `modelDef` objects retain the user environment (previously it was retained only in `BUGSdecl` objects)
- modify `distClass` and how its objects are populated so that they get the `buildDerivs` field from any user-defined distribution (and `TRUE` or `FALSE` for built-in distributions).
- modify and add testing for `modelDef` method for checking for AD support in a distribution.


paciorek:
@perrydv ok to merge given the Win/MacOS failures are unrelated?

perrydv:
@paciorek I'm more cautious about this branch since it does change some pervasive internals like the `distClass` objects, even if only slightly.  I might have missed the test outcomes. At the moment they are showing as canceled. Did they otherwise run and pass except for the current general test issues we're seeing?

paciorek:
@perrydv I think my commit was from the GH browser just dealing with a trivial merge conflict.

Linux tests passed on your last commit - Win/Mac failures were unrelated.

So that sounds like we are ready to merge.

PR #1454, opened 2024-05-31
Allow macros to be found by evaluating expressions

perrydv:
Macros are currently an internal and unused feature, but we have substantial development going on with them for future releases. We need macros to be found by namespace operators. This PR enables something much more general, which is to search for possible macros by evaluating arbitrary expressions, which should include namespace operators. It is not easy to write a test for finding a macro in a namespace, but everything else and an indirect test on namespaces works and is added to test-macros.R. Macros are still disabled by default (`nimbleOptions("enableModelMacros")` is FALSE), so the chances of this PR breaking something should be small.


paciorek:
@perrydv ok to merge given the test failures are unrelated?

perrydv:
Yes.

PR #1455, opened 2024-06-01
AGHQ and nimOptim

perrydv:
@paul-vdb @weizhangstats @paciorek @danielturek 

This is the long-awaited PR with AGHQ from @paul-vdb along with more flexible `nimOptim` and many other changes.

A few quick notes:

- defaults for inner and outer optimization are `nlminb`.
- "Simple LME" test failed with `nlminb` and works with `BFGS`.
- Documentation needs attention.
- last test and associated topics need attention.
- `runLaplace` or such idea is not drafted
- more


perrydv:
Updates to this PR:

- Consolidate scattered methods to update settings into omnibus `updateSettings` method.
- Make some internal names more internally consistent. Also support more thorough set of control options.
- Revise roxygen.
- Add basic `runLaplace` and `runAGHQ`
- Update tests.

Remaining to-dos:
- Add `summaryAGHQ` to call `summaryLaplace to match name pairing system (AGHQ and Laplace, equivalent but both there).
- potential for more testing
- potential for ways (or one omnibus way) to reset various caches, which would require a bit of care.

perrydv:
I cleaned up tests.  Two notes:

- Some tests in `test-ADaghq` emit lots of Warnings about non-cached values. I haven't investigated. It may be a numerical tolerance issue where this is checked. But the tests still run ok.
- In `tetst-ADaghq`, I had to relax some tolerances (but still reasonable) and also set some outer methods to "BFGS" (which they would have been using when the tests were written, before the switch to "nlminb" as default.
- In `test-ADlaplace`, I had to put a tolerance of 0.2 (quite large) for one of the standard error results. This is not good.

paul-vdb:
@perrydv I wrote some RTMB code that does Laplace or AGHQ in a very basic way as a reality check. The beta-binomial example in the test takes a lot of nodes to be accurate. It might be helpful as a check that we didn't break anything in this update for some of the other examples (code is commented out on line 271 of test ADaghq) and very easy to setup for other examples.

I like the simplification of the setup code that you've done moving it into its own function. Reading through `updateSettings` that looks really good. Much better to have it all in one place as it was getting messy with the number new methods added for such small controls.

**To add is a test on 2 dimension AGHQ in test-ADaghq.**

paciorek:
merging so can proceed with other steps. We can check testing on devel.

PR #1464, opened 2024-06-04
Use safeDeparse instead of as.character in macro parameter processing

kenkellner:


perrydv:
Thanks @kenkellner . This makes sense. Is a similar change needed in `getParametersFromCode`, or are those `as.character` calls safe?

kenkellner:
@perrydv The output of `getParametersFromCodeInternal` should already be character (ultimately from `all.vars()`) or `NULL`. So all those `as.character` calls should be doing is converting instances of `NULL` to `character(0)`. `safeDeparse` would actually give the wrong result (`"NULL"`). Probably this should be more clear/explicit - I can look into it tomorrow if needed.

PR #1466, opened 2024-06-07
fix a few PROTECT and related issues flagged by CRAN checks

perrydv:
@paciorek I think this will fix the PROTECT and related warnings flagged by CRAN checks. 

Actually, per the note on lines 833-834 of cppInterfaces_nimbleFunctions.R (in `copyFromRobject`), it looks like for predefined nimble lists (as opposed to dynamically generated ones), we never use the C++ "copyFromRobject" and instead use alternative mechanisms to get content copied from R to C++. The "copyFromRobject" methods in C++ were added a some point (years ago) to speed up over variable-by-variable copying, which we still have in place as a pathway and use for these nimbleLists. As a result, I don't have a way to test this code and it is possible we could simply remove these methods. The intention was likely to bring them back into use, but instead now they are stale. However at the current juncture I would prefer to leave them in to make the smallest possible change (in case, e.g., that comment itself is not 100% correct).

Counting PROTECTs and UNPROTECTs is tricky in our code because some of them are within C++ macros. But it looks like CRAN checking somehow does that and the counts did need correcting. They may have never tripped run-time errors because of the above point, that they might never actually be called.

The issue in the long-named C++ function relating to `findVarInFrame` was, I think, telling us to add some PROTECTs, which I did. This function should be used a fair bit (but for AD only), so if tons of tests blow up, then I made a mistake.


paciorek:
I'm not sure of the details of how the counting works, but I am wondering if there are mismatches as well in some `copyFromRobject` items that CRAN didn't flag, e.g., `waicNimbleList::copyFromRobject` has `UNPROTECT(6)` but only 3 macro calls.

perrydv:
The macro `SETUP_S_xData` contains 3 `PROTECT` statements. Generally each of the `COPY_<SOMETHING>_FROM_R_OBJECT` macros contain 1 `PROTECT`. In `waicNimbleList::copyFromRobject`, there is `SETUP_S_xData`(a macro with no arguments) and then 3 `COPY_<STUFF>` macros, so that adds up to 6.

PR #1474, opened 2024-07-18
Reinstate RW_multinomial sampler

danielturek:
The title says it all.

Do not merge.  This is just to have an open PR, and easily see the diff.


danielturek:
Not going to use this version of the sampler.  Will remake a different version of `RW_multinomial` sampler in a different PR.

Issue #1471, opened 2024-06-24
error trap multivariate samplers assigned to target nodes of RJMCMC


paciorek:
As discussed with a user, it doesn't make sense to have a mv sampler assigned to a node that is being configured for RJMCMC. That sampler will be toggled off and wouldn't sample the other parameters that it is supposed to sample. And we don't have a way to have the mv sampler operate in the reduced dimension space when the target node is not in the model anyway.

So we need to add an error trap that having mv sampler on a target node is not allowed.

While we are thinking about this, we should also consider what happens if the target node has two samplers (either two univariate or one univar and one mv) assigned to it. I don't know that we would handle this correctly - presumably we want to toggle off both samplers. I haven't checked what would actually happen.

danielturek:
Agreed that these cases need some attention.  My initial thought would be to error-trap any of these non-standard cases: when anything other than a single, univariate sampler, is operating on a node specified for RJMCMC.

I'd be happy to take care of this a bit later in the summer, unless anyone else gets to it first.

paciorek:
Ack, ignore the weirdness that the message came from "jasa-acs". I was mistakenly signed in to that GH account.

danielturek:
Addressed in PR #1475.

@paciorek In regard to your question: we previously *already did* handle the case of "the target node has two samplers (either two univariate or one univar and one mv) assigned to it", in which case our previous "handling" was to:

- toggle the first such sampler according to RJMCMC
- completely remove any other samplers assigned to `targetNode`

You'll read in the comments above that I changed the handling of this case, so the new behaviour is:
- quit

danielturek:
Merged #1475 after tests passed.

Closing issue.

PR #1475, opened 2024-07-19
updated checks for configureRJ

danielturek:
Updated checks for `configureRJ` function:

- Did basic refactoring
- Fixed a bug in one set of checks (previously was omitting a call to `skip`)
- In the case of "more than one sampler already on target node", we now bail out of `configureRJ` routine, rather than toggling the first sampler, and removing the other samplers (the previous behaviour)
- New check added: when an existing sampler is updating the `targetNode` *and also other nodes*, we now bail out of the `configureRJ` routine (this case was the impetus for this work).

Also added relevant testing.


danielturek:
Fixes #1471

salleuska:
Thanks, Daniel for taking care of this! I had it on a to-do list since the user discussion, but I haven't had much time to think about it.

danielturek:
@salleuska No problem!

PR #1479, opened 2024-07-21
remove template arguments from NimArr constructors and destructors for C++20

perrydv:
This attempts to fix the recent new CRAN R-devel warnings on fedora-gcc.

@paciorek Do you have a way to test whether this resolves the CRAN testing warnings on that platform?

### DIFF
--- a/packages/nimble/inst/include/nimble/NimArr.h
+++ b/packages/nimble/inst/include/nimble/NimArr.h
@@ -119,7 +119,7 @@ class NimArr<1, T> : public NimArrBase<T> {
     return (*NimArrBase<T>::vPtr)[calculateIndex(i)];
   }
 
-  ~NimArr<1, T>() {}
+  ~NimArr() {}
 
   template <class Tother>
   NimArr<1, T> &mapCopy(const NimArr<1, Tother> &other) {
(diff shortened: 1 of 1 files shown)


paciorek:
Not sure - we could try `rhub` along the following lines (but choosing the platform option that makes most sense),

```
Rscript -e "library(rhub);check(platform = c('ubuntu-gcc-devel','windows-x86_64-devel','linux-x86_64-rocker-gcc-san','fedora-clang-devel'))" >> /tmp/rhub.out
```

but when I last tried it I ran this, so we'd have to dig into that:


```
As of 2024-06-06, having SSL certificate issues. Also it looks like rhub has updated its workflow substantially so a bit of work may be needed to get this working again.
Error in curl::curl_fetch_memory(url, handle = handle) : 
  SSL peer certificate or SSH remote key was not OK: [builder.r-hub.io] SSL certificate problem: self-signed certificate
```

I'm happy to look into it later this week (have a retreat tomorrow and Wednesday).

paciorek:
Per email discussion (and updated info in `RELEASE_INSTRUCTIONS`) I got rhub working again, and specifically use the `gcc14` platform to verify that nimble builds without warnings with the fixes in this PR.

PR #1480, opened 2024-07-23
fix uncompiled use of updateSettings in Laplace or AGHQ

perrydv:
A minor bug in buildAGHQ came to my attention. If `updateSettings` method is used *prior to compilation* with inputs that change `constant_init_par`, it can break the `one_time_fixes` system by omitting the trailing -1 needed to confirm it is a vector for compilation. This PR fixes that. I have not added a test that triggers the bug and shows it is fixed by this PR. First I am putting this up to run current tests and can add a new test later.

### DIFF
--- a/packages/nimble/R/Laplace.R
+++ b/packages/nimble/R/Laplace.R
@@ -314,6 +314,8 @@ buildOneAGHQuad1D <- nimbleFunction(
           }
         }
       }
+      if((!one_time_fixes_done) & (length(constant_init_par) == 1))
+         constant_init_par <<- c(constant_init_par, -1)
       if(optimWarning != -1) {
         warn_optim <<- optimWarning != 0
       }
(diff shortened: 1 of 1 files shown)


danielturek:
@perrydv @paciorek Should these changes be made in `nimble-dev/nimble`, or in `nimbleQuad`?

I understand the importance of keeping the `nimble-dev/nimble` release of Laplace/AGHQ correct.  But this workflow may not be realistict, for eventual migration (of any number of such changes) to `nimbleQuad`.

paciorek:
`nimble` is still the active location for Laplace/AGHQ so we definitely need it here. And we need it in `nimbleQuad` (thanks @danielturek for the issue posted there). 

I'm hoping that next nimble release removes Laplace/AGHQ so this does not repeat.

PR #1496, opened 2024-10-17
enhance handling of optimization for Laplace

paciorek:
WIP: do not merge without discussion.

In particular @perrydv see my changes to `custom_optim_inner`.

Flagging @weizhangstats @paul-vdb @kenkellner in case anyone has thoughts on item 3 below or other thoughts on readiness to merge this into devel.

This makes a few enhancements to the "laplaceRpieces" work:

 1. Perry added capability for inner optimization to use inner Hessian to allow use of Newton's methods
 2. Wei fixed a small bug in `buildOneAGHQuad`
 3. Chris modified `custom_optim_inner` to return `NaN` rather than stopping if inner optimization fails. This allows outer optimization to proceed (e.g., in cases where BFGS tries an outer parameter value that causes inner gradients of NaN/Inf/huge values). He saw that for the CDFW carnivores example, such inner gradients occur (including when using TMB via `unmarked`), so we need robust handling of such inner optimization failures.

One open item. We have `innerOptimWarning` (default of FALSE) in our Laplace system to avoid overwhelming user with warnings about problems in the inner optimization. This needs to be extended into Chris' changes to `custom_optim_inner` presumably. Right now all warnings are being printed. 

Also we'll presumably need to change the default inner optimization to `nlminb` per [NCT issue discussion](https://github.com/perrydv/nimbleCoreTeam/issues/514).


weizhangstats:
The changes look good to me. :) I fixed a wee bug in setupMargNodes in `nimbleQuad`. Not sure when Laplace will be reallocated to `nimbleQuad`, so I just fixed it here as well.

paciorek:
The test failure is because `mcmcse` is not being installed on the runner, causing `test-mcem.R` failures. It fails in this case because Rcpp fails because of a compilation issue. Strangely on `devel`, `mcmcse` is installed but no mention of installing  Rcpp is given, and then there is a later message (after successful `mcmcse` installation) saying that `mcmcse` is not available. But then `test-mcem.R` runs fine. 

I'm retriggering the failed test batch in hopes this was a stochastic issue for the runner.

Update: Rerunning resulted in `test-mcmcse.R` being successful.

paciorek:
The most recent commit tries to finalize our Laplace interface and documentation.  

@perrydv @weizhangstats @paul-vdb please let me know of any thoughts you have.

These include: 

-  Make nlminb the default for inner/outer optim for Laplace.
-  Make last.best default starting point for inner Laplace.
-  Silence R error messages in `custom_optim_inner` so that messages about inner convergence failure only occur if `innerOptimWarning = TRUE`.
-  Update manual on Laplace to highlight `runLaplace` (mentioning individual Laplace methods second, and then discuss user provided optimizers.
- Clean up messaging in Laplace.R to better follow nimble standard.
- Return logLik and df in `summaryLaplace` (to allow for model selection)

weizhangstats:
All looks good to me @paciorek. I just edited the `last.best` description a bit - to me, it means that the optimised random effects values that correspond to the currently best outer optimisation (i.e. largest marginal log likelihood value) will be used for inner optimisation. The previous wording "best inner optimisation so far" is a bit vague.

weizhangstats:
@paciorek `testADaghq` failed here due to a wee tolerance issue but it seems to work ok on my mac.

paciorek:
@weizhangstats thanks for your recent improvements.
I'll figure out the tolerance issue on that last failing test.

perrydv:
This all looks good to me. Thanks @paciorek

PR #1499, opened 2024-10-21
manual information for macros

paciorek:
WIP: This PR is for us to discuss content related to macros for the nimble manual.

I've taken a stab at giving some high level information and brief examples (copied from Ken's nimbleMacros vignette) into the Writing Models and User-defined chapters.

Related notes:
 - I think we can basically leave all of what Ken has in the nimbleMacros vignette there.
 - Furthermore, Ken has a full set of roxygen in both `nimble` and `nimbleMacros` for user-facing macro-related functions.

@kenkellner @perrydv @danielturek Let me know of any comments/suggestions.


kenkellner:
I removed the reference to `LINPRED_PRIORS`.

I also made some changes to the code. 

* It's not clear to me if these code blocks are meant to be executed since most of the other chunks in the file are `eval=FALSE` (but these ones are not). If they are meant to be run I think we need to load `nimbleMacros`.
* If splitting constants and data, need to specify the index range of `weight` (it needs to be initially in the constants in order for `LM` to auto detect that index)
* I tried running `nimbleMCMC` on both models. First one worked OK. In the second one I found that if the SD has `dhalfflat` prior there are MCMC issues. Maybe need to set inits? Instead of doing that I just changed to a uniform prior which works fine, but happy with whatever you think is best here.

PR #1501, opened 2024-10-26
catch case of discrete (integer) target with only one viable next value

perrydv:
... when at modal value.

### DIFF
--- a/packages/nimble/R/MCMC_samplers.R
+++ b/packages/nimble/R/MCMC_samplers.R
@@ -872,7 +872,6 @@ sampler_slice <- nimbleFunction(
         maxSteps               <- extractControlElement(control, 'sliceMaxSteps',          100)
         maxContractions        <- extractControlElement(control, 'maxContractions',        1000)
         maxContractionsWarning <- extractControlElement(control, 'maxContractionsWarning', TRUE)
-        eps <- 1e-15
         ## node list generation
         targetAsScalar <- model$expandNodeNames(target, returnScalarComponents = TRUE)
         ccList <- mcmc_determineCalcAndCopyNodes(model, target)
@@ -886,7 +885,8 @@ sampler_slice <- nimbleFunction(
         if(getNimbleOption('MCMCsaveHistory')) {
             saveMCMChistory <- TRUE
         } else saveMCMChistory <- FALSE
-        discrete      <- model$isDiscrete(target)
+        discrete <- model$isDiscrete(target)
+        eps <- 1e-15
         ## checks
         if(length(targetAsScalar) > 1)     stop('cannot use slice sampler on more than one target node')
     },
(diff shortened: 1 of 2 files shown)


danielturek:
@perrydv I'm pushing minor cleanup to this, but the logic looks sound.

This is a good patch to have - and I'm also fairly surprised we've never encountered this before.

Also, a push will re-trigger testing, to see if the previous failure with the PG sampler is (perhaps?) intermittant?

perrydv:
Thanks @danielturek . Would you also be up for making the changes in `sampler_slice_noncentered`?

I have not investigated the test failures in detail. The one in test-polyagamma looks like it might be a simple monte carlo tolerance issue, where if we inspect and trust the (very slightly different from previous) result we can tweak the tolerance for the test to pass. The test-mcmc failures is a case where gold-file results for a slice sampler example (line 336 of test-mcmc) are no longer matched. That makes sense too. However, superficially the numbers look a bit more meaningfully different, so we'd have to look at whether this is expected to be a well mixed, converged result, etc and decide whether we are confident updating the gold file. Do you want to look at these? Noting for @paciorek too.

danielturek:
@perrydv I've made similar changes to the `slice_noncentered` sampler.

Do you have a test case that triggers this error (with contractions to a single integer), which we could add into the MCMC testing?

In any case, the MCMC testing gold file looks like it will need to be updated.

paciorek:
@perrydv We discussed closing this PR without merging as a mistaken diagnosis. Can you comment here on why the changes made in this PR don't make sense to do (perhaps regardless of the user's use case)?

perrydv:
Yes let me explain and leave a record here. This PR arose from a user report of excessive warnings from slice sampler of "maximum number of contractions". The case involved using HMC/NUTS on some parts of the model and slice sampler for Poisson-distributed latent states (in an ecological N-mixture model). At the time I mis-diagnosed the problem as arising purely from the slice sampler and thought I saw a way to catch the situation and handle it (in this PR). When I came back to this again I realized that the problem really arose from specific iterations in warmup mode of the HMC/NUTS sampler that could leave the model in an incorrect state. There is a PR on nimbleHMC to fix that, and I think we can close this nimble PR without merging. @danielturek it would be good to loop you in here as well.

danielturek:
@perrydv It still raises the question from @paciorek as to whether these defensive steps (in this PR) are a good idea, regardless.  If we're concerned about runtime/efficiency, that's a valid discussion.  But if not, then it seems this PR is only safe and defensive against problems we may not have encountered yet.

perrydv:
Thanks @danielturek . Actually I'm not convinced I got these defensive steps right or that they are necessary. For a discrete (integer) node, we treat it as if it is a continuous node with flat pdf intervals. In the relevant case where there is only one plausible value, the stepping out will get outside that value and then when the contractions hit a value anywhere inside the interval, it should be accepted. I think in my previous try at debugging, I didn't realize the model state on entry was wrong, and then I mis-diagnosed a problem that isn't really there. And also I think the proposed stopping condition could in some cases stop the contractions when they really should continue. So I'm thinking since it wasn't actually broken we shouldn't try to fix it (at least without further investigation). (First version of this comment had a missing "not" and has been edited.)

danielturek:
@perrydv This makes more sense to me.  I never dug into this issue, but I agree with your current take on how the slice sampler **should** operate in the discrete case.

I'm less convinced that "the proposed stopping condition could in some cases stop the contractions when they really should continue", but only because I haven't seen such a case, and also haven't thought about it.

That said, I can definitely be on board with "if it isn't broke, don't fix it".

danielturek:
If we close this PR, it would still be available (as a closed, un-merged PR), should we ever decide we wanted it.

PR #1502, opened 2024-10-28
various cleanups for 1.3.0

paciorek:
This does mostly cosmetic work for 1.3.0.

It includes renaming of some macro-related code.

Also addresses case where `addSampler` is given a sampler name that includes the namespace.


paciorek:
@kenkellner I've merged in `cleanups_1.3.0`, which has the macro-related changes we've been discussing, into `devel`. 
So I think you can go ahead and update `nimbleMacros`  to be consistent with `devel`.

PR #1505, opened 2024-10-30
removeExtraBrackets handles for loops without {}

### LINKED ISSUE #1503: bug in `removeExtraBracketsInternal`
@kenkellner it looks like there is a bug in the logic of `removeExtraBracketsInternal` when handling for loops that don't have the body in `{...}` because the body is a single line of code (so the brackets are optional).

```
code = nimbleCode({
                  for(j in 1:3)   # NOTE no bracket
                    for(l in 1:2) {
                    y[i,j,l] ~ dnorm(mu[1+k1[i], 2, k2[i]], var = s0)
                    mu[i, j, l] ~ dnorm(0, 1)
                }
       })
nimble:::removeExtraBracketsInternal(code)
```
Note that the `mu` declaration is stripped out in the result below. If I have the `j` loop body be in `{...}`, then things are fine.
```
[[1]]
`{`

[[2]]
for (j in 1:3) for (l in 1:2) y[i, j, l] ~ dnorm(mu[1 + k1[i], 
    2, k2[i]], var = s0)
```

This is what is causing the two test failures in `test-dynamicIndexing.R` of the PR for branch `cleanups_1.3.0` on GitHub. I'm also guessing/hoping it is causing the test failure in `test-copy.R`.

kenkellner:
Should fix #1503. I added the example from the issue to the tests along with some additional tests.

I checked the two failing tests in `test-dynamicIndexing.R` and they work for me now. I ran `test-copy.R` and everything seemed to work although I was a bit confused by the structure there, wasn't clear to me exactly where the test failure was.


packages/nimble/tests/testthat/test-macros.R
```diff
@@ -531,7 +531,7 @@ test_that('duplicate nested indices from macro expansion error-trapped correctly
         }
         )), regexp = 'Stopped after processing macros and if/then/else statements')
     )
-    expect_true(grepl("dnorm(mu[i], 1)", out[2], fixed=TRUE)) 
+    expect_true(grepl("dnorm(mu[i], 1)", out[3], fixed=TRUE))
```
kenkellner, line 534: Previously the result of the macro after processing was all on one line
```r
{
for (i in 1:3) for (i in 1:10) x[i] ~ dnorm(mu[i], 1)
}
```

Now it is
```r
{
    for (i in 1:3) for (i in 1:10) {
        x[i] ~ dnorm(mu[i], 1)
    }
}
```

Based on the way the macro code is written I think the updated version is more correct (though they should be equivalent?).

PR #1512, opened 2024-11-08
Robustify handling of numerical issues in samplers

paciorek:
WIP

This is a draft that addresses [NCT issue 538](https://github.com/perrydv/nimbleCoreTeam/issues/538).

Per @danielturek , it creates `checkLogProb`, which has two purposes - one to replace NaN values with -Inf (since we generally want to take a calculation that produces NaN as a rejection), and to print a warning message when an Inf is found.

I've modified various samplers and utility functions to use the new function. It feels a bit kludgey in that we handle logProb calculations (and some degree of checking already) in various ways in our samplers. I've tried to interject `checkLogProb` in the most reasonable locations in the samplers, but my choices are open to other ideas. Given the accretion of the various components of acceptance/rejection, it was hard to do this in clean, consistent way. My goal is simply to make the samplers more robust.

@danielturek I'd particularly like your review of this. 

Will be interesting to see what happens with testing. Given we haven't seen a lot of tickets related to numerical issues, I'm hoping these changes will have little effect. However, I am a bit worried that there have been cases of using samplers where `Inf` values might arise (such as the stickbreaking nimble-users post that inspired all this) where the sampling might give invalid results.

I'll also note that PR #1509 is related to this.

Other things to consider:

 - [ ] perhaps use `checkLogProb` in the new code in `sampler_CRP` in PR #1509.
 - [ ] look at SMC samplers to see if we want to do this type of work there
 - [ ] look at HMC samplers as well


danielturek:
@paciorek Thanks for tagging me on this.

This feels like I'm overlooking something for certain, but how is it that `decide` is defined as a regular R function (as opposed to a `nimbleFunction`), and is used extensively in `run` function code of samplers?  What am I overlooking?

Would another call to `checkLogProb` be appropriate for line 372 of `MCMC_samplers.R`?  This is in the core `RW` sampler, where we separate the logProbs for the target node and dependent nodes:
```
            logMHR <- logMHR + model$calculateDiff(calcNodesNoSelf) + propLogScale
```

How about on line 121 of `MCMC_utils.R`, which is in the function `setAndCalculateOne` ?
```
       lp <- model$calculate(calcNodes)
```

Same for line 164 of `MCMC_utils.R`, which is in the function `setAndCalculate` ?

Same for line 180 of `MCMC_utils.R`, which is in the function `setAndCalculateDiff` ?

(I know that sometimes the result of these functions in the `setAndCalculate` family are later wrapped in `checkLogProb`.  But from my review, this isn't always the case, so my suggestions above are defensive).

For as much as I've tried to standardize things, looking over this has really shown me what a hodgepodge of different structures and approaches the various samplers use ...

Overall, this PR feels like a good approach, given that we want to handle this corner case.

paciorek:
@danielturek thanks, I will look at your suggestions in detail.

As far as `decide`, it's also a C++ function in `Utils.cpp`.

paciorek:
Thanks for catching the missing `checkLogProb` in `sampler_RW`.

As far as potential uses in those locations in `MCMC_utils.R`, my logic was to use `checkLogProb` just before the result would be used for a accept/reject decision or something similar, and not just whenever we call `calculate`. So I'm inclined to leave them out of `MCMC_utils.R` except in `decideAndJump`.

perrydv:
@paciorek Any chance you'd be up to run a very simple benchmark or two? It seems like it shouldn't cost much, but if it is called so much it could be worth checking. One idea would be to code it directly in C++, so it is added to the DSL, and then we could mark it as `inline`, encouraging the compiler to always insert it in place. The message itself could be contained in another function, in order to keep it out of inlining, since it is only needed when there is a problem.

paciorek:
Yeah, good idea. I took a dead simple normal-normal model with one observation and a single (mean) parameter and tried RW and slice under current nimble and with this change.

It was surprisingly hard to get robust timings, but it looks like for the RW it's about 5% slower and for the slice 8-10% slower.

Since that is for a worst case scenario with the nearly the least amount of calculation relative to the new checking that seems possible, I'm inclined to just merge it in. We could always improve efficiency in a future release.

Thoughts?

paciorek:
@perrydv I went ahead an inlined `checkLogProb`. Please take a quick look and see if there's anything that is not what you were thinking. Also do you know if it should be `inline` or `static inline`?

paciorek:
Also, I meant to say that when I run the simple normal-normal case with the inlining, the time seems to be the same as without `checkLogProb` at all, so the inlining does seem to help.

perrydv:
Thanks @paciorek that looks good to me and good to hear the performance improved.

I think it's better to use `inline` rather than `static inline`. The internet has some interesting explanations on this. Basically we don't need the `static`.

paciorek:
@perrydv in `Utils.h` we have `static inline` for `isTRUE`: `static inline bool isTRUE(bool x) {return(x);}`.

Should I remove 'static'?

Issue #1485, opened 2024-07-31
can't use `nimCat` with nimbleList


paciorek:
```
mynf <- nimbleFunction(
    setup = function() {
        myList <- nimbleList(x = double())
    },
    run = function() {
        ll <- myList$new()
        ll$x = 3.0
        z = 5.0
        nimCat("z: ", z, "; x: ", ll$x, "\n")
    }
)

rmynf <- mynf()
cmynf <- compileNimble(rmynf, showCompilerOutput=TRUE)
```

One can of course work around this by setting a temporary variable equal to `ll$x` and cat/printing that.

Issue #1484, opened 2024-07-31
attempt to use non-existent nimbleList not cleanly error trapped


paciorek:
```
mynf <- nimbleFunction(
    setup = function() {
        myList <- nimbleList(x = double())
    },
    run = function() {
        ll <- myList$new()
        branch$x = 3.0  # misnamed
    }
)

rmynf <- mynf()
cmynf <- compileNimble(rmynf)
# Error in if (!(isSymFunc || isSymList)) stop(exprClassProcessingErrorMsg(code,  : 
#  missing value where TRUE/FALSE needed
```

PR #1518, opened 2024-11-25
Fix list issues: allow printing of list elements with nimCat and trap undefined lists

paciorek:
This addresses issues #1485 and #1484.

1. Currently we have an untrapped compilation error if using `nimCat(myList$x)`, because we don't dereference the pointer when using `myList$x` in `nimCat`. Usually this happens in `sizeNFvar`. I've taken some of that processing into a new helper function `dereferenceListElements`, called from `cppOutputCout{,NoNewLine}`. It feels a bit hacky to do it there rather than earlier in processing. That said, it's not clear to me why the current handling of this is in `sizeNFvar` as it doesn't seem to really relate to size processing. 
2. Currently we have an untrapped error if using a nimbleList that hasn't been created. This PR traps that.

@perrydv see what you think specifically regarding the first item.


paciorek:
It turns out that printing list elements via `nimPrint` works fine. Presumably because we are forcing `nimPrint` code through eigenization but not `nimCat`. I think this is just an oversight, so I am adding a commit to this PR trying to harmonize handling of `nimCat` and `nimPrint`.

perrydv:
@paciorek These look like relatively simple changes and they look like they make sense. LGTM.

PR #1527, opened 2025-02-12
Fix sd for dnorm in parameterizationTableLong.md

adamgorm:
It said that sd = 1 / sqrt(var), where it should have been sd = 1 / sqrt(tau).


paciorek:
Thanks for the catch!

PR #1532, opened 2025-02-26
parameterTransform handles multivariate user dists, with a warning

danielturek:
The title says it all.

In such cases, no transformation is applied to the node(s) following the multivariate user-defined distribution, and an informative warning is issued.  In addition, there's a new system-level option (`parameterTransformWarnUserDists`) which can turn off the warning.

I did some a quick test, and I believe this works as intended.

Anyone else is welcomed to modify the warning message, if they want.


danielturek:
@perrydv Only addressing your comment about `parameterTransform` and multivariate user-defined distributions.

I just checked, and those changes (addressing this issue) appear to still be on branch `transform_user_dist`, which is not merged.

danielturek:
My previous comment was intended for NCT issue 567, and has been repeated there.

Issue #1533, opened 2025-03-06
internal error when running compileNimble


ralmond:
I've set up a model as follows:

```
>brownianModel1 <- nimbleCode({

  #... Skipping temporal part of the model

 for (i in 1L:N) {
    for (it in 1L:NT) {
      et1[i,it] <- ilogit(1.7*(a*theta[i,it]-b1))
      et2[i,it] <- ilogit(1.7*(a*theta[i,it]-b2))
      p[i,it,1L] <- 1-et1[i,it]
      p[i,it,2L] <- max(0,et1[i,it]-et2[i,it])
      p[i,it,3L] <- et2[i,it]
      y[i,it] ~dcat(p[i,it,])
    }
  }

 #... Skipping parameter priors
})
```
The full model (as well as some data generation code can be found in):

[hmm-bug.Rmd.txt](https://github.com/user-attachments/files/19112373/hmm-bug.Rmd.txt)

The model is an item response theory graded response model.  `et1` represents the probability of scoring a 2 or higher, and `et2` the probability of scoring a 3 or higher, the categorical probabilities are then the differences among those curves.  The `max` is to ensure that I don't accidentally get a negative probability.

I then set up initial values planning to run MCEM

```{r}
>bminits <- list(
  mu_0 = 0,
  sigma_0=1,
  a=1, b1=-1,b2=0,
  growthSD=rep(.1,N),
  growthRate=rep(.25,N)
)
>brownianModel <- 
  nimbleModel(brownianModelCode1,
              constants=list(N=N,NT=NT),
              dimensions=list(p=c(N,NT,3L)),
              inits=bminits,check=FALSE,
              data=list(y=y),
              buildDerivs = TRUE)
                                        
```

These steps run with some warnings about unitialized variables.

The problem comes with
```
cBrownianModel <- compileNimble(brownianModel)
```

This fails with the error:
```
Error in if (iName == "log" && (is.numeric(iArg) | is.logical(iArg))) { :
missing value where TRUE/FALSE needed
```

The error happens in `cppOutputNimDerivsPrependType`, where `iName` is equal to `NULL` and 
```
> NULL=="log" && TRUE
[1] NA
```

1) There is pretty obviously an unhandled case in `cppOutputNimDerivsPrependType` when `iName` is null.

2) It would be really useful if `compileNimble` could catch low-level exceptions and add context inform (e.g., what line of code or operation it was working on when the error occurred) as that would aid debugging.

ralmond:
[...]

PR #1557, opened 2025-06-21
Fix compilation error with min/max under certain circumstances

paciorek:
In the following case, we have an untrapped error (see issue #1533):

```
code <- nimbleCode({
    x <- max(0,y-z)
})
m <- nimbleModel(code,inits=list(z=-1,y=5), buildDerivs=TRUE)
cm <- compileNimble(m)
```

It occurs with `if(iName == 'log' && (is.numeric(iArg) | is.logical(iArg)))` in `cppOutputNimDerivsPrependType`. When the condition after `&&` is TRUE, this errors if `iName` is NULL. If condition after `&&` is FALSE (e.g., if the model code is `x <- max(w,y)`, then execution continues.

I've wrapped the check in `isTRUE` as a rudimentary fix.

(Side note: `max(a,b)` is converted to `pairmax` behind the scenes, and I think this is why `cppOutputNimDerivsPrependType`, but I haven't fully grasped the details.)


paciorek:
Also, just noticing we need to convert `|` to `||` to be consistent with R's handling of scalar versus vector boolean arithmetic (and I think R is getting more strict about this in some recent or future version).

PR #1567, opened 2025-07-03
Add derived quantities info to manual

paciorek:
This adds a section to the MCMC chapter of the manual. It also mentions derived quantities in various places in the manual (the lightning intro and various sections of the MCMC chapter).

It's basically the content in [Daniel's vignette](https://danielturek.github.io/public/derived_quantities/derived_quantities.html), but I've rearranged and reworded the content a bit, added some framing text, as well as shortening somewhat.

I think it's in decent shape, so while I welcome comments and am sure there are places that could be improved, commenting may not be a high priority, though I expect @danielturek will want to take a look when he has a chance. 

I'm making this a PR for comments only -- I've stopped the CI as there is nothing code-wise in here.


danielturek:
@paciorek Thanks for the opportunity to review this.  I have some suggestions and comments below.

In `chapter_MCMC.Rmd`:

- Line 33. Do we want quotes around "derived" only?  I would perhaps suggest omitting any quotes entirely, or quoting all of "derived quantities".  Noting also that "derived" is not quoted in any other places.
- Line 103.  I'll have to review also, but I'm not sure it's accurate to say "and the summary statistics (but not the stand-alone means and variances or the WAIC) use the thinned samples".  IIRC the means and variances accept an `interval` on which to use samples for calculations, and `interval` defaults to the MCMC `thin` interval.  Ah, reading this text again now, I understand that "stand-alone means and variances" might be referring to the summary statistics output in the `summary`; in which case, that certainly only uses the thinned samples, since it's calculated directly from the collected `samples`.  Perhaps I'm not understanding this text (line 103), but it doesn't seem correct to me.
- Line 120, says "By default, only posterior samples are returned", I don't think that's correct, when `logProb = TRUE`, I think the joint model log-prob will also be returned.
- Line 132, typo "overal" should be "overall".
- (I, personally, prefer and intentionally use the all-lower-case style for R code comments; that is, not capitalizing the first letter of "sentences"; oh well).
- Line 200, maybe add an additional sentence, to the effect of "When doing so, samples for any predictive nodes will no longer be included with the MCMC samples array, but rather will appear as part of the `derived` MCMC output."
- Line 479, this paragraph is a little casual.  Looking beyond that, there's a typo near the end, should include the word "to", in "it's up to the person..."
- Line 675, probably back-ticks should be added to surround the *second* instance of both "muSq" and "z", in addition to the first instances of each.

paciorek:
Thanks for the careful read; I've fixed all this to your specification. Except that regarding line 120, it is the case that *by default* only the samples are returned. It's just that in the example code, `logProb=TRUE`, overriding the default. So I've kept that as is.

And as far as lower-case comments, my understanding is that upper case is generally the standard amongst developers,
so I've generally shifted to that.

Given no other comments I will probably merge this in soon. We can always make subsequent changes on `devel`.

danielturek:
@paciorek Looks good to me.

PR #1569, opened 2025-07-03
Pg simple update

paul-vdb:
Three changes:
1) Initiate `inflationNodes` to avoid a small error when `check = FALSE`.
2) Ensure `initializeX=FALSE` if a design matrix is passed
3) Add a warning if infinite values are found in the design matrix. Suggests to scale the design matrix in advance.

@paciorek Should we put a full stop if infinite values are detected in the design matrix instead of a warning?
Does it make sense to allow the user to override this warning? Currently `infiniteOkay` is a list input. Better name? I need to add it documentation still.

### DIFF
--- a/packages/nimble/R/MCMC_samplers.R
+++ b/packages/nimble/R/MCMC_samplers.R
@@ -3174,6 +3174,7 @@ sampler_polyagamma <- nimbleFunction(
             if(initializeX & !infiniteOkay) {
               for(j in 1:nCoef){
                 if( any(abs(X[, j]) == Inf) ){
+                  ## @CJP: Maybe we should change this to a stop? I suspect it can happen if we have uncaught zero inflation on a stochastic column...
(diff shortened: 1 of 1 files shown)


paciorek:
Sorry about the delay in looking at this. I have some questions.

1. Regarding the user providing `inflationNodes` if checking is not being done, I don't see why we need to have the user provide. Checking is just to make sure things are all valid and if the user doesn't want that done, I think we just find the `inflationNodes` using `getParents` and `setDiff(inflationNodes, nonTarget)`. I.e., move those lines of code that initially determine `inflationNodes` outside of `if(check)`. Am I missing something?
3. Could we actually have valid computation in the sampling if there are Infs in X? I think we would get an invalid sample. So unless I'm missing something I think we should error out and not provide an override.

paul-vdb:
@paciorek Thanks. I'll make the suggested changes this afternoon and merge into devel if there are no conflicts etc.

PR #1570, opened 2025-07-09
Add `glueNimbleCode()`

adamlilith:
`glueNimbleCode()` combines output from different calls to `nimbleCode()` so it can be processed by `nimbleModel()`.

Useful for cases where we have the same basic `nimble` code, but want to tack on different blocks depending on the workflow. Obviates having a different script for every small difference between models. A very simple example:
```
code_start <- nimbleCode({
   sigma ~ dunif(0, 100)
   for (i in 1:n) {
      y[i] ~ dnorm(mu, sd = sigma)
   }
})

broad_prior <- nimbleCode({ mu ~ dnorm(0, sd = 100) })
reg_prior_prior <- nimbleCode({ mu ~ ddexp(rate = 1) })

if (dataset == 'a') {
   model_code <- glueNimbleCode(code_start, broad_prior)
} else if (dataset == 'b') {
   model_code <- glueNimbleCode(code_start, reg_prior)
}

model <- nimbleModel(model_code, ...) # etc
```

### DIFF
--- a/packages/nimble/R/BUGS_readBUGS.R
+++ b/packages/nimble/R/BUGS_readBUGS.R
@@ -149,6 +149,40 @@ nimbleCode <- function(code) {
 
 BUGScode <- nimbleCode
 
+#' Combine outputs of two or more calls to 'nimbleCode()'
+#'
+#' Combines the outputs of two or more \code{\link{nimbleCode}} calls into one, which can then be used by \code{\link{nimbleModel}}.
+#'
+#' @param ... Two or more outputs from \code{\link{nimbleCode}.
+#' @author Adam B. Smith
+#' @returns Concatenated `nimbleCode()` objects.
+#' @examples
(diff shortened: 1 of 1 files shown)


paciorek:
@adamlilith thanks for this contribution. This is a nice idea and we can appreciate the appeal of being able to glue together different model components. 

We've had a bit of discussion of this on the dev team about whether to accomplish this as you've outlined or instead allow `nimbleCode` to take `...` , allowing a user to provide multiple model code chunks that way.

I'm hopeful we'll have something along these lines in the next release (which isn't too far away), though it might not be exactly your approach.

adamlilith:
I'd be happy with the concept, however it's implemented!
A
________________________________

paciorek:
We're taking the approach of adding functionality to `nimbleCode` in the `nimbleCode_combine_code_objects` branch.

PR #1573, opened 2025-07-12
add manual info for nested approximation

paciorek:
This adds manual info for the nested approximation to the Other Algos chapter and moves Laplace manual info there as well from the AD chapter.

In addition to the actual manual text, this is a call for feedback on the user interface for the nested approximation, which consists of `buildNestedApprox`, `runNestedApprox` and the various methods of the `approxSummary` class, which allow one to make further refinements to the inference from the nested approximation.

@paul-vdb please take a careful look to the manual and make edits as you desire. 

@weizhangstats @jmhewitt @perrydv @danielturek we welcome any input, in particular if you have thoughts on the user interface. I put a lot of thought into it and Paul has taken a look at it already, and I think we are pretty happy with it. But now is the time for critical feedback, before we release. I would not be surprised if we've overlooked things or aspects might be confusing to users.

Some things to think about:

1. I haven't done much wordsmithing or careful editing yet. So if you see typos or places that just need some additional attention, feel free to just point things out (and don't worry about typos as I'll catch those later) rather than doing the fixing yourself.
2. I kept the manual info fairly high level, with usage details saved for the roxygen (with a separate nimbleQuad PR forthcoming). 
3. We also have the math technical details in an [Overleaf doc of Paul's](https://www.overleaf.com/project/6470dedc771053dda37c7c66). We could move some of that into the manual or create a separate technical vignette if we think that would be useful. Given how hard it was to find the details of what is actually implemented in INLA, it would probably be good for us to provide more public info.  
4. Should we have a new separate chapter for Laplace+nested and leave SMC and MCEM in Other Algos? I lean toward a separate chapter but right now it is all one chapter.


paciorek:
Hmm, I thought by merging devel into this branch that the Files Changed area would only show the stuff changed in the branch for the nested approx-related stuff. But that doesn't seem to be the case. Maybe because I merged devel into the branch after making the PR.

In any event, the best thing to do is probably simply be looking at the `UserManual/src/chapter_otherAlgorithms.Rmd` file or the corresponding chapter in the compiled pdf/html of the manual.

perrydv:
@paciorek I made a few minor tweaks in `chapter_AD.Rmd`. We could take a fresh look later at whether a running example from early on would help or whether with the additional content elsewhere it works without that.

I think splitting "other algorithms" into multiple chapters is a good idea. There is plenty of critical mass to do so and clear things to group together.

PR #1575, opened 2025-07-18
Automate assignment of dmnormAD and add dmnormAD tests

paciorek:
Also this fixes a bug that `getParam` for dmnormAD was only returning the upper triangle for the precision.

This PR changes `dmnorm` to `dmnormAD` whenever model derivs are enabled, except if the `cholesky` parameterization is enabled (in which case a warning is printed).

It adds some testing of dmnormAD that is mostly copying tests that use `dmnorm` from our other test files and setting the nimble option to build model derivs generally so that the tests are run with `dmnormAD` being used.

I'd like for @perrydv to take a look before I merge this into `fix-cppad-memory-explosion` to make sure he is ok with how I have set up the automation.


paciorek:
@perrydv and I discussed and are comfortable moving forward now that there is an option for users to use `dmnorm` (not `dmnormAD` even when derivs are being built.

Issue #1564, opened 2025-06-28
bug in `getConditionallyIndependentSets` in somewhat-involved case with particular order for `nodes` argument


paciorek:
With a particular order of the latents, this incorrectly splits the random effects into two sets. This is arising from nested approximation where we generally put fixed effects with the latents instead of into params.

```
code <- nimbleCode({
        for(i in 1:5) {
            y[i] ~ dnorm(b1*x[i] + mu[i], sd = exp(b2*x[i]+ mu2[i]))
            mu[i] ~ dnorm(0, tau)
            mu2[i] ~ dnorm(0, tau2)
        }
        b1~dflat()
        b2 ~dflat()
        tau ~ dhalfflat()
        tau2~dhalfflat()
    })
m <- nimbleModel(code, data = list(y = rnorm(5)))
given <- c('tau','tau2',paste0('y[', 1:5, ']'))

## Correct
latents <- c('b1','b2',paste0('mu[', 1:5, ']'),paste0('mu2[', 1:5, ']'))
m$getConditionallyIndependentSets(nodes = latents, givenNodes = given, unknownAsGiven = TRUE)

## Incorrect with different order for the latents: `mu2[1]` split out into its own set
latents <- c(paste0('mu[', 1:5, ']'),paste0('mu2[', 1:5, ']'),'b1','b2')
m$getConditionallyIndependentSets(nodes = latents, givenNodes = given, unknownAsGiven = TRUE)
```

```
[[1]]
 [1] "b1"     "b2"     "mu[1]"  "mu[2]"  "mu[3]"  "mu[4]"  "mu[5]"  "mu2[2]"
 [9] "mu2[3]" "mu2[4]" "mu2[5]"

[[2]]
[1] "mu2[1]"
```

paciorek:
I think this might be another manifestation of the bug, but I haven't looked at all. Just recording for when I come back to this.

```
code <- nimbleCode({
   for(i in 1:n) {
     y[i] ~ dbern(p[i])
     logit(p[i]) <- beta0 + beta1 * x[i] + sigma*b[i]
     b[i] ~ dnorm(0, 1)
   }
   mu ~ dflat()
   sigma ~ dunif(0, 100)
   beta0 ~ dflat()
   beta1 ~ dflat()
})
n <- 30
x <- rnorm(n)
y <- rbinom(n, 1, 0.5)

model <- nimbleModel(code, constants = list(n = n), data = list(y = y, x = x))
[...]

PR #1576, opened 2025-07-22
Candidate fix to #1564 (but in getConditionallyIndependentSets)

perrydv:
I am throwing this up as a candidate fix. It fixes Issue #1564 and it passes test-setupMargNodes.R. 

We should probably tests in (at least) nimbleQuad (in development at this time) and nimbleSMC,

If this seems to work, I will document more about the change and add a regression test.

The basic idea is: don't mark deteministic nodes as touched during getConditionallyIndependent sets traversals of the graph. The reason is that as we are iterating over a set of parent (or presumably in some cases child) nodes, the recursions could pass through a deterministic node that then needs to be passed through by a later parent (or child) node of the original iterations. But if it has been "touched" (tagged as done), then it blocks activity. But in the case of a deterministic node, this is not really correct. So it takes a particular setup of nodes to trigger the issue. I think it is actually correct to never  tag deterministic nodes, since the algorithm is fundamentally about stochastic nodes.

### DIFF
--- a/packages/nimble/inst/CppCode/nimbleGraph.cpp
+++ b/packages/nimble/inst/CppCode/nimbleGraph.cpp
@@ -856,7 +856,8 @@ void nimbleGraph::exploreUp(vector<int> &deps,
         exploreDown(deps, thisParentCgraphID,
                     isGivenVec, isLatentVec, unknownsAsGiven, recursionDepth + 1);
     }
-    thisParentNode->touched = true; // Touch it now in case it was deterministic and hence not touched above
+    // Experimental: Commenting out this line to try fixing Issue #1564
+    // thisParentNode->touched = true; // Touch it now in case it was deterministic and hence not touched above
   }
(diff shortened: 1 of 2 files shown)


paciorek:
For discussion - it's not clear to me why tests would go into `nimbleQuad` given this is really about cond. indep. sets.

I've added a (the?) regression test for the example in #1564.

I'll note that `setupMargNodes` is mostly about Laplace-related stuff but is used in MCEM too. So my plan was to keep it in nimble and not move to nimbleQuad.

paciorek:
This passed testing previously; I just canceled the tests to avoid a bunch of test computation when adding the simple regression test.

paciorek:
This passes the  (limited new) tests in `nimbleQuad` and the `nimbleSMC` tests. I'm going to merge in.

(Actually a test-quadrature.R test is failing, but that was apparently failing before.)

PR #1581, opened 2025-07-26
Fix various issues in initializing nimOptim, plus use of Laplace updateSettings

paciorek:
See NCT 571 for a full description of the bug and the necessary changes.

### DIFF
--- a/packages/nimble/R/Laplace.R
+++ b/packages/nimble/R/Laplace.R
@@ -323,6 +323,7 @@ buildOneAGHQuad1D <- nimbleFunction(
       ##   transMethod <<- gridType
       ## }
       if(replace_optimControl) {
+        optimControl$fnscale <- -1
         optimControl_ <<- optimControl
       }
     },
@@ -1163,6 +1164,7 @@ buildOneAGHQuad <- nimbleFunction(
         transMethod <<- gridType
       }
       if(replace_optimControl) {
+        optimControl$fnscale <- -1   
         optimControl_ <<- optimControl
       }
     },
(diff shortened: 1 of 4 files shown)


paciorek:
Note that in Laplace we should only set fnscale to -1 if it is set to the default value of 1, so as to allow users to set it to something different (e.g., -10). I've done that in `nimbleQuad`, but not here.

perrydv:
@paciorek Thanks for catching the needs for `as.numeric` and `working_maxit`.

My idea with allowing replacement of the optimControlList was that it is an advanced thing to do, and so if a user does it, they must attend to all elements of the optimControlList. I did not want to put in the effort to somehow allow changing just individual elements, and I suspect it would get clunky. So it could be little discordant to stick in `fnscale=-1`, i.e. to force one element to one value that we think must be right. While I can't imagine a case where it should be positive, could they experiment with `fnscale = -0.5` or `-10`? Instead of forcing it to `-1`, would we want to check if it's non-negative and throw an error or warning if so? Of course, it is always possible to intervene in control settings by providing one's own optimizer (which could call a standard optimizer but change control values or other settings). `-1` is likely safe and helpful in many or most cases, so we could stick with it.

paciorek:
@perrydv sorry to be confusing about the `fnscale` business. Here in `nimble`, I just stick in `-1`, which as you say is too prescriptive. But in `nimbleQuad` I only stick in `-1` if the value is `1`. And the code in `nimble` will disappear when I delete the Laplace code for the 1.4.0 release. So I think we are now on the same page.

PR #1583, opened 2025-08-01
Allow nimbleCode to combine multiple code chunks

kenkellner:
This PR modifies `nimbleCode` to take an arbitrary number of arguments instead of a single argument. Each argument can be either bracketed code (how it currently works) or objects containing bracketed code. The code is then combined (#1570). For example, the following would now be possible:

```r
c1 <- nimbleCode({
  x ~ dnorm(0, sd = 1)
})

c2 <- nimbleCode({
  y ~ dunif(0, 1)
})

nimbleCode({
  z ~ dnorm(0, sd = 1)
}, c1, c2)
```

with a result of

```
{
    z ~ dnorm(0, sd = 1)
    x ~ dnorm(0, sd = 1)
    y ~ dunif(0, 1)
}
```

There is error trapping to catch un-bracked code and objects that do not actually contain code.

This requires a bit of complexity in the implementation because we need to avoid converting the arguments `...` to a list at any point, because if `...` contains code, R will try to evaluate that code and error. My solution was to immeadiatly use `substitute` on `...` to avoid this, and then detect what type of input each argument is. If it's an object containing code, we extract the code with `eval`. While not ideal to need `eval` least the current approach avoids messing with environments.

I've tried to do the minimum processing in the case when there's a single argument directly providing code, to minimize the chance of causing errors when using `nimbleCode` in the established way. Adding this functionality still makes me a bit nervous though considering the importance of `nimbleCode` in the ecosystem.

### DIFF
--- a/packages/nimble/R/BUGS_readBUGS.R
+++ b/packages/nimble/R/BUGS_readBUGS.R
@@ -157,41 +157,46 @@ nimbleModel <- function(code,
 #'     mu ~ dnorm(0, sd = prior_sd)
 #' })
 #'
+#' code_new <- nimbleCode({
+#'     prior_sd ~ dhalfflat()
+#' })
+#'
 #' # Combine multiple previously saved code objects
-#' code2 <- nimbleCode(code, code)
+#' code2 <- nimbleCode(code, code_new)
 #'
 #' # Combine code and previously saved code objects
 #' code3 <- nimbleCode({
 #'    y ~ dnorm(mu, sd = 1)
-#' }, code, code2)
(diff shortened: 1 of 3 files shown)


paciorek:
LGTM, and thanks, @kenkellner 

@perrydv @danielturek do either of you want to take a look before I merge this in?

(side note I am canceling tests given trivial, non-code changes I made)

danielturek:
Thank you, unfortunately I’m away from a computer this week, and won’t be
able to review it.

If it combines quoted code chunks in curly braces, it’s ok with me.

PR #1590, opened 2025-09-04
improve efficiency of make_modelDerivsInfo

paciorek:
This extends the work in PR #1435 to keep node info in `makeModelDerivsInfo{,_impl}` at the node level until the final return statement. As seen in a moderate size spatial GLMM, some of the steps can be quite slow and use a lot of memory, e.g., `model$expandNodeNames(c(neededParentNodes, nonWrtStochCalcNodes))`.


paciorek:
@perrydv did you want to review my changes in this PR to `makeModelDerivsInfo{,_impl}`?

I'd like to merge this and `setData-omit-RHSonly` into devel shortly.

paciorek:
Perry and I decided we should just merge this in to clean things up.

paciorek:
So my switch to working at the node level in `makeModelDerivsInfo` introduced a bug in the case that `wrtNodes` includes a scalar element of a non-scalar node. 

I think what we need is that in that case we work on the element level and otherwise we work on the node level (the latter for efficiency). Also, we need to make sure `nimDerivsInfoClass_init_impl` interacts correctly with `makeModelDerivsInfo_impl`, since `nimDerivsInfoClass_init_impl` doesn't go through `makeModelDerivsInfo`.

PR #1596, opened 2025-11-06
Support for dgamma - dcar_normal conjugacy for tau

danielturek:
*Not to be merged*.

Just want to see what testing does at this point.

It would be a small miracle if this ever gets merged, due to implicit assumptions made in the conjugacy system regarding size / dimensionality, that have come to light.


danielturek:
Noting in advance that some upcoming pushes will cause testing to fail.  WIP.

danielturek:
This PR makes conjugacy processing more general.  Dependent nodes of target conjugate nodes no longer must have the same sizes as the target node.  Rather, dependent nodes (and, their parameters) can have different sizes (and dimensions) from the target node.

In particular, this suports the conjugacy between a gamma prior for `tau`, when `tau` is used as the precision parameter in `dcar_normal` distributions. Support for this relationship is also added, as well as a test of the asymptotic correct distribution.

@paciorek @perrydv I hope to merge this.  But it's a minor under-hauling of conjugacy processing. I know it's minor in the scheme of things.  I welcome any feedback.  I think it's sound.

danielturek:
@paciorek I'm hoping this can also be included in the upcoming release.

paciorek:
@danielturek I'm happy to see this included in the release. I haven't reviewed the implementation details of the conjugacy processing (I did look at the conj contributions code), but if you're confident in them, I'm happy to merge this in, particularly given I think the testing system is probably reasonably good when it comes to conjugacy processing.

danielturek:
@paciorek Sounds good to me also.  I'm going to merge this in.

PR #1602, opened 2025-11-26
add iterationTotal to MCMC

danielturek:
Addresses NCT issue 591.

This PR adds an `iterationTotal` counter to the core MCMC run method, which tracks the total number of MCMC iterations executed (since the last call with reset = TRUE).  This produces a more logical schedule of saving samples, in particular when the thin interval is not 1 and an MCMC run is continued using reset=FALSE.

See what testing turns up.  I think it's all correct, but I only did very modest testing myself.


danielturek:
@paciorek Is there a goal of getting testing to start passing again on core nimble (devel branch)?

paciorek:
I had been planning on merging the `no-laplace` branch (which excises laplace) in and then moving toward release from there. I'd rather not diagnose new testing failures until laplace is excised. So I guess I will merge that in and then see where things stand.

paciorek:
Ok, this is passing now.

danielturek:
@paciorek @perrydv I'm ok with this getting merged.  I think it's solid.  That said, it does touch things at the very base.

@perrydv These changes should simplify matters for using the MCMC within the MCEM.

PR #1605, opened 2025-12-18
updates for correct dimension

danielturek:
WIP

### DIFF
--- a/packages/nimble/R/MCMC_conjugacy.R
+++ b/packages/nimble/R/MCMC_conjugacy.R
@@ -570,17 +570,27 @@ conjugacyClass <- setRefClass(
                 }, list(DEP_NAMES        = as.name(paste0(  'dep_', distLinkName, '_nodeNames')),
                         N_DEP            = as.name(paste0('N_dep_', distLinkName)),
                         DEP_CONTROL_NAME = as.name(paste0(  'dep_', distLinkName))))
+                forSizeDetermination <- character()
+                ## need to include dependent node 'value', if multivariate or if doing dependentScreen:
+                ## Dec 2025
+                if(distDim > 0 || (getNimbleOption('allowDynamicIndexing') && doDependentScreen)) {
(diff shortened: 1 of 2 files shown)


paciorek:
How critical is this. Are we going to want to put up a 1.4.1 soon or can you direct the user to a GH branch?

danielturek:
@paciorek I was planning just a providing a patched branch for him - although, since he's using it in a package, that solution might feel lacking to him...

But that said, I don't think any/many other users will be experiencing any problems with this (fingers crossed), it's a relative corner case.

And my apologies about missing these details on the first pass.

danielturek:
This fix addresses some prior (and previously safe) assumptions that were made in the original conjugacy processing.  But, with the addition of #1596, which actually increases and generalizes the conjugacy system, those assumptions (previous being made) were no longer correct.

Specifically, the old conjugacy size processing system wasn't able to account for dependent nodes (of the target conjugate node) which themselves had multivariate parameters *with a different size from the dependent node itself*.  The motivating case of this was addition of the conjugacy for the `tau` parameter of a `dcar_normal` distribution, when `tau` follows a gamma prior.  The issue was that the (multivariate) `dcar_normal` dependent node(s) have length say `N`, but the (multivariable) parameters of these dependent `dcar_normal` node(s), say the multivariate `adj` parameter, have length `L`, which is different.  The old conjugacy system wasn't developed enough to handle that.

#1596 fixed the issue above, but also uncovered some assumptions that were made in the original conjugacy system.  Specifically, this had to do with the determination of the sizes of the `coeff` and `offset` terms for linear / multiplicative links.  Previously, it was sufficient to assume the sizes of the `coeff` and `offset` terms had the same size as the multivariable **dependent node itself**, since that was always (assumed) to be the same as the multivariable parameters of the dependent node.  But now, with the update in #1596, we need to correctly get the sizes of the `coeff` and `offset` terms from the size(s) of the multivariable parameters themselves.  That's what is accomplished in this PR.

Issue #1608, opened 2026-01-07
Impossible to change the Thin parameter in a $run for MCMC with version 1.4.0 of Nimble


gosselinf:
Dear all,

I am unsure this is the right place to send my message but here it is. 

As the maintainer of the runMCMCbtadjust package that relies partly on Nimble, I was surprised that the  runMCMC_btadjust function no longer works with version 1.4.0 of Nimble due to the impossibility to change the thin parameter in an update of the MCMC with reset=FALSE. The error message is:
"cannot alter the value of thin, when using reset = FALSE. Called from: CModelMCMC[[i]]$run(niter, thin = thin, progressBar = TRUE, reset = FALSE, resetMV = as.logical(max(control.MCMC$resetMV, resetMV.temporary)))"

My question is whether there is a workaround that I could take into account in runMCMCbtadjust  and whether this is a definitive change in Nimble. 

Sincerely,

Frédéric

danielturek:
@gosselinf Thanks for the report.  Certainly apologize for breaking your existing code.

@perrydv This warrants some discussion, about desired behaviour.

Referring to changes made in #1602 , which were in response to [NCT issue 591](https://github.com/perrydv/nimbleCoreTeam/issues/591) ("MCMC$run with reset=FALSE does not match results from a single run if previous niter was not a multiple of thin"

perrydv:
@danielturek I am not giving this a thorough look right now, but my quick thought is that the issue of matching results from one longer run vs multiple shorter runs (which we made more robust with #1602 ) should be moot in the case of switching `thin`, because that can only be done by using multiple runs anyway. I guess one would still want to be able to easily align the results of a longer run with `thin=1` (or something small) vs multiple smaller runs with larger `thin` intervals, for example, but that does seem to get into more remote cases. Maybe instead of a hard stop, we can check if it matters (i.e. if there were "extra" iterations, beyond the last saved iteration, in a previous run) and if so just emit a message instead of a stop? I'll let you suggest any work-arounds that might involve direct manipulation of the MCMC object.

gosselinf:
Dear Daniel & Perry,

I thank you very much for your replies.

What you have done in [#1602](https://github.com/nimble-dev/nimble/pull/1602) seems interesting and for me should not preclude from changhing thin with reset=FALSE. Actually in runMCMCbtadjust, the preferred method to update thin is to make it a multiple (multiplication by an integer) of the previous thin, which would pose no pb according to me.

[...]

PR #1609, opened 2026-01-14
safe changing of MCMC thinning intervals when reset=FALSE

danielturek:
WIP, do not merge.

### DIFF
--- a/packages/nimble/R/MCMC_build.R
+++ b/packages/nimble/R/MCMC_build.R
@@ -245,8 +245,7 @@ buildMCMC <- nimbleFunction(
             nburnin_extraWAIC <- 0
         }
         iterationTotal <- 0
-        iterUntilSave  <- 0
-        iterUntilSave2 <- 0
+        iterUntilSaveVec <- c(0, 0)
         firstRun <- TRUE
         setupOutputs(derivedTypes)
     },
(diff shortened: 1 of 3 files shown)


packages/nimble/R/MCMC_build.R
```diff
@@ -300,8 +300,12 @@ buildMCMC <- nimbleFunction(
             ## when reset = FALSE, only permit changing the value of
             ## thin when iterUntilSave == thinToUseVec[1], and thin2 when iterUntilSave2 == thinToUseVec[2]
             ## (which means saving on the previous thinning interval just took place)
-            if((thin  != -1) & (thin  != thinToUseVec[1]) & (iterUntilSave  != thinToUseVec[1]))  stop('can only change value of thin after a complete thinning interval, when using reset = FALSE.')
-            if((thin2 != -1) & (thin2 != thinToUseVec[2]) & (iterUntilSave2 != thinToUseVec[2]))  stop('can only change value of thin2 after a complete thinning interval, when using reset = FALSE.')
+            if((thin  != -1) & (thin  != thinToUseVec[1]) & (iterUntilSave  != thinToUseVec[1]))
+                ##stop('can only change value of thin after a complete thinning interval, when using reset = FALSE.')
+                cat('Warning: value of thin is being changed, in the midst of a thinning interval.\n')
```
paciorek: @danielturek I'd generally like us to use `messageIfVerbose` unless there is a specific reason not to.

danielturek: @paciorek I don't believe `messageIfVerbose` is supported in DSL code - am I missing something?

paciorek: Right, sorry.

When I submitted nimbleQuad, CRAN wanted all calls to `cat` and `print` behind verbosity checking. I added that in even for nimbleFunction code rather than get into our compilation process.

For nimble itself when a new version is submitted, I don't think they do the kind of fine-grained checking as they did on initial submission of nimbleQuad, so this is probably fine as is, and of course there are lots of other uses of print/cat in our nimbleFunction code.

packages/nimble/R/MCMC_build.R
```diff
@@ -289,25 +289,40 @@ buildMCMC <- nimbleFunction(
...
-            if((thin  != -1) & (thin  != thinToUseVec[1]))   stop('cannot alter the value of thin, when using reset = FALSE.')
-            if((thin2 != -1) & (thin2 != thinToUseVec[2]))   stop('cannot alter the value of thin2, when using reset = FALSE.')
+            if(nburnin != 0)   stop('cannot specify nburnin when using reset = FALSE.')
+            ## now allowing changing of thin and thin2, when reset = FALSE; Jan 2026.
+            ## when reset = FALSE and a new value of thin (or thin2) is provided,
+            ## issue a warning if the previous MCMC run didn't exactly finish a thinning/save cycle.
+            ## this is checked using iterUntilSaveVec[1] == thinToUseVec[1] or iterUntilSaveVec[2] == thinToUseVec[2],
+            ## which means saving on the previous thinning interval just took place.
+            if((thin  != -1) & (thin  != thinToUseVec[1])) {
+                if((iterUntilSaveVec[1] != thinToUseVec[1]) & warnUnfinishedThinningInterval)
```
paciorek: We should use `&&`. CRAN now says that `&` is only for vectorized comparisons.

danielturek:
See discussions in issue #1608

PR #1612, opened 2026-01-26
MCMC Sampling of partially-observed multivariate normal nodes

danielturek:
Addresses NCT issue 296.

This PR adds support for MCMC sampling of partially-observed multivariate normal nodes.  That is, multivariate-distributed nodes, where some of the dimensions are observed data, and some of the dimensions are unobserved (latent).  The `partial_mvn` sampler essentially operates by assigning univariate RW samplers to the (individual) unobserved dimensions of partially-observed (and therefore also partially-unobserved) MVN-distributed nodes.

More specifically, several cases for sampling the unobserved dimensions are handled:
- Any latent dimensions which *have no stochastic dependencies* are assigned `posterior_predictive` samplers.
- If `multivariateNodesAsScalars = TRUE`, then all remaining unobserved dimensions are assigned univariate `RW` samplers.
- If `multivariateNodesAsScalars = FALSE`, then all remaining unobserved dimensions are assigned a single multivariate `RW_block`.

A variety of test cases are also added.


danielturek:
Testing passed.  I welcome any thoughts or review from @paciorek or @perrydv

paciorek:
So do we think the most common use case would be that `y[1:n]` is a leaf node containing some data and some non-data elements and therefore the non-data are predictive nodes? In that case it looks like we assign a RW_block sampler. So @danielturek I'm not sure what you meant by "Any latent dimensions which have no stochastic dependencies are assigned posterior_predictive samplers." More substantively, should we work out the true MVN conditional distribution in that situation and sample from the true predictive?  If we do think this is a common case, then I think it might be worth it. (And I would be happy to lend a hand if we decide to do it.)

danielturek:
@paciorek Please see the addition of the `partial_mvn_pp` sampler ("partial MVN posterior predictive" sampler), which is used internally in the `partial_mvn` sampler.

When one or more dimensions (from among the unobserved dimensions) of a partially-observed MVN node themselves have no data dependencies, then a `partial_mvn_pp` sampler is now assigned to these dimension(s) - instead of either `RW` or `RW_block` samplers, as before.  This new `partial_mvn_pp` generates draws directly from the conditional MVN distribution.

@paciorek A second set of eyes would be appreciated.  There might also be some efficiency improvements with the linear algebra within the `partial_mvn_pp` sampler, which could be improved upon from the present implementation.

packages/nimble/R/MCMC_samplers.R
```diff
@@ -3456,7 +3456,113 @@ sampler_barker <- nimbleFunction(
...
+#' @export
+sampler_partial_mvn <- nimbleFunction(
+    name = 'sampler_partial_mvn',
+    contains = sampler_BASE,
+    setup = function(model, mvSaved, target, control) {
+        ## control list extraction
+        multivariateNodesAsScalars <- extractControlElement(control, 'multivariateNodesAsScalars', default = getNimbleOption('MCMCmultivariateNodesAsScalars'))
+        ## node list generation
+        targetAsScalar <- model$expandNodeNames(target, returnScalarComponents = TRUE)
+        isdataBool <- sapply(targetAsScalar, function(targetAsScalar) eval(parse(text = targetAsScalar)[[1]], envir = model$isDataEnv))
```
paciorek: can we use `isDataBool`?

packages/nimble/R/MCMC_samplers.R
```diff
@@ -3456,7 +3456,113 @@ sampler_barker <- nimbleFunction(
...
+        mu[,1] <<- model$getParam(mvNode, 'mean')
+        Sigma  <<- model$getParam(mvNode, 'cov' )
+        mu1[,1] <<- mu[ind1,1]
+        mu2[,1] <<- mu[ind2,1]
+        Sigma11[1:n1,1:n1] <<- Sigma[ind1, ind1]
+        Sigma12[1:n1,1:n2] <<- Sigma[ind1, ind2]
+        Sigma21[1:n2,1:n1] <<- Sigma[ind2, ind1]
+        Sigma22[1:n2,1:n2] <<- Sigma[ind2, ind2]
+        Sigma12 <<- Sigma12 %*% inverse(Sigma22)
+        mu1[,1] <<- mu1[,1] + (Sigma12 %*% (values(model,given) - mu2[,1]))[,1]
```
paciorek: We should get cholesky of Sigma22 and then do `backsolve(ch, forwardsolve(t(ch), <item>)` in lines 3553 and 3554.

packages/nimble/R/MCMC_samplers.R
```diff
@@ -3456,7 +3456,113 @@ sampler_barker <- nimbleFunction(
...
+    name = 'sampler_partial_mvn',
+    contains = sampler_BASE,
+    setup = function(model, mvSaved, target, control) {
+        ## control list extraction
+        multivariateNodesAsScalars <- extractControlElement(control, 'multivariateNodesAsScalars', default = getNimbleOption('MCMCmultivariateNodesAsScalars'))
+        ## node list generation
+        targetAsScalar <- model$expandNodeNames(target, returnScalarComponents = TRUE)
+        isdataBool <- sapply(targetAsScalar, function(targetAsScalar) eval(parse(text = targetAsScalar)[[1]], envir = model$isDataEnv))
+        targetNonDataComponents <- targetAsScalar[!isdataBool]
+        predictiveBool <- sapply(targetNonDataComponents, function(n) length(model$getDependencies(n, self=FALSE, downstream=TRUE, dataOnly=TRUE)) == 0)
```
paciorek: getDeps takes ~250 microseconds. Not terrible but we might want to first check if the entire MVN vector has any dependencies non-data deps and only if it does do the more detailed chekcing on a per-element basis.

packages/nimble/R/MCMC_samplers.R
```diff
@@ -3456,7 +3456,113 @@ sampler_barker <- nimbleFunction(
...
+        mu2 <- array(0, c(n2, 1))
+        Sigma   <- array(0, c(n,  n ))
+        Sigma11 <- array(0, c(n1, n1))
+        Sigma12 <- array(0, c(n1, n2))
+        Sigma21 <- array(0, c(n2, n1))
+        Sigma22 <- array(0, c(n2, n2))
+        ind1 <- match(target, mvNodeComponents)
+        ind2 <- match(given,  mvNodeComponents)
+        ## checks
+        if(length(mvNode) != 1)                         stop('something went wrong in sampler_partial_mvn_pp setup')
```
paciorek: can we put informative messages or reword as "unexpected error"? I think "something went wrong" sounds too colloquial. I know these should never be invoked, so ok with me to just reword without further information.

packages/nimble/R/MCMC_samplers.R
```diff
@@ -3456,7 +3456,113 @@ sampler_barker <- nimbleFunction(
...
+                } else {
+                    samplerList[[1]] <- sampler_RW_block(model, mvSaved, targetNonDataNP, control)
+                }
+            }
+        }
+        if(length(targetNonDataPP) > 0) {
+            samplerList[[ length(samplerList)+1 ]] <- sampler_partial_mvn_pp(model, mvSaved, targetNonDataPP)
+        }
+        ## checks
+        if(model$getDistribution(target) != 'dmnorm')   stop(paste0('The node ', target, ' is parially observed. NIMBLE only handles this case for multivariate normal distibutions.'))
```
paciorek: `stop(paste0(...))` is not needed. We can just use `stop(...)`

paciorek (changes requested):
thanks, @danielturek . I took a reasonably careful look though I didn't look too carefully at the tests. In addition to my line-specific comments (some of them niggly - sorry!):

- We should discuss whether to cache the pieces of the conditional mvn, in particular choleskies if we know that the covariance is fixed. We could leave this for a future improvement, though if you have momentum on all this, perhaps we do it now?
- I'd like us to check if partial_mvn_pp is correctly assigned, so how about we build the MCMC in some of the tests and check the internal sampler assignments?
- I'm not sure about whether we should hide this behind an option. Perhaps it's fine if we don't, though if we add the caching and are less certain about that, then maybe we should.
- Can we check `MCMCuseBarkerAsDefaultMV` and use Barker instead of sampler_RW_block if it is on?
- Also I think we should have a test that checks correctness of the partial_mvn_pp sampler in a basic case.

I'm happy to help with any of this if you like.

packages/nimble/R/MCMC_samplers.R
danielturek: @paciorek Would you please do me a favor: leave these current lines (but commented out), and also make this change you suggest.  I will take a careful look at the old code and the changed code, to understand it better.

packages/nimble/R/MCMC_samplers.R
danielturek: @paciorek I always liked the hometown feel of "something went wrong", but fine.

danielturek:
@paciorek Thanks for the careful review.

Repeating the request from above, my first preference would be to make the `backsolve` and `forwardsolve` changes myself.  But that would take a bit more time for me to understand and make sure I do it right, and I'm out of time for today.  If you are willing to make this change yourself, and leave the original (current) code there *commented out*, I would very much appreciate that, and I'll take a very careful look later.

I'd be happy to discuss strategies for "caching" the components of the covariance matrix.  But I think it would have to be done with some thought and some caution. *Even if* the covariance node has no parent nodes (stochastic or otherwise), the contents of the node can still be modified (regardless if the value is provided by constants, data, or initial values).  So I can imagine a few ways we might try to do this, for example:

"If the covariance node has no parent nodes, then we have a indicator internal to the sampler (say `firstRun`) which indicates if this is the first run of the MCMC (initially TRUE), and when `firstRun` is TRUE, we pull the value of the covariance from the model, and perform the factorizations etc, and cache those values in the sampler.  Then, on subsequent runs of the sampler, when `firstRun` is FALSE, we just used the cached values.  However, when the `reset` method is called, indicating starting a new run of the MCMC, we set `firstRun` back to TRUE, in case the user changed the value of the covariance (inside the model) in between MCMC runs.  (and, we also take it on blind faith that a user didn't change the value of the covariance, then continue an MCMC run using `reset = FALSE`, which would make the entire MCMC a little questionable, and make the subsequent MCMC results incorrect in some sense)."

This is a tricky one, unless you have some better ideas that I'm not imagining.

In the case of assigning the `barker` sampler in place of the `RW_block` sampler, need we do anything to enable derivatives in the model?

@paciorek Yes, please go ahead and make / add any tests for any of the following:

- any testing with changing the nimble option `MCMCuseBarkerAsDefaultMV` to `TRUE`, and correct assignment and operation of the `barker` sampler.
- check if `partial_mvn_pp` is correctly assigned, build the MCMC in some of the tests and check the internal sampler assignments
- test that checks correctness of the `partial_mvn_pp` sampler in a basic case.

paciorek:
Ok I pushed a commit that has the implementation of the Cholesky of Sigma22. I didn't run tests (though CI will of course) or check correctness yet.

As far as Barker, yes, that's correct the model would need to have derivs built. But I don't think anything is different with using Barker here than in our default MCMC config if that Barker option is set.

Regarding caching, I will comment on that in a moment below.

I will try to work on some of the other things this afternoon as well. Or tomorrow.

Unless Frederic indicates urgency in response to my email I am now planning that we include this in 1.4.1 and am thinking I will push this to CRAN on Tuesday. I will be out of town over the long weekend.

paciorek:
Regarding caching, I think the case that we can safely use cached values is if the covariance is RHS-only. If it has no parent nodes it could still be being sampled in an MCMC (e.g., if it is `dwishart` or `dlkj` with fixed prior parameters). 

Two other thoughts:
- we could do similar machinations with the mean so as to save recalculation of the conditional mean, which involves matrix-vector manipulations. Clearly less important but if we are working out the caching perhaps we might as well.
- we could in cases where we don't know if the values might have changed, actually check if the cached covariance (and perhaps mean) are the same as the current values from `getParam`. Then if they are, we can avoid some calculations. Checking should be O(n^2), while the saved calculations are O(n^3) (dominated by the Cholesky of either Sigma11 or Sigma22). But in cases where the covariance is changing every iteration then we are doing a bunch of useless checking. So I think it probably doesn't make sense to do this.

I guess if my reasoning about checking for RHS-only is correct, I propose we combine that with your sketch of using `firstRun`. And I think we might as well check if the mean is RHS-only too.

packages/nimble/R/MCMC_samplers.R
paciorek: Done.

paciorek:
I'm working on some additional tests.

@danielturek I don't understand the logic in "Values change at same rows for unobserved parts of partially observed node when multivariateNodesAsScalars = FALSE" test. Based on the model is uses the `_pp` sampler, so `multivariateNodesAsScalars` is not involved. Don't you want a model where `y` is intermediate?

danielturek:
@paciorek That test predated the addition of the `partial_mvn_pp` sampler - meaning, that test existed before today.  The idea behind that test was to confirm that all unobserved dimensions of the MVN node change *in unison* (or, none of them change) on each MCMC iteration, since when `multivariateNodesAsScalars = FALSE`, the unobserved dimensions should have a `RW_block` sampler assigned (and hence all change simultaneously, or all remain unchanged, on each MCMC iteration).  This is of course now moot, since for the model in this test the `partial_mvn_pp` now updates all unobserved dimensions on *every* MCMC iteration, so they will all (always) change on each iteration.  I'm not wedded to this test, there's more of a backstory here than I'll get into, and I would be happy to have this test be removed, or modified, if you're so inclined.

danielturek:
@paciorek I worked through the linear algebra.  I imagine it's simple to you, but this was helpful for me, and it was nice to see especially how the `forwardsolve` can be used to calculate the conditional posterior for `Sigma`, leveraging the transpose-relationship between `Sigma12` and `Sigma21`.  I now have a much better understanding of the uses of `forwardsolve` and `backsolve`.  I also fixed an extremely minor dimension mistake.

paciorek:
Ok, so my new code had some issues with dimensions that I just fixed. Results are now the same as Daniel's implementation.

I added some tests, but the one I still want to add is one to check correctness of the _pp samples. I'll do that tomorrow morning.

I'll let you decide what to do about the 'moot' test.

paciorek:
I think we stepped on each other's commits a bit, but I think everything is ok now.

danielturek:
@paciorek `expect_success` does not operate as one might imagine.  Use `expect_no_error` instead.

danielturek:
@paciorek I took a go at adding the "caching" of the values derived from the MVN covariance and mean parameters.

The criterion I used for either of them being considered "constant" is: the (covariance/mean) parameter has no upstream stochastic parent nodes (including itself).  I think this covers all possible cases, but please let me know if that doesn't seem right.

I also added the notion of `firstRun` to the sampler.  On the first run, the `partial_mvn_pp` sampler *always* pulls values of the MVN covariance and mean from the model, and uses them to recalculate the parameters of the conditional MVN posterior.  After the first run, the relevant values are only updated if either (covariance or mean) is deemed non-constant.

A second set of eyes would be appreciated.

paciorek:
I'm adding a test of mcmc correctness.

Ok, I see you just added some caching. I'll take a look.

paciorek:
If mu is constant but sigma is not, then we still need to recalculate 

```
mu1[,1] <<- mu1[,1] + (Sigma12 %*% backsolve(Sigma22, forwardsolve(t(Sigma22), values(model,given) - mu2[,1])))[,1]
```

no? So I think we need to execute that line if `!muConst | !sgConst | firstRun`.

Also, I think `if(model$getDistribution(mvNode) != 'dmnorm')   stop('unexpected error in sampler_partial_mvn_pp')` should be at the start, as otherwise one might try to get the `mean` or `cholesky` for another distribution and get an untrapped error.

danielturek:
@paciorek Nice catch about the calculation of `mu`.  I pushed changes just now, I hope I'm not over-complicating it, but you'll see what I did.

As for the "checks", they'll largely symbolic anyway, since this `partial_mvn_pp` sampler should only be assigned from within the `partial_mvn` sampler.  Given that that, for the purposes of consistency and standards, I'd like to keep the checks at the end.

danielturek:
@paciorek Thanks for the help putting these tests together.

PR #1616, opened 2026-03-24
Update from Rf_findVar and Rf_findVarInFrame while maintaining backward compatibility for R < 4.5.0

### LINKED ISSUE #1615: fix use of non-API calls to R's C code
[CRAN is now flagging this issue](https://cran.r-project.org/web/checks/check_results_nimble.html) with a deadline of April 9.

```
Version: 1.4.1
Check: compiled code
Result: NOTE
  File ‘nimble/libs/nimble.so’:
    Found non-API calls to R: ‘Rf_findVar’, ‘Rf_findVarInFrame’
  
  Compiled code should not call non-API entry points in R.
  
  See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual,
  and section ‘Moving into C API compliance’ for issues with the use of
  non-API entry points.
```

From quick glance at the docs, it looks like we can [replace those non-API calls](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Some-API-replacements-for-non_002dAPI-entry-points-1). The docs say "Use R_getVar or R_getVarEx, added in R 4.5.0. In some cases using eval may suffice. "

[Rcpp discussed this](https://github.com/RcppCore/Rcpp/issues/1421), making the comment about using the old code for older R versions. I guess we need to think about whether we are ok with having nimble 1.4.2 require R>=4.5.0. Probably?

perrydv:
Fixes #1615 I think.

This PR includes R preprocessor checks for R version. For versions >= 4.5.0, uses of `Rf_findVar` and `Rf_findVarInFrame` are replaced with `R_getVar` and `R_getVarEx` with a little additional code. It is possible that testing will turn up some issues because evidently the older API calls were looser in some handling of some arguments, and I don't know for sure if we were relying on relevant cases in any way.

On a basic level (with different details), we followed the idea of [Rcpp](https://github.com/RcppCore/Rcpp/issues/1421) (noted by @paciorek in #1615) to check the R_VERSION in C++. But I'm not 100% sure if this will work across nimble.so and also on-the-fly compilation. For nimble.so (or DLL), the version will be checked on the build machine. For on-the-fly, the version will be checked for the user's machine. I am also not sure whether CRAN checking will complain about *any* appearance of `Rf_findVar` and `Rf_findVarInFrame` even when only for older R versions. At least in this PR I have those isolated to a single place.

### DIFF
--- a/packages/nimble/R/cppDefs_nimbleList.R
+++ b/packages/nimble/R/cppDefs_nimbleList.R
@@ -59,7 +59,7 @@ cppNimbleListClass <- setRefClass('cppNimbleListClass',
 ##   ptrToSmartPtrBase = dynamic_cast<nimSmartPtrBase*>(ptrToSmartPtr);
 ##   ptrToPtr = ptrToSmartPtr->getVoidPtrToRealPtr();
 ##   reinterpret_cast<nfRefClass_84*>(*static_cast<void**>(ptrToPtr))->NO_hw();
- 
+
 ##   PROTECT(SptrToSmartPtrBase = R_MakeExternalPtr(ptrToSmartPtrBase, R_NilValue, R_NilValue));
 ##   PROTECT(SptrToPtr = R_MakeExternalPtr(ptrToPtr, R_NilValue, R_NilValue));
 ##   PROTECT(Sans = Rf_allocVector(VECSXP,2));
@@ -83,11 +83,11 @@ cppNimbleListClass <- setRefClass('cppNimbleListClass',
                                                                       'PROTECT(SptrToSmartPtrBase = R_MakeExternalPtr(ptrToSmartPtrBase, R_NilValue, R_NilValue));',
                                                                       'PROTECT(SptrToPtr = R_MakeExternalPtr(ptrToPtr, R_NilValue, R_NilValue));'))
                                           allocVectorLine <- cppLiteral(paste0('PROTECT(Sans = Rf_allocVector(VECSXP,', 2, '));'))
-                                          
+
                                           packListLines <- cppLiteral(c('SET_VECTOR_ELT(Sans,0,SptrToSmartPtrBase);',
                                                                         'SET_VECTOR_ELT(Sans,1,SptrToPtr);'
                                                                         ))
-                                          
+
                                           codeLines <- substitute({
                                               ## Finalizer registration now happens through nimble's finalizer mapping system.
                                               UNPROTECT(3)
@@ -99,7 +99,7 @@ cppNimbleListClass <- setRefClass('cppNimbleListClass',
                                                                               args = args,
                                                                               code = cppCodeBlock(code = allCode, objectDefs = CBobjectDefs, skipBrackets = TRUE),
                                                                               returnType = cppSEXP(),
-                                                                              externC = TRUE)                                          
+                                                                              externC = TRUE)
                                       },
                                       buildSEXPgenerator = function(finalizer = NULL) {
                                           ## Build a function that will provide a new object and return an external pointer.
@@ -197,7 +197,7 @@ cppNimbleListClass <- setRefClass('cppNimbleListClass',
                                         listElementTable <- symbolTable()
                                         listElementTable$addSymbol(cppSEXP(name = "S_newNimList"))
                                         listElementTable$addSymbol(cppSEXP(name = "S_listName"))
-                                      
+
                                         newListLine[[1]] <- substitute({PROTECT(S_listName <- Rf_allocVector(STRSXP, 1));
                                           SET_STRING_ELT(S_listName, 0, PROTECT(Rf_mkChar(LISTNAME)));},
                                           list(LISTNAME = nimCompProc$nimbleListObj$className))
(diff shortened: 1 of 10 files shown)


perrydv:
Failing on Windows. @paciorek Does this look like any general testing issues you've seen recently? I'm guessing there was a failure compiling the package DLL, but the information is not very clear. My windows setup is stale so I'll have to update or find another option.

paciorek:
No the DLL load failure doesn't look familiar. But we've had lots of such squirrely failures in the prep steps for testing on various OS, that it's possible it's not related to the changes here.

perrydv:
@paciorek I've refreshed my Windows setup and successfully built from this branch and tested `graphStructure` (which happens to be the immediate failure here, but probably not due to its specific tests). So I can't reproduce the Windows problem here. What do you suggest?

paciorek:
You might have also done this, but it looks fine on Winbuilder too. 

I guess we proceed, but if it continues to fail CI testing on Windows that is going to be a pain.

perrydv:
I updated the first comment describing this PR. Based on discussion with @paciorek I am going to merge this based on seeing that the Windows test failure does not occur on my Windows setup or on win-builder.

paciorek:
So this does seem to have removed the `Rf_findVar*` calls from nimble.so. So from what I see in CRAN check output we see currently on the status page, I think that they will be satisfied.

I did see that `Rf_findVar*` still appears in a few places in `src/eigenUsingClasses.cpp` and `src/RcppUtils.cpp`.

I assume that is intentional?

perrydv:
No. Four of the five cpp files in src are copied from CppCode during package building, I think. See configure.ac 215-218. Possibly those files should not be in src at all and were accidentally committed?

paciorek:
Right of course. I guess those are just local files on my end; they are not committed.

paciorek:
@perrydv I am starting on release steps. Our CI only runs a few of the test files on Windows. I don't know how cautious we should be, but what/how many of the test files did you run locally on your Windows VM.

perrydv:
@paciorek I didn't run much. Just a quick check that it wasn't immediately broken. It seems like a CI issue but I'm a bit cautious if we're not sure. The errors did refer to some problem with "/usr/bin/time". Does that ring any bells?

paciorek:
We might be talking past each other somewhat.

1. I agree the CI failures seem like an unrelated CI issue. On that node the core error related to loading nimble.dll not to something with `/usr/bin/time`.
2. Given the CI failures, and also recognizing that even under CI we only run a limited number of our test files on Windows, I was just asking how many of the test files you ran locally. Hopefully enough to test some nimble compilations where the modified API calls would come into play.

perrydv:
Right and I only ran test-graphStructure because that happened to be the one that showed the failure on CI. I will run more tests on my Windows setup.

perrydv:
@paciorek It's been a while since I've stared at test results. Is this what we expect?
```
> test_package("nimble", filter = "^models$")
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Using example in BUGS example directory of the NIMBLE package.
Incorrect number of dimensions in copying
Warning: R object of different size than NimArrDouble. R obj has size 2 but NimArrDbl has size 6.
Warning: R object of different size than NimArrDouble. R obj has size 5 but NimArrDbl has size 3.
Warning: R object of different size than NimArrDouble. R obj has size 8 but NimArrDbl has size 6.
Warning: R object of different size than NimArrDouble. R obj has size 2 but NimArrDbl has size 3.
Warning: R object of different size than NimArrDouble. R obj has size 4 but NimArrDbl has size 6.
[ FAIL 0 | WARN 1 | SKIP 0 | PASS 26665 ]

══ Warnings ══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
── Warning (test-models.R:422:5): test of using dimensions of data when dimension information not available: ──
  [Note] Inconsistent dimensions between data and dimensions arguments: mu; ignoring dimensions in data.
Backtrace:
    ▆
 1. ├─testthat::expect_error(...) at test-models.R:422:5
 2. │ └─testthat:::quasi_capture(...)
 3. │   ├─testthat (local) .capture(...)
 4. │   │ └─base::withCallingHandlers(...)
 5. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
 6. └─nimble::nimbleModel(...)
 7.   └─md$setupModel(...)
 8.     └─nimble (local) assignDimensions(dimensions, inits, data)

[ FAIL 0 | WARN 1 | SKIP 0 | PASS 26665 ]
```

paciorek:
Yes, that is 'normal'.

PR #1617, opened 2026-03-30
Fix find var api calls expt1

perrydv:
do not merge.

This branch extends `fix-findVar-API-call` with further changes. See #1615 .

This branch removes use of `R_UnboundValue`, changes `GET_SLOT` to `R_do_slot`, removes use of `R_BracketSymbol`, and add a `PROTECT` around an `Rf_getAttrib` result passed into `getSEXPdims` (the new `PROTECT` really looks unnecessary as `getSEXPdims` does call any R function, but it is being newly flagged by `rchk` as far as we can tell). 

The change to `R_do_slot` was also done with superficial changes that reduced the number of `PROTECT` calls and matching `UNPROTECT` counts.

A difficulty in deciding what to do is that CRAN's demands for making API usages stricter are not completely clearly listed and documented.


perrydv:
Oops I guess I should have branched from devel instead of `fix-findVar-API-calls` because the latter was already merged. I had to manually resolve all the commits in favor of the new branch.

paciorek:
rhub checking suggests the changes are good. 

R_MissingArg is the only thing being flagged in terms of non-API calls.

The rchk getSEXPdims(SEXPREC*) error is gone, though the other weird rchk issues remain as we expected.

perrydv:
All non-Windows tests passed. We think there is an issue with our Windows CI setup, so I am running some tests on Windows locally. So far they are passing. I will report when they are completed.

@paciorek tried checks on r-hub and reports the confusingly phrased `PROTECT` issue seems to have been resolved. We think the additional issue (more poorly phrased) had to do with testing infrastructure as it occurred previously but did not manifest in official CRAN check results.

perrydv:
Following up: I was able to run all of the tests normally run on Windows successfully with one exception that does involve the PR: In `test-crossVal`, we apparently dynamically pull down an example that caused an error

`cannot open URL 'https://stat.columbia.edu/~gelman/arm/examples/arsenic/wells.dat': HTTP status was '403 Forbidden'`

I also ran `test-nimbleList` because I think that uses some of the changed code in this PR.

So I consider this PR to be passing tests.

paciorek:
I'll plan to merge this. Ok @perrydv ?

I'll rerun R CMD check, send to win-builder and run the key r-hub checks, and then resubmit.

perrydv:
Yes that sounds good to me.

PR #1624, opened 2026-07-21
Add propose/accept/reject logic to `prior_samples` sampler

danielturek:
This adds MH propose/accept/reject logic for the `prior_samples` sampler.

There is no longer an option for "sequential" draws from the provided `samples`, but rather on every MCMC iteration, a random draw from the provided `samples` is used as the proposal value(s).


perrydv:
This looks good to me.
It would be an easy extension to allow a vector of weights to be provided as well, with draws proportion to weights.
We are trying keep the next release really minimal and practical, but I am noting this for discussion.

danielturek:
@paciorek Are you able to re-generate the MCMC gold file using this branch `prior_samples2`?  I think that's the only issue with `test-mcmc.R`.

paciorek:
Yes, I can look at that.

paciorek:
@danielturek just to make sure....


Previously in the gold file, I see this pertaining to prior_samples sampler:

"""
[Note] prior_samples sampler will recycle sequential draws from prior samples, since 10 samples were provided, and 11 MCMC iterations will be run.
  [Note] prior_samples sampler will recycle sequential draws from prior samples, since 3 samples were provided, and 11 MCMC iterations will be run.
"""

Now when I manually run lines 2888-2931, I only see:

"""
  [Warning] No samplers assigned for 1 node, use conf$getUnsampledNodes() for node name.
"""

Is that what you expect? I guess the omission of the previous message makes sense given the new correct behavior, but I don't understand why the "No samplers" message would only now appear in the last MCMC configuration where 'b' is not assigned a sampler.

danielturek:
@paciorek Yes, this looks correct to me.  The MCMC configuration created on line 2926 specifies `nodes = NULL`, so no samplers are assigned by default.  Then samplers are manually added for `a` and `c`, but not for `b`.  So issuing this warning (from line 2929) makes sense to me.  In all the cases preceding that one, the MCMC configuration by default assigns samplers to all nodes, so such a warning is never issued.  Does this make sense?

